Issue #30, opened 2014-02-27
Is it possible to add support for custom query function like the select2?


mfzl:

VictorioBerra:
@kencaron That involves filter having a reference to the options globally.

PR #31, opened 2014-02-28
Expose $select.search for external / remote filtering.

mfzl:
Here is a working plunkr, http://plnkr.co/edit/1keJMP?p=preview

My attempt at fixing #30. I'm not sure it's the best solution, but it's a start and I'm using it. Might help others.


alflennik:
Great feature, just what I needed, but it looks like it's not in the docs?

PR #93, opened 2014-07-01
style(ngRequired): mark when invalid

dimirc:
When using ng-required a red border on control is shown (styles added all 3 themes)

The styles will show only if the control is _dirty_ and _invalid_, so to reproduce it you should first fill values on each field and then press `Clear ng-model`

You can try it with this [plunker](http://plnkr.co/edit/RN9Fry?p=preview)

screen shot 2014-07-01 at 12 15 59 am

### DIFF
--- a/src/select.css
+++ b/src/select.css
@@ -6,6 +6,10 @@
 
 /* Select2 theme */
 
+/* Mark invalid Select2 */
+.ng-dirty.ng-invalid > a.select2-choice {
+    border-color: #D44950;
+}
 
 /* Selectize theme */
 
@@ -17,6 +21,11 @@
 /* Fix dropdown width for Selectize theme */
 .selectize-control > .selectize-dropdown {
   width: 100%;
+}
+
+/* Mark invalid Selectize */
+.ng-dirty.ng-invalid > div.selectize-input {
+    border-color: #D44950;
 }
 
 
@@ -60,4 +69,9 @@
 .ui-select-match.ng-hide-add,
 .ui-select-search.ng-hide-add {
     display: none !important;
+}
+
+/* Mark invalid Bootstrap */
+.ui-select-bootstrap.ng-dirty.ng-invalid > button.btn.ui-select-match {
+    border-color: #D44950;
 }


abobwhite:
+1 more for multiple. Hitting a roadblock now that I find out this isn't supported. Has anyone come up with a solution?

hatpick:
Yes, you can watch the ng-model, if the selected array is empty $setValidity("required", false);

abobwhite:
@hatpick Thanks. Are you referring to modifying the ui-select code itself or doing this in my app?

I add this code from the pull request:
```
if ($select.multiple) {
              var defaultIsEmptyFn = ngModel.$isEmpty;

              ngModel.$isEmpty = function(value) {
                return (angular.isArray(value) && value.length === 0) || defaultIsEmptyFn(value);
              };
            }
```

It appears to work (i'm using angular auto-validate) by showing if there is an error after I submit the form but never before (adding then removing). After I submit the form, it shows as an error (correctly) but the error indication never goes away after that even after making a selection.

wry791:
+1 for multiple

jmorvan:
@hatpick thank you!

I got it working like a charm with the following directive:

```
myModule.directive('uiSelectRequired', function () {
    return {
        require: 'ngModel',
        link: function (scope, element, attrs, ngModel) {
            scope.$watch(function () {
                return ngModel.$modelValue;
            }, function (newValue) {
                if (ngModel.$modelValue && ngModel.$modelValue.length) {
                    ngModel.$setValidity("required", true);
                }
                else {
                    ngModel.$setValidity("required", false);
                }
            });
        }
    };
});
```

ma-zal:
Better version without "$watch":

``` javascript
angular.module( "myApp )
	.directive( 'uiSelectRequired', function() {
		return {
			require: 'ngModel',
			link: function( scope, element, attrs, ngModelController ) {
				var idName = attrs.name;
				ngModelController.$validators[idName] = function( modelValue, viewValue ) {
					return modelValue !== undefined && modelValue !== null && modelValue && modelValue.length > 0;
			};
		}
	};
} );

**Comment by zjhiphop** [comment]
Cool!!

**Comment by tb** [comment]
I use this with angular-fromly for validate ui-select with multiple:

    .directive('multipleRequired', function() {
      return {
        require: 'ngModel',
        link: function (scope, element, attrs, ngModel) {
          if (attrs.multipleRequired) {
            ngModel.$validators.multipleRequired = function(modelValue, viewValue) {
              return !!(modelValue || viewValue).length
            }
          }
        }
      };
    })

    <ui-select multiple ng-model="model[options.key]"
               multiple-required="{{to.required}}"
               ng-disabled="{{to.disabled}}">
    ...
    </ui-select>

**Comment by dsalvagni** [comment]
Hey, guys!
I just made some changes to your code to enable/disable the required option.
```javascript
.directive('uiSelectRequired', function(){
  return {
            restrict: "A",
            require: '^ngModel',
            scope: {
              uiSelectRequired:"="
            },
            link: function(scope, elm, attrs, ngModel) {
                ngModel.$validators.uiSelectRequired = function(modelValue, viewValue) {
                    var determineVal;
                    if (angular.isArray(modelValue)) {
                        determineVal = modelValue;
                    } else if (angular.isArray(viewValue)) {
                        determineVal = viewValue;
                    } else {
                        return false;
                    }

                    return scope.uiSelectRequired ? determineVal.length > 0 : true;
                };
            }
        };
})
```

Template:

```html
<ui-select multiple ng-model="...">
<ui-select-match ui-select-required="!filterModel.allOptions">
    {{$item.name}}
</ui-select-match>
<ui-select-choice>
    <div ng-bind-html="item.name | highlight: $select.search"></div>
</ui-select-choices>
</ui-select>
```

ma-zal:
You are using `ngModel.$validators.uiSelectRequired = ...`
How does it works if you have more ui-selects in one `<form>` ?

It is because I used:

    var idName = attrs.name;
    ngModelController.$validators[idName] = ...

Is my way right or wrong?

arkarkark:
@dsalvagni I misread your template and added ui-select-required to the ui-select element and had a problem with two directives wanting different isolate scopes. it was easily fixed with removing the scope: from the directive object and replacing the return line with.

```
return scope.$eval(attrs.uiSelectRequired) ? determineVal.length > 0 : true;
```

(could also be `!scope.$eval(attrs.uiSelectRequired) || determineVal.length > 0` )

It makes more sense to me to put this on the ui-select element since that's the one with the ng-model on it. (similar to ng-model and ng-required elsewhere).

dsalvagni:
Great point, @arkarkark. Thanks for sharing it.

You're right, it makes more sense. I tried to add `ui-select-required` to the `ui-select` element but I had the same problem that you mentioned. I didn't realized that I could use the `scope.$eval` to execute an expression on an attribute. I'll fix my code.

The second point, about the return statement, I think this is more readable: 
````javascript 
return scope.$eval(attrs.uiSelectRequired) ? determineVal.length > 0 : true;
````

arkarkark:
yeah, I kinda agree wth you on the return value @dsalvagni :+1:

PR #167, opened 2014-08-25
make CI build faster

PeterDaveHello:


aaronroberson:
@PeterDaveHello Please review this [SO](http://stackoverflow.com/questions/31278233/disadvantages-of-shallow-cloning-on-travis-and-other-ci-services) and provide additional details why you think a shallow clone is warranted and will not cause other unintended side effects.

PeterDaveHello:
Travis-ci uses shallow clone as default behavior for a long time, the tried 100 as depth before and now 50, in fact there is no problems.

PR #170, opened 2014-08-26
Avoid double-compilation of ui-select-choices directive.

dlongley:
Notes:

- Instead of replacing the ui-select-choices element using
  an angular template and then later transcluding content
  and recompiling the entire element at post-link time, do
  not provide a template but edit the template element in
  compile prior to linking. This approach allows the template
  to be fully constructed before the link phase and avoids
  any complications with double-linking the same element
  and/or issues with binding the transcluded contents to
  the wrong scope and introducing memory leaks.
- This patch was started prior to a fix that avoided linking
  the transcluded content to the wrong scope (which was
  also producing memory leaks), but solves the problem
  in a different way -- namely preparing all of the content
  before compilation. This also eliminates any need for an
  additional helper directive.
- This approach may help future-proof the uiSelectChoices
  directive as well -- as it was using "replace: true" which
  is deprecated in Angular. However, because the ui-select-choices
  element is no longer replaced, some changes had to be made to
  the CSS to account for its existence. These may need to be
  tweaked/simplified in the future -- the changes made were
  intended to be as non-intrusive/deviate as little as possible
  from any original intent.
- A test was changed to look for different exception text as
  the element that is transcluded is found by element tag
  or attribute now, not by classname.

### DIFF
--- a/src/select.css
+++ b/src/select.css
@@ -41,7 +41,8 @@
 }
 
 /* Fix dropdown width for Selectize theme */
-.selectize-control > .selectize-dropdown {
+.selectize-control > .selectize-dropdown,
+.selectize-control > ui-select-choices > .selectize-dropdown {
   width: 100%;
 }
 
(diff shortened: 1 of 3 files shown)


wesleycho:
Going to close this as too old - feel free to open a new PR with changes against current `master`.

PR #210, opened 2014-09-22
Fix IE onbeforeunload problem, with ui-select and bootstrap theme

harry-kalligeros:
Fix IE onbeforeunload problem, with ui-select and bootstrap theme


aaronroberson:
@harry-kalligeros Can you confirm if a more recent update has already addressed this issue? If so, please close.

PR #238, opened 2014-09-29
Fix label for input

### LINKED ISSUE #236: No id on input means it can't be attached to a label
You currently can't attach an external label to the search input and allow it to be focusable because the input doesn't get an id.

georeith:
Fixes #236


dmitry-dedukhin:
Hi.
I have the same issue, could you please update it?

juusopimia:
Hi,

I can't get external  "label for" to work, tried with 0.14.1 (should have the fix merged in as far as I can tell).

What I see with debugger is that id value is correctly placed for the search input field, but because it has "display: none" initially, clicking the label does not change focus to input.

Example Plunkr: http://plnkr.co/edit/u8dIsIuy4L90sMNYUfud?p=preview

Any suggestions?

wesleycho:
Closing due to history being broken (there should be no merge or revert commits), and this being significantly outdated.

Feel free to open a new PR with issues addressed re-implementing everything desired.

PR #249, opened 2014-10-01
ctrl.selected added to check where it's undefined

bartlenaerts:
When you have arrays coupled to ng-model that are null or undefined, you get errors when using ctrl.selected that also is undefined.

Added an extra check in the if-statements to be sure that ctrl.selected isn't null or undefined.


swalters:
I'm getting the same issue.  Any plans to fix this PR or should I make a new one?

aaronroberson:
@bartlenaerts and @swalters can you confirm if commit d95268c has resolved this issue? If so, please close this PR.

PR #251, opened 2014-10-01
set default value to empty array

ivanvoznyakovsky:
in case if items source is not defined in $scope when initializing multiselect instance of ui-select, we'll get an undefined from ctrl.parserResult.source($scope). then next line will throw an exception when truing to filter undefined.

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -201,7 +201,7 @@ uis.controller('uiSelectCtrl',
     }
 
     function setPlainItems(items) {
-      ctrl.items = items;
+      ctrl.items = items || [];
     }
 
     ctrl.setItemsFn = groupByExp ? updateGroups : setPlainItems;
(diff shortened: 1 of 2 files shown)


aaronroberson:
@ivanvoznyakovsky Can you update this PR in the uiSelectController.js file (~ line 197)?

Jefiozie:
@ivanvoznyakovsky  as @aaronroberson  mentioned above can you fix this PR?

ivanvoznyakovsky:
sorry. totally missed previous notification.
looks like the the case I've described is already handled but I found another one related to the same situation when items source is not defined.
updating

Jefiozie:
@ivanvoznyakovsky  thanks for the commit. Can you make a unit test for this?

ivanvoznyakovsky:
@Jefiozie added tests

PR #256, opened 2014-10-03
Add possibility to find value according to a property of the model

JeffGreat:
I needed that feature to preload existing model objects that are not in the source list.
The comparison is made upon the property setted via the "track by" of the repeater.


JeffGreat:
@amcdnl - I've rebased with latest and pushed it on my fork.
What do you mean by "Add the condition for the single value"?

abobwhite:
Hey guys - any movement on this? I'm surprised this didn't work from the start. As such, I am developing with the component and am stuck until the fix is merged or I hack up my own solution.

Just curious if it's planned to be merged soon.

brianfeister:
@amcdnl can you and @jeffGreat sort out your request here?

JeffGreat:
I've updated the pull request, all the test passed so it just need to be merged now. @dimirc can you merge?

brianfeister:
@JeffGreat - I can merge it but it sounded like you and @amcdnl still hadn't resolved one of his bullet points.

Flightfreak:
:+1:

JeffGreat:
@brianfeister you can merge since the last bullet point seems to be an improvment. On my side the feature work as is and can help others.
when @dimirc will explain me what is the last bullet I'll do another pull request.

burzum:
:+1:  Yes, please merge this ASAP, I've just wasted a good amount of time just to figure out this is not working with ui-select. :(

amcdnl:
@JeffGreat Just apply the track by change for single selections.  Currently you only have it for multi

amcdnl:
Merged ... I'll fix the other issue.

hsrobmln:
@JeffGreat Thank you so much for fixing this! I was going insane thinking I was doing something wrong. Exact same situation, I have a list of possibilities loaded from the server, and the selections are pre-populated on a different object, so they weren't matching. This solves the problem I've spent the last 3 hours on.

Ir1sh:
Can a tag be created that contains this update?  0.9.6 dosn't seem to have it

jukkasi:
I cannot make this work in my use case using multiselect, options are fetched async with refresh="fetch($select.search) from HTTP  backend

The problem is that selecting works fine and I can save to backend, but when backend returns results, UI is not updated accordingly.

My fetch() does not return all the options with empty search parameter but only limited list where previously selected value is or isn't included. The matching logic in this PR does not help since empty fetch is performed in some point and matching is performed against options where the selected value does not exist.

So my fetch() is really dynamic and the resulting options depend on given search parameter. I can easily fix this so that fetch() with empty search term would return all the results, but this does not make sense in performance point of view.

Any hints for this scenario?

Ir1sh:
@jukkasi I don't think this fix has been released yet, it is in master but it dosn't seem to be in the 0.9.6 build that is the latest on bower

@amcdnl is this correct? can we get a build done?

jukkasi:
@Ir1sh I understand it's not released and tested as local patch. I believe the problems with multiple are much bigger than this PR adresses.

Ir1sh:
I agree for one I don't understand why the object has to be in the choices object to be added to the selected

jukkasi:
Exactly. And other major issue reported elsewhere by many is that setting the model does not update view accordingly all of this is more or less related.

Ir1sh:
@JeffGreat if I change the track by to person.email instead of person.email in your test it still passes.  should it fail?

JeffGreat:
it should fail. Actually since the email is different there is no reason to get a selected value. May be the assertion is not right.

Ir1sh:
even if I remove the track by in el it still passes

Ir1sh:
works correctly in demo-multi-select.html

hkothari:
@dimirc @amcdnl : to @Ir1sh 's point above, would it be possible to get a 0.9.7 tag with this fix? It seems like there is sufficient desire from users to take advantage of "track by" support enabled by this.

zukasmichael:
hi guys ? how to get this patch or tag preloading working ? spent 4 hours yesterday on trying many workarounds and i'm about going to select2

brianfeister:
@amcdnl  - hoping to get a new release with some bugfixes out soon.  Should I assume those are 0.9.7 and this is 0.10.0?

zukasmichael:
i have tried 0.9.7 - still can't preload values if they are not in ui-select-choices http://plnkr.co/edit/o0pAujq9u4gCq2X96WEZ?p=preview

brianfeister:
Next release. 0.9.8

zukasmichael:
I'm still struggling how to properly load selected tag model.
This is plnkr with 0.9.8 ui-select http://plnkr.co/edit/o0pAujq9u4gCq2X96WEZ?p=preview
please click change model button twice slowly

zukasmichael:
with 0.9.8 model binding works just fine. Thank you for fast fix Brian!

zhendong590:
:+1:

RensBonnez:
Does the track by work already for single selection? Can't make it work.

bademux:
doesn't work without proxy object *multipleDemo*  *$scope.multipleDemo = {};* on v 0.12
seee for mo infohttp://plnkr.co/edit/o0pAujq9u4gCq2X96WEZ?p=preview

Issue #217, opened 2014-09-24
Is there any way to change multiple selection mode at the run time?


lzwsoar:
I would like use it like this: `<ui-select mulitple="{{isMultiple}}">`

rwlogel:
+1

Jefiozie:
Closing it down as it is a old issue. Please create new issue if this is still a bug/ problem.
Also i'm trying to get a good overview of recent bugs.

Thanks for understanding.

PR #292, opened 2014-10-10
Avoid having an ng-if if mutiple:

jmcpeak:
https://github.com/angular-ui/ui-select/issues/217 - Is there any way to change multiple selection mode at the run time?

Instead of the accessor (in case something needs to be modified) - how about this:

<ui-select-match placeholder="{{espLookupCtrl.placeholder}}">
      {{$select.match.fields[espLookupCtrl.displaycolumn]}}
</ui-select-match>


wesleycho:
Closing due to age - feel free to open a new PR against current `master` if this is still an issue.

PR #326, opened 2014-10-17
Minimum input length

iamWh1sp3r:
This adds the ability to set a minimum input length before search is triggered.


user378230:
Superseded #1291 (merged).

PR #327, opened 2014-10-17
Tagging support (continues #260)

brianfeister:
@dimirc I've rebased the latest from @tvervest and @juanpasolano from #260.  I've also added auto-tokenization support at your request, and custom labeling (with optional no-label) for new tags.  Note that javascript ignores trailing / leading whitespace characters inside data attributes so users will need to use the constant / literal `SPACE` if they want tagging with that character.  You can pass in pipes (so the only invalid character for tagging is `|`) to get multiple tag characters.  This can be seen in the tagging demo I've created in [my fork here](https://github.com/brianfeister/ui-select/blob/tagging-support/examples/demo-tagging.html).
#### Usage
1. `tagging` - this attribute accepts an optional string which is a scope function.  That function will be passed a string from the select input and return the transformed result will be injected into `ctrl.items` (available for selection). e.g. `<ui-select multiple tagging="tagTransform">`
2. `taggingTokens` - accepts a string that will be listened for in `keyup`, this keyboard character auto-tokenizes the current value and selects it.  **Multiple** tokens are supported by separating characters with `|`'s.  **SPACE** is a special case since whitespace characters are ignored in input fields by jQuery, as a result, `tagging-tokens=" "` fails to eval to whitespace, so the special key value `SPACE` is used instead.  e.g `<ui-select multiple tagging tagging-tokens="SPACE|,|/">` would listen for space, comma, and forward slash.
3. `taggingLabel` - (`false`|`string`) accepts an optional string for the label appended to tags in the dropdown (defaults to `(new)` if empty or undeclared.  If the string `false` is declared it will omit the injected helper (`{string} (new)`) value from the dropdown and tagging will trigger via tokenization shortcuts or `ENTER` key.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -201,6 +201,12 @@
 
         ctrl.activeIndex = ctrl.activeIndex >= ctrl.items.length ? 0 : ctrl.activeIndex;
 
+        // ensure that the index is set to zero for tagging variants
+        // that where first option is auto-selected
+        if ( ctrl.activeIndex === -1 && ctrl.taggingLabel !== false ) {
+          ctrl.activeIndex = 0;
+        }
+
         // Give it time to appear before focus
         $timeout(function() {
           ctrl.search = initSearchValue || ctrl.search;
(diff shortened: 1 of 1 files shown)


windmaomao:
thank you @brianfeister. I just run the example, exactly what i wanted, so now i just need to learn how to use it. Thanks all for the hard work.

pruchai:
Is there any way to disable new tag creation? I want users to only be able to pick tags available in the drop down menu.

brianfeister:
;)

darrudi:
@pruchai: Take a glance at examples. All this thread is/was about adding free tag support. :)

trickpattyFH20:
great feature to add tagging but I was really hoping this would work for just a regular select dropdown not just the "multiple" select. I have a single select and I'd like it if a user could add options to this list. I tried using tagging on a single select in the demo code and in my own code and it doesn't seem to work.  @brianfeister

pwchan:
+1

GFoley83:
@ceesvanegmond @brianfeister @parliament718 
As of version: 0.9.6, with tagging enabled, I'm still getting 
`Cannot read _uiSelectChoiceDisabled of null`.

Updating the following line:
https://github.com/angular-ui/ui-select/blob/master/src/select.js#L671 to 
`ctrl.select( newItem || undefined, true);` 
solved it for me. In other words, passing `undefined` instead of `null`, as @parliament718 mentioned.

Issue #328, opened 2014-10-18
Group Names do not display in IE


kentcooper:
When using IE (I tried IE 11) the group name's are not visible. I tried to trouble shoot it and it looks like {{$group.name}} is not rendering in IE. However if you switch to the ng-bind syntax it seems to work.

PR #329, opened 2014-10-18
Changed the choices.tpl to fix bug with IE groups

kentcooper:
Modified all 3 of the choices.tpl.html to fix a bug that was not
allowing IE to render group names. Instead of using {{}} for the name I
switched it to ng-bind syntax. This should fix the issue in #328

### DIFF
--- a/src/bootstrap/choices.tpl.html
+++ b/src/bootstrap/choices.tpl.html
@@ -3,7 +3,7 @@
     ng-show="$select.items.length > 0">
   <li class="ui-select-choices-group">
     <div class="divider" ng-show="$select.isGrouped && $index > 0"></div>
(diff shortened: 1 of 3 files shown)


just-boris:
@kentcooper, are you sure that you fix something?
Here is demo with v 0.9.1 (before your fix) http://jsfiddle.net/Ltc9kvo6/embedded/result/ and it looks good for me
How in latest version 0.9.6 it doesn't work anywhere http://jsfiddle.net/Ltc9kvo6/1/, because `ng-bind-html` requires sanitize passed content

just-boris:
@dimirc, I vote for revert this commit. Now I can't use groupping without `ngSanitize` module, which actually doesn't need for me

PR #335, opened 2014-10-22
An empty array should be considered as the empty value for multiple mode...

osmirnov:
.... ngOptions directive has the same behavior:

https://github.com/angular/angular.js/issues/2365
https://github.com/angular/angular.js/pull/5337


konrad-garus:
+1: Looks very straightforward and is an annoying bug/behavior.

Meanwhile a monkey patch for the impatient:

```
.directive('nonEmptyArray', function () {
    return {
        require: 'ngModel',
        link: function (scope, element, attrs, ngModelCtrl) {
            var original = ngModelCtrl.$isEmpty;
            ngModelCtrl.$isEmpty = function (value) {
                return original.call(ngModelCtrl, value) || value.length === 0;
            };
        }
    }
});
```

Issue #178, opened 2014-09-09
Maximum Selection Size


sagar-ranglani:
Allow the developer to limit the number of items that can be selected in a multi-select control.

georgeportillo:
+1

ankit-saxena:
+1

maxisam:
+1

jesselpalmer:
+1

DustinArmstrong:
+1

evillemez:
If anyone wants to have a max limit, see the following comment:

https://github.com/angular-ui/ui-select/pull/348#issuecomment-63736469

LeleDev:
+1

Lecss:
+1

pnvpratik:
+1

frankhock:
+1

rolivares:
+1

omkarv:
+1

alexgandy:
+1

cdjackson:
#1110

dimirc:
Should be solved with https://github.com/angular-ui/ui-select/pull/1110

PR #348, opened 2014-10-27
Add option 'limit' to limit the number of items that can be selected

fdevibe:
In multiple mode, use this option to limit the number of items that can be selected at the same time.

Quite possibly a fairly naïve approach. The tests also aren't optimal, but I haven't found a better way to test them without touching more code. At least it is a suggestion. :-)

This addresses #178.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -520,6 +520,10 @@
           processed = _handleMatchSelection(key);
         }
 
+        if (!processed && ctrl.multiple && ctrl.limit !== undefined && ctrl.limit <= ctrl.selected.length) {
+          processed = true;
+        }
+
         if (!processed && ctrl.items.length > 0) {
           processed = _handleDropDownSelection(key);
         }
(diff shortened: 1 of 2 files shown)


maxisam:
+1 Thanks for this @amoshche

davidstump:
:+1:  Thanks @amoshche!

evillemez:
:+1: Yeah, thanks @amoshche!

KennyKaplanHP:
Rather than doing another directive I adjusted the ctrl.activate function by adding in an if block around the remainder of the function. Using the attribute set logic from @fdevibe to add the limit to $select you can use the following to prevent the dropdown list from showing up:

if (ctrl.limit === undefined || ctrl.selected.length < ctrl.limit)

icfantv:
I don't see this added yet to the current code base - probably due to the merge conflict above.  Is there any intent of adding this feature to the released version?  Thanks.

voznik:
@amoshche +1 Thanks

user378230:
Superseded by #1110 (merged)

PR #349, opened 2014-10-28
Multiple async loading issue

ivanblazevic:
Options are not loaded if multiple select, $select.selected becomes
rewritten to null


wesleycho:
Closing due to age - if this is still an issue, feel free to open a new PR against current `master`.

PR #390, opened 2014-11-10
Added override for onSelect to control adding to selected

enderdickerson:
Needed the ability to define "onSelect" with complete control over which items were added to the selected array. The current implementation would always add the selected item to "selected" regardless of what occurred in the specified "onSelect" function. 

Test included.

e.g. User selects "All products" and instead of adding "All products," onSelect will add all the items as specified in the onSelect function.

Also, fixed broken test unrelated to feature change so CI would build correctly as it is currently broken on Master.


Jefiozie:
As this PR is very old and needs rebasing I think it is best to close it down. If needed please make a new PR with the adjustments.

PR #391, opened 2014-11-11
Initialise refreshDelay from uiSelectConfig

stschwark:
Not setting a refresh-delay attribute results in no delay although the uiSelectConfig shows the refreshDelay should default to 1000ms. I think this is because the refreshDelay setting gets initialised from within the attribute observer. This PR moves the initialisation out of the observer. Not sure if there is a better place though.


just-boris:
@stschwark, please rebase upon latest master. It can fix a test

src/select.js
```diff
@@ -901,6 +901,7 @@
             $select.refresh(attrs.refresh);
           });
 
+          $select.refreshDelay = uiSelectConfig.refreshDelay;
```
just-boris: here should be
```js
if(!attrs.hasOwnProperty('refreshDelay')) {
  $select.refreshDelay = uiSelectConfig.refreshDelay;
} else {
   attrs.$observe('refreshDelay', function() {/*etc*/})
}
```
because that's how `$observe` function works. It don't call a handler when attribute is doesn't exsits and we have to set  default value manually

PR #395, opened 2014-11-13
inline javascript may cause an error if the content policy restricts it

sclausen:
so I think a blank href may be the best option.

In Chrome I got:

<blockquote>Refused to execute JavaScript URL because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.
<blockquote>


Sjors:
The solution used by [ui.bootstrap](https://github.com/angular-ui/bootstrap/commit/fb302c6086543420b3561be3ed5c49ac3974c243) is to replace `href="javascript:void(0)"` with `ng-click="$event.preventDefault()"`. I'll make a new PR for that.

sclausen:
:+1:

sclausen:
The problem is, an a-tag without a href isn't valid anymore.

PR #405, opened 2014-11-16
Update select.js

perymimon:
change `reult==value` to `angular.equals(result,value)`
it very important!
without that init objects that came with the model not match objects in option list


ilanbiala:
@brianfeister @dimirc @amcdnl this does fix issue #540 and probably a couple others when using 0.9.5.

dimirc:
Same as #410

PR #410, opened 2014-11-18
 if (result == value)  to  "if (angular.equals(result,value)){"

perymimon:
without that object from server list not match object  on model list


dimirc:
This change will be applied at https://github.com/angular-ui/ui-select/pull/748

PR #420, opened 2014-11-19
Use object property of `track by` for comparison

richardallen:
This expands on PR #256 to improve efficiency, clean up the code, and use `track by` in single as well as multiple configurations.

* Moved the regular expression that determines the property name into `RepeatParser`.
* Cleaned up the formatter logic by breaking into functions that are created only once when the formatter is added.
* Used the object property specified via `track by` for comparisons in both single and multiple configuration.
* Added unit tests for both single and multiple configuration.
* Used `angular.equals` instead of `==` for comparison. The AngularJS [`ngOptions` directive](https://code.angularjs.org/1.2.26/docs/api/ng/directive/select) notes that it [compares by reference, not value](http://jsfiddle.net/qWzTb/), so it seems natural that angular-ui-select would do the same. However, several developers have posted issues on this behavior causing problems, so using `angular.equals` seems like a better choice. This change was made with a separate commit, so the rest of this PR could be merged without that commit.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -678,9 +678,9 @@
                 result = $select.parserResult.modelMapper(scope, locals);
 
                 if ($select.parserResult.trackByProp) {
-                  match = result[$select.parserResult.trackByProp] == inputValue[$select.parserResult.trackByProp];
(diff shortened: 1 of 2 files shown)


user378230:
Closing given that codebase has changed significantly since this PR was created and many of the changes/fixes probably no longer apply.

Issue #258, opened 2014-10-05
ng-required not validating form on submit


mouryaamit:
when using ui-select with ng-required or required attribute, form is not validating on submit action.

abobwhite:
@RichardAllen Has your fix been merged?

ravishivt:
+1

noetix:
+1

vtortola:
+1

olso:
+1

eliias:
+1

gsf-greece:
+1

dmitrygaraev:
Multiselect validation doesn't work in 0.9.9

yaneq:
Not working for me either with ui-select 0.9.9 and angular 1.3.11

yaneq:
I fixed it now by adding a custom validator for the multiple ui-select

``` javascript
module.directive('uiSelectRequired', function() {
  return {
    require: 'ngModel',
    link: function(scope, elm, attrs, ctrl) {
      ctrl.$validators.uiSelectRequired = function(modelValue, viewValue) {
        return modelValue && modelValue.length;
      };
    }
  };
});
```

``` slim
// in view
ui-select {
  multiple='true'
  ng-model='something.something'
  ui-select-required
}
```

[Plunkr](http://plnkr.co/edit/vGY70qhMLlpX8mGvHBRP?p=preview)

zentetsukenz:
@yaneq solution almost works for me. The condition `modelValue && modelValue.length;` is not always return true or false, because the `modelValue` sometimes is `undefined`, so I have to check first wether `modelValue` or `viewValue` is an actual array, and finally check its length.

Here's the code
```javascript

app.module.directive('uiSelectRequired', function() {
  return {
    require: 'ngModel',
    link: function(scope, elm, attrs, ctrl) {
      ctrl.$validators.uiSelectRequired = function(modelValue, viewValue) {

        var determineVal;
        if (angular.isArray(modelValue)) {
          determineVal = modelValue;
        } else if (angular.isArray(viewValue)) {
          determineVal = viewValue;
        } else {
          return false;
        }
        
        return determineVal.length > 0;
      };
    }
  };
});
```

This is my markup
```HTML
<ui-select name="items" multiple
  ui-select-required
  ng-model="item.items">
  <ui-select-match>
    {{$item.name}}
  </ui-select-match>
  <ui-select-choices repeat="item in items | filter: $select.search">
[...]

PR #422, opened 2014-11-19
#258 Validates ng-required when multiple is used.

richardallen:
This PR overrides `NgModelController.$isEmpty` when `multiple` is used so validation with `ng-required` works as expected.


abobwhite:
+1 Can we verify this works and get it merged?

user378230:
#1492 merged

Issue #423, opened 2014-11-20
ui-select should set css class ng-touched after the user touched the select


wald-tq:
I am using the combination `.ng-invalid.ng-touched` to show only validation errors after the user touched the form field.

The ui-select field should follow angular core and add this css class.

If I have time I can look into that and supply a pull request. It shouldn't be a big issue. But I would like to get feedback on the issue first.

graup:
+1

Purpwood:
+1

leandro-freitas-softdevelop:
+1

brunobastosg:
+1

wald-tq:
is already fixed in version 0.12

PR #426, opened 2014-11-20
Add ng-touched, if the user touched the input field.

wald-tq:
Pull request for my feature request #423


KennyKaplanHP:
There is a simpler way to do this without having to bubble up the Touched custom attribute. If you call the $setTouched() function then this will elicit the correct behavior.

E.g. 
ctrl.ngModel.$setTouched();
instead of 
ctrl.Touched = true;

PR #427, opened 2014-11-20
Add ng-touched, if the user touched the input field.

wald-tq:
Pull request for my feature request #423.

Change to #426:
* Removed debugger call

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -365,6 +365,7 @@
       if (!ctrl.open) return;        
       _resetSearchInput();
       ctrl.open = false;
+      ctrl.touched = true;
       if (!ctrl.multiple){
         $timeout(function(){
           ctrl.focusser.prop('disabled', false);
--- a/src/select2/select.tpl.html
+++ b/src/select2/select.tpl.html
@@ -1,7 +1,8 @@
 <div class="select2 select2-container"
      ng-class="{'select2-container-active select2-dropdown-open': $select.open,
                 'select2-container-disabled': $select.disabled,
-                'select2-container-active': $select.focus }">
+                'select2-container-active': $select.focus,
+                'ng-touched': $select.touched}">
   <div class="ui-select-match"></div>
   <div class="select2-drop select2-with-searchbox select2-drop-active"
        ng-class="{'select2-display-none': !$select.open}">
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -204,6 +204,15 @@ describe('ui-select tests', function() {
     expect(isDropdownOpened(el)).toEqual(false);
   });
 
+  it('should append ng-touched css class after it has been touched', function () {
+
+    var el = createUiSelect({theme: 'select2'});
+
+    clickMatch(el);
+    el.find(".ui-select-toggle").click();
+
+    expect($(el).hasClass('ng-touched')).toEqual(true);
+  });
 
   it('should open/close dropdown when clicking caret icon', function() {
 


jmayday:
Hi,

I'm encountering some problems with the code you provided. Also test you extended doesn't cover my case unfortunately. 
According to what I see in code while inspecting it in Firefox I would say everything is fine - 'ng-touched' appears in class. Reality is different. I can't make use of myform.myselectui.$touched. Please have a look a my plunker: http://plnkr.co/edit/6otoQUgon1HoJoKXyOyR

Is it still a bug or I'm doing something really wrong?

Aonlaan:
This "fix" will only address this issue with the select2 theme with single selection. There are about 5 templates that would need this changed.

jmayday:
Yes, I have noticed it. But I think I've patched necessary entries in template cache. OK, just be sure - I've just tried with 'select2' theme patching 'select2/select.tpl.html' and updated plunker. Doesn't seem to work anyway. Despite class gets 'ng-touched' I can't make use of $touched.

wald-tq:
Hi,

sorry, was on vacation :smiley: 

I think my "fix" is just a workaround for me. I'm not using the property $touched. Only the css class ng-touched for styling of errors.

I also think there should be a better way for doing that (use the ng-model controller).

jmayday:
How do style then? 

Let's talk about it having my plunker as example. 

I'm doing something like  data-ng-class="{'has-error':myform.selectize.$invalid && myform.inputselect.$touched}" on one of my form groups I want to have validated (in this case I want to validate field with name selectize). myform.inputselect is an input, I added the name attribute "inputselect" in bootstrap/select.tpl.html to have access to it.

I was thinking that I'm really close to achieve acceptable solution (similar to what $touched would give me) using this approach. On plunker this code works just fine! Unfortunately in my project it's not working. And I observed something extremely strange. Maybe you could help me with tracking the issue source?

So it's like this:

- in my select-ui.js file I also patched the bootstrap/select.tpl.html file to have access to inputselect
- to have better overview on what's going on I like to put myform.inputselect | json inside <pre> tags somewhere on my site
- after clicking on select control on my site I was expecting to see myform.inputselect.$touched set as true, but I didn't.... until I came back to my computer after leaving it locked for something like 5 minutes! I was also somehow able to recreate this scenario once more. But it's really nondeterministic... 

So, the question is what is going on? I thought that accessing <input> from select.tpl.html directly will be simple and easy and will give me working solution for what I want to achieve (show touched and invalid form group on a form). And it's like this on plunkr but not on my project which has some really big lag on 'refreshing' real values of myform.inputselect ($viewValue, $modelValue, $touched and so on). 

Any ideas?

wald-tq:
I'm styling like this:

less:
```
.form-control.ng-touched.ng-invalid,
form.ng-submitted .form-control.ng-invalid {
    // normal fields, not select2
    &:not(.select2) {
        border: solid 1px @state-danger-text;
        background: @state-danger-bg;
        .form-control-focus(@state-danger-text);
    }

    // for select 2 fields we style the child a-element:
    &.select2 {
        // invalid select2 field
        a.select2-choice {
            border: solid 1px @state-danger-text;
            background: @state-danger-bg;
        }
        // for focus we have to use the select2-container-active class
        // we can't use .form-control-focus
        &.select2-container-active a.select2-choice {
            .box-shadow(~"inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px @{state-danger-text}");
        }
    }
}
```

In a nutshell I'm only applying a red border if I have a field which has the class ng-invalid and ng-touched aswell.

Sorry, can't take a closer look in your code for now.

jmayday:
@wald-tq - thx. I will share my solution at the evening. It's quite pure one (of course there are some doubts).

jmayday:
Hi again. 

Because plunker is running slow right now and I can't run demo site I'm not sure if I updated my plunker correctly so I'll describe step by step what I did. 

In templateCahce I modified entry for bootstrap:/select.tpl.html by adding name='{{inputName}}' attribute to input tag. 

In link function of uiSelect directive I had to make use of attrs.inputName by adding <code>scope.inputName = attrs.inputName;</code>

Then finally in my sites html file I used such snippets:

data-ng-class="{'has-error':myform.myfield.$invalid && myform.myfieldinput.$touched}" for form-group
div ui-select name="myfield" input-name="myfieldinput"...

and some span for displaying error block:

ng-if="form.myinputfield.$touched"

I'm not sure if what I did with scope in link function is done correctly. Normally I could write such code: 

scope: { inputname = "@" } to use one way binding. But then I break inheritance of scope. So, did I combine these 2 things right?

When plunker is back I'll refactor code in my demo to make it more clear and straightforward. I won't do it now because probably I would only break something ;)

Cheers.

jmayday:
OK, plunker is working again. I did some fixes and can confirm that the example is working.

MiguelMike:
+1

AndrewEastwood:
when is this gonna be merged?

wald-tq:
I get class ng-touched in version 0.12.x

brunobastosg:
I only get class ng-touched if I use the mouse to cause the focus loss. Using the keyboard it stays as ng-untouched.

wald-tq:
@brunobastosg yeah, can confirm that. I think you should open a new issue for that. This is a very old and quick and dirty pull request from me.

brunobastosg:
@wald-tq , thanks for the quick reply. I opened issue #1090 for this.

PR #437, opened 2014-11-25
Prevent selection of hidden search input on close

Nilanno:
Here is a case of issue:
1) click on uiSelect element with search enabled
Result: select options shown, search input is focused, user can enter any filter text
2) click on any another text input in document
Expected: another input is focused and user can enter text into it
Actual: another input is focused, but in a next moment it looses focus, user can't enter text to it

Unfortunately, I don't have enough experience to write karma test in reasonable amount of time.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -840,7 +840,7 @@
           }
 
           if (!contains && !$select.clickTriggeredSelect) {
-            $select.close();
+            $select.close(true);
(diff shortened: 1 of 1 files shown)


brianfeister:
This code has moved and is using $scope.digest() instead now.  Closing until we have verification it's still present.

PR #441, opened 2014-11-26
pass the event to the onSelectCallback

jcompagner:
a small patch so that the $event is also given as an argument to the onSelectCallback for the user to use


Jefiozie:
Looks like this needs some work. Closing it down as it is a very old PR. If needed reopen with the rebased code. Thanks.

PR #471, opened 2014-12-04
close after blur

JulianWielga:
close select after blur, eg. from TAB. was broken when search found nothing.


Jefiozie:
@aaronroberson or @user378230 , looks like it is fixed as I cannot reproduce it any more with the latest?

PR #486, opened 2014-12-09
Add Allow-Clear support to Bootstrap Template

DanielClausen:
Bootstrap theme support for allow-clear attribute.

### DIFF
--- a/src/bootstrap/match.tpl.html
+++ b/src/bootstrap/match.tpl.html
@@ -1,9 +1,12 @@
-<button type="button" class="btn btn-default form-control ui-select-match" tabindex="-1"
-        ng-hide="$select.open"
-        ng-disabled="$select.disabled"
-        ng-class="{'btn-default-focus':$select.focus}";
-        ng-click="$select.activate()">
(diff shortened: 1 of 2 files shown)


howieweiner:
@bampakoa's suggestion works well, but in my case, it's not practical to add multiple methods to multiple controllers, so I wrote a child directive. Suits my purpose.

I'm wondering if this might not be a cleaner solution than attempting to rework the ui-select-match directive - considering the introduction of bootstrap columns caused so many issues?

Here's my directive: https://gist.github.com/howieweiner/f9db8d386a507defad2b

abobwhite:
This works, but the styling is royally jacked up. See issues #578.

dimirc:
Based on @bampakoa [comment](https://github.com/angular-ui/ui-select/pull/486#issuecomment-67307275) I pushed PR https://github.com/angular-ui/ui-select/pull/667

Demo [plunker](http://plnkr.co/edit/QGElZ3h8FV9g1Quf38pG?p=preview)

howieweiner:
Unfortunately, this doesn't work in IE (10). The dropdown opens. I have the same issue with my directive mentioned above.. really can't work out how to prevent event propagation when the X is clicked

Issue #158, opened 2014-08-12
Mouseenter (selection) doesn't work in 1.2.16+ using IE8


sambomartin:
On mouseenter the selected item moves to the last one.  The mouseover event fires for each row in the collection - always leaving the last one selected.

Current work-around is to use 1.2.15.

Josh68:
@careywalker, your code (minor modification based on version) worked for me with 0.9.6. I'm going to look to see if this could apply cleanly to the latest version, because it wouldn't seem to cause a merge issue.

gillius:
#754 seems to work for me. We have no choice to support IE8 and IE11 now during a enterprise migration to IE11, and Angular 1.2.16 has a bug impacting IE11, but upgrading beyond select 0.2 doesn't work in IE8, and it doesn't work in Angular past 1.2.16. So now we will use @Josh68's code.

jroom36:
https://github.com/Josh68/ui-select/commit/8317a0c55e25f629427b212e6e696a78c0a82dd5
This fixed for me follow issue:
"On mouseenter the selected item moves to the last one. The mouseover event fires for each row in the collection - always leaving the last one selected."

angular js 1.2.28
select.js from https://github.com/Josh68/ui-select

cgonul:
I am sorry but it is still not working by my side. Can you eleborate an example. I am helpless. Thank you very much.

PR #488, opened 2014-12-09
Fixed the event delegation in ng-repeat

jonathanchris:
I used jlfama/ui-select#4d14014d943a5450a2ad7bd24d8acc5810a3345a from #158, as a reference and only kept the fixes for the event delegation issues.


user378230:
Closing as this fix or a similar one made its way into master at some point (https://github.com/angular-ui/ui-select/blob/9a40b6f3348345b7628e6d9b5caf3b9218c808ca/src/uiSelectChoicesDirective.js#L52)

Issue #41, opened 2014-04-01
Append the list to body


jimmywarting:
I have a select in a div with overflow hidden/auto
In every (most?) usecase it's desirable to let the select go outside of the overflow element, like the native select and [select2][1] dose)

Its true for all kind of typeahead, custom `<datalist>` polyfill things

  [1]: http://ivaynberg.github.io/select2/

lebster:
@jimmywarting @pabuisson hmm looks like i am  having the same problems with scrolling when using position: fixed. Did you guys come with any alternative solutions? I think I am going to have to go back to the normal select. All other solutions out there seem to have issues too.

jpdesigndev:
There seems to be little to no movement on this. I'm thinking the best (perhaps only) way to fix the issues around dropdowns and overflow content is to append the dropdown to the body element, and position it absolutely. My team and I are racing toward a deadline, and the UI relies on `ui-select` elements frequently. That being said, I will be working on this solution at some point in the future (date unknown).

Does anyone have any recommendations/requirements, assuming `<body>` append and `position: absolute` that I should keep in mind?

drewbeck:
@jpdesigndev Why do you consider hiding dropdown on body scroll a requirement? This is not standard behavior for regular select dropdowns and I'm not sure why it should be implemented here.

jpdesigndev:
@drewbeck Well, that's what I've seen.

Here's a quick breakdown (All latest browsers)

* OS X
  * Chrome
    * Disallow body scroll
  * Safari
    * Disallow body scroll

* Windows 7
  * Chrome
    * Allow scroll, but hide select dropdown immediately 
  * IE
    * Disallow body scroll

*** 

Given this small dataset, perhaps looking into disabling scrolling for everything other than the dropdown list is desirable. That could prove rather complex, I imagine. Thoughts?

drewbeck:
Thanks for that breakdown! I didn't know Windows Chrome was different. I wonder why that is?

I think disallowing body scroll makes more sense, given these options. With 2-finger scroll it's easy to accidentally have your cursor in the wrong box — having the dropdown disappear in that case I think would be very frustrating. 

My ideal from a UX perspective would be that the dropdown scrolls with the browser. This isn't standard but I've seen it and I like it quite a bit more than the no-scroll case (see dropdowns on seamless.com for an example). It would be amazing to have this as an option. But disabling body scroll entirely would be consistent with most browsers and most users' experiences I believe.

[...]

PR #491, opened 2014-12-10
Solves Append the list to body #41

dhaval-patel:
Created a layer at body level and appended dropdown to it. So it will not affect the height of actual control and it will also prevent mouse wheel after dropdown is opened.


dmccown1500:
If this works well lets get in in!

leipert:
@dhaval-patel Could you please add the changes to the corresponding files in the src folder instead of the dist file? We cannot see what you changed.

amcdnl:
Please remove the dist folder and only submit the src changes.

cmlenz:
#718 also implements this feature.

dhaval-patel:
@amcdnl done !!

dhaval-patel:
@cmlenz My solution has default as "Append to Body". But I can make it as your saying. It will be much better.

leipert:
@dhaval-patel . No, please have a look at #718 for example. You should not have changes in the dist folder. The dist should not appear at all!
And the changes you made should be in the individual src files. We do not see what you changed. If you change the uiSelect directive, change uiSelectDirective.js.

If you want to test ui-select with your new feature you have to
```bash
$ npm install
$ gulp test
$ gulp build
```

If you do not have gulp installed globally, do it with `npm install -g gulp`

dhaval-patel:
@leipert I don't see any individual files in my repo. My repo contains only select.js. May be you guys have changed your repo structure after I pulled.

leipert:
Ah okay, I did not know. I am not a maintainer ;) Well hopefully one will come along and give some advice and either merge this PR or #718

brianfeister:
@dhaval-patel there was a major structural change in v0.9.9 and your CSS indicates that your branched off at 0.9.4.  Sorry for the inconvenience but I can't see your changes with the current PR, you need to rebase current master onto your branch before I can review this.  Also, I don't think this will be merged because #718 seems to do the same thing and is already properly rebased.

PR #496, opened 2014-12-10
Added uiSelectFooter directive for bottom of list

csebold:
My humble attempt to contribute a footer directive for the bottom of the dropdown menu


mrwarrens:
I would be a fan if this was merged.  My project has several cases where we need a footer, and at the moment we are hacking up a solution involving two sets of dom nodes inside of ui-select-choices and turn them on/off if it is the last element.

csebold:
This is a feature of Select2, but I don't see it on ui-select's roadmap.  I'd love to see this rolled in, though.  If there's something I can do to make the code fit better, I'm willing to make it happen.

joshrickert:
A header would be helpful as well... Same basic idea.

shawn9273:
any idea on if/when this will be merged and released?

dimirc:
@csebold I think that this might be good, but I see some trouble on each theme:

## Bootstrap
* footer is shown when the dropdown is closed
* footer is shown on top of the dropdown

image
image


## Selectize
* footer is shown when the dropdown is closed
* footer isn't shown when dropdown is open
image
image


## Select2
You mention:
>This is a feature of Select2, but I don't see it on ui-select's roadmap

But I dont see this on select2. I wanted to see if there were some select2 css clases to keep same style as right now seems like without any padding or default style

image

plong0:
+1
anything more about having a ui-select-footer / ui-select-header?  I would really like this.

dorr-fg:
This would be very helpful for me as well. For instance, for providing a "toolbar" for creating new entities in lookup fields that use this combo-box.

plong0:
PR https://github.com/angular-ui/ui-select/pull/1323

Adds:
ui-select-header
ui-select-footer

user378230:
Closing in favour of #1323 when it works.

PR #501, opened 2014-12-11
on-remove and on-select should fire AFTER addition/removal of the item

gotenxds:
Fixed the issue where on-remove and on-select would fire before the scope was updated.


brianfeister:
Released in v0.9.6

PR #503, opened 2014-12-12
Fixed undesired submit action

andaviaco:
If you use ui-select inside a form, the form gets submitted when you click the 'caret' button, by adding a type in the button it prevents this behavior :3.
tested in Chrome v.38.0.2125.122 using AngularJS v.1.2.27


CesarLanderos:
not really, the second button doesnt have a type:

```
<button class="ui-select-clear" ng-if="$select.allowClear && !$select.isEmpty()" ng-click="$select.select(undefined)">
    <i class="glyphicon glyphicon-remove"></i>
</button>
```

popoklopsi:
Isn't really a great thing to fix, so why just not merge the pull request?
Because that is a really annoything bug.

brianfeister:
Released in v0.9.6

PR #506, opened 2014-12-12
Trim the trailing space left by a new tag's label.

pderksen:
Already done in #447, but got removed by https://github.com/angular-ui/ui-select/commit/41ba28ffb3f807c71634bc449f282e1b259a1c0a.


brianfeister:
Released in v0.9.6

Issue #289, opened 2014-10-10
Form containing the ui-select get set dirty when value hasn't been changed


pandamouse:
http://plnkr.co/edit/5RJRul3Cge2L1yEuS5R5?p=preview 

The form is always $dirty as soon as it loads if it contains the ui-select.

cmlenz:
This was fixed via #509. This issue should be closed.

PR #509, opened 2014-12-15
#289 - Multiple select to not set form to dirty as soon as it load

pandamouse:
Fixes #493 
Reincorporating the changes from @CasperTDK + test fix.


dennislloydjr:
Thanks for the fix! This works great. Any guesses when it will make it into a release?

pdeveltere:
This is still a problem for me... :(

PR #537, opened 2014-12-19
Improve semantics / architecture of bootstrap templates

koorb:
Right now the bootstrap templates end up looking clunky and they have no identifiable classes so are impossible to restyle. This set of changes removes the `col-` classes and adds targetable classes `.ui-select-placeholder` `.ui-select-match-text` and `.ui-select-clear`. It moves the caret and clear buttons into the the main toggle and allows the main toggle button to be disabled (which was a aesthetics more than functionality)


brianfeister:
Hey @koorb - thanks again for this.  I found that your direction didn't work for the multi select permutation of the directive.  Thanks for your pointers though, v0.9.6 includes some of what you did here and apologies for not including it, seemed like more to back out than to include.

koorb:
Thats fine @brianfeister it was my first pull request and I got it in a bit of a mess, but glad it was useful.

PR #545, opened 2014-12-23
Improve evaluation of group-by expression

richstandbrook:
I was unable to group-by nested properties eg. `group-by="country.name"`or functions. This PR uses Angular's `$parse` method to resolve the expression and seems to work better.

I have had to leave the `$scope.$eval` call in for compatibility – without it you need to write the expression without quotes `group-by="country.name"` rather than `group-by="'country.name'"` which I would prefer but breaks the tests and therefore existing implementations.


src/select.js
```diff
@@ -244,8 +244,8 @@
       function updateGroups(items) {
         ctrl.groups = [];
         angular.forEach(items, function(item) {
-          var groupFn = $scope.$eval(groupByExp);
-          var groupName = angular.isFunction(groupFn) ? groupFn(item) : item[groupFn];
+          var groupFn = $parse($scope.$eval(groupByExp));
```
wesleycho: This is incorrect usage of $parse. $parse is used such as
```js
$parse(groupByExp)($scope)
```
This is essentially equivalent to $scope.$eval as far as this use goes though.

user378230:
I'd prefer the idea of dropping the single quotes maybe we just stack this with any other breaking changes till then.

Jefiozie:
As there is not comments on this PR  (only from maintainers). I will close it down. If needed people can reopen/ comment and we can have a look at it.

Issue #282, opened 2014-10-09
Inproper selection behavior on scroll if mouse hovering over list


kikonen-fiksu:
If mouse is hovering over dropdown list and list is scrolled using UP or DOWN key then behavior is inproper.

What happens:
- say dropdown fits 7 elements and mouse is hovering over 3rd visible element
- If dropdown content is scrolled up or down (via keys) then when scroll happens, selection jumps into element under mouse hovering position, and selection moving via arrow keys continues from there

=> Causes very irritatingly "jumping" selection when browsing through elements.

Ganasist:
I just experienced this behavior now. Is there a way around it?

dimirc:
Should be solved with https://github.com/angular-ui/ui-select/pull/1211

PR #547, opened 2014-12-23
fix for #282

alannesta:
This issue is caused by the inproperly triggered 'mouseenter' event when scrolling happens.
To fix this issue, we need to capture the event and prevent it from propagating if it is triggered by a scroll (which is handled by the _ensureHighlightVisible function call ).
Due to the fact that mouseenter event is triggered differently than normal dom events that bubbles, see: https://developer.mozilla.org/en-US/docs/Web/Events/mouseenter, changing ng-mouseenter to ng-mouseover would enable as to capture the mouseover event and prevent it from propagating.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -167,8 +167,10 @@
     var ctrl = this;
 
     var EMPTY_SEARCH = '';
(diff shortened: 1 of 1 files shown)


brondavies:
:+1:

dimirc:
Should be solved with https://github.com/angular-ui/ui-select/pull/1211

Issue #527, opened 2014-12-17
Accessibility of ui-select


aeisenberg:
We are starting to use ui-select in one of our products. However, we have a requirement that our product be accessible [WCAG 2.0 AA](http://www.w3.org/TR/WCAG20/). In order to meet these requirements, all inputs elements must be titled (eg- have a title attribute, an associated label, etc).  As far as I can tell, ui-select does not meet these requirements.

Am I missing something? Are there any plans to fix this? Would you accept a PR to fix this?

amcdnl:
Thanks for the PR.  I'll review and get back to you.

aeisenberg:
Great. To be clear, this PR is based on work started by @0xF013.

user378230:
Closing as PR was merged.

PR #559, opened 2015-01-02
Add acessibility

aeisenberg:
Pull request for Issue #527.

### DIFF
--- a/examples/demo-bind-to-single-property-async.html
+++ b/examples/demo-bind-to-single-property-async.html
@@ -65,7 +65,7 @@
   <h3>Select2 theme</h3>
   <h4>Single property binding with async data</h4>
   <p>Selected: {{personAsync.selected}}</p>
-  <ui-select ng-model="personAsync.selected" theme="select2" ng-disabled="disabled" style="min-width: 300px;">
+  <ui-select ng-model="personAsync.selected" theme="select2" ng-disabled="disabled" style="min-width: 300px;" title="Single property binding with async data">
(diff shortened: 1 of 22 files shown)


brianfeister:
Thanks @aeisenberg. A cursory glance of this looks good.  I'd love a second set of eyes on it @amcdnl.  In the meantime can you rebase?

aeisenberg:
Rebased.

amcdnl:
Overall it looks good.  There is a ton of bindings introduced but I dont know what/if any impact that might have on perf.  I like the generated ids concept too.

Have you checked out: https://docs.angularjs.org/api/ngAria ... was introduced recently, haven't used it myself yet.

aeisenberg:
Thanks for the review.  I was not aware of ngAria. I'll have to take a
look. ui-select needs to work on 1.2.x, so I don't think it would be
appropriate in this project.

amcdnl:
Merged

0xF013:
@aeisenberg  :+1:
@amcdnl ngAria does a lot of stuff but it doesn't cover sophisticated cases like marking the active descendant of the combobox (meaning that the screen reader won't tell the user there are autocompleted options or which of them is currently selected).

Another accessibility issue is the multiple selection. It allows arrow navigation and backspace to delete the selected item but the selected block is not focused, so the screenreader doesn't know that you actually selected something. That might be addressed in a separate PR

amcdnl:
@0xF013 Ya, lets do that in a separate issue / pr

lindenthal:
Hi guys,
what is the status here? We use UI select in openproject.org. And we need it to be fully accessible. Can we help out?
Regards
Niels

amcdnl:
This was merged. ^^^

Issue #561, opened 2015-01-03
Style: Bootstrap. Missing type="button" for template bootstrap/match.tpl.html


patricklehmann:
This causes in an error, where the form is submitted when hitting this button

realityking:
Also #536

wtilton:
Just got this as well.

realityking:
Fixed with 1e25650d78242abfe54bfd1c50d4997485729b8c

wesleycho:
Closing as fixed.

PR #536, opened 2014-12-19
Buttons should always have an attribute type with the value button.

realityking:
This prevents submitting a form that contains a UI Select.

PR #562, opened 2015-01-03
Adding type=button to template #561

patricklehmann:


realityking:
You've only changed the generated file, not the actual template. See #536.

wtilton:
Definitely a bug. I just ran into this problem as well. Without type="button" the form the control is contained in submits.

realityking:
Fixed with 1e25650d78242abfe54bfd1c50d4997485729b8c

PR #572, opened 2015-01-10
Fix placeholder not being shown for select2

gonzedge:
The placeholder is currently not being shown for the single-selection select2 theme. This pull request adds the missing placeholder attribute and tests that the placeholder is present on all themes.


gonzedge:
Nevermind, this just duplicates the empty option text as a placeholder for the search text box in the select2 theme, which is not very helpful to be honest. Because of this, I'm closing this pull-request.

I would love for there to be a way to specify the search box's placeholder as a separate option for the select2 theme. Maybe that's a feature request instead.

itchyny:
I think this is actually a good patch. I use the select2 theme in my product but due to this placeholder problem, I have to overwrite the theme in my script.

Issue #582, opened 2015-01-13
multiple: input resizing on container resizing + container padding


klmdb:
When we put padding on .ui-select-container, the input doesn't size properly.
Also, the input doesn't resize when the container's size or padding changes.

I'll create a pull request with my fix.

wesleycho:
Going to close as more information is needed - feel free to open an issue with reproduction in Plunker based on the latest release with more information if this is still an issue.

andreash:
This is also the case with the "selectize" theme - couldn't check it with the select2-Theme. Would be great, if that could get fixed. Tried myself to find the root cause - it appears, the sizeSearchInput() is not called, when the search in put becomes visible. But I can't location the point, where it is being made visible (the click on the container).

PR #583, opened 2015-01-13
Better input resizing

klmdb:
fix for [#582](https://github.com/angular-ui/ui-select/issues/582)
- resize input on container resize
- take padding into consideration


src/select.js
```diff
@@ -513,29 +513,39 @@
       return ctrl.placeholder;
     };
 
-    var containerSizeWatch;
     ctrl.sizeSearchInput = function(){
-      var input = _searchInput[0],
-          container = _searchInput.parent().parent()[0];
+      var input                    = _searchInput[0],
+          container                = _searchInput.parent().parent()[0],
+          matchedElementsContainer = angular.element(input).parent().children()[0];
```
wesleycho: This can be rewritten as
```js
matchedElementsContainer = _searchInput.parent().children()[0];
```

src/select.js
```diff
@@ -513,29 +513,39 @@
...
         }
-      }, 0, false);
+
+        var matchedElementsContainerOuterWidth = getSizeProperty(matchedElementsContainer, 'width') || matchedElementsContainer.clientWidth || matchedElementsContainer.offsetWidth;
+
+        return Math.max( 50, containerInnerWidth - matchedElementsContainerOuterWidth - 10);
+      };
+      var setWidth = function(width){
+
+        _searchInput.css('width',width+'px');
```
wesleycho: px addition isn't necessary, if it is a number, it is assumed to be in px.

src/select.js
```diff
@@ -513,29 +513,39 @@
...
+
       _searchInput.css('width','10px');
-      var calculate = function(){
-        var newWidth = container.clientWidth - input.offsetLeft - 10;
-        if(newWidth < 50) newWidth = container.clientWidth;
-        _searchInput.css('width',newWidth+'px');
+
+      var getSizeProperty = function(element, property){
+
+        return parseInt(window.getComputedStyle(element, null).getPropertyValue(property) , 10);
```
wesleycho: This is a very expensive check - may be a good idea to find a way to cache the information of `window.getComputedStyle` and find a better way to hook into it.

Jefiozie:
Closing it down as this PR is not been updated more than a year

PR #586, opened 2015-01-14
Revert "Add acessibility"

amcdnl:
Reverts angular-ui/ui-select#559 due to errors


myabc:
Why?

amcdnl:
Build was failing, was going to check to see if that was the cause.

PR #590, opened 2015-01-16
Remove duplicate file comment in select.css

realityking:
Was introduced in 230216516ce5d929665e4089a45cab9ea3ed2f69

Since the comment is preserved my minifiers this wastes some bytes if the file is directly included.


realityking:
Anyone?

Issue #308, opened 2014-10-15
dropdown appears cutoff when used inside a div with overflow: auto


pjfsilva:
Problem:
When ui-select is used inside a div with overflow: auto defined it doesn't appear "on top" of the div but instead is contained inside that div that obviously causes usability problems.

Example: http://plnkr.co/edit/cAZ548
click on the select to open it: the dropdown is inside the red div but it should be "always on top" of everything.

Solution:
The default <select> from HTML always appears on the top of everything and the original jquery select2 already solved the problem (angular-ui/ui-select2 also solves the problem).

The problem is simple: ui-select uses a relative positioning where it needs to use absolute positioning to "break out" from the overflow div. jquery select2 solves this in a very smart way by appending the list to the body element, effectively avoiding any div parents with overflow or relative positioning that can cause other problems. (an absolute div inside a relative div isn't anymore absolute to the whole document: http://css-tricks.com/absolute-positioning-inside-relative-positioning/ )

Using absolute positioning can also cause problems because if this dropdown is used inside a modal window that is draggable or if the user scrolls the page: the dropdown (select2-drop) position needs to be adjusted to the new position of the content.

select2 solves using a "select2-drop-mask" around the whole document and when a user clicks anywhere closes the dropdown which is a good way to deal with this because the default <select> from HTML also does this.

select2 example working fine: http://plnkr.co/edit/4jERkO - click on the select: content is on top of everything.

I believe the solution is already on jquery select2 plugin and it's just a matter of borrowing the good ideas from them. I couldn't find the issue that originated that specific fix on select2, but this seems to be a good lead: https://github.com/ivaynberg/select2/issues/149

My end goal is to use the select inside a Bootstrap modal window (modal-body has overflow auto content) that would be draggable but I simplified my test case to this specific scenario.

I'll be happy to help and provide more clarification as I investigated this quite a bit.

Tim-Erwin:
This does the trick for me:
```
    .ui-select-choices {
        position: fixed;
        top: auto;
        left: auto;
        width: inherit;
      }

wesleycho:
Closing as a duplicate.

ibecir:
Tim-Erwin thanks dude, this stupid issue took me 3 days to resolve.

ibecir:
@Tim-Erwin thanks dude, this stupid issue took me 3 days to resolve.

JBongars:
on div.btn-group add style="z-index:auto;" like so: 

````html
[...]

Issue #234, opened 2014-09-29
Dropdown options are hiding down, if there is no room space. 


DineshkumarT:
![roomspace_issue](https://cloud.githubusercontent.com/assets/8824303/4436940/8ed922b6-4787-11e4-8310-5336d503dae7.JPG)

Normal dropdowns are displaying the options over the debugging tools,but the ui-select dropdown is displaying the options under the debugging tools.

Not able to select the options in this scenario.

drewbeck:
+1 for drop above. Default browser select controls do it, anything that wishes to replace that functionality should also do it!

I'm a ui-select2 user who is aching to be able to switch to ui-select, and this is definitely a requirement.

dtomaszewski:
+1

Oceanswave:
+1

Jefiozie:
I think this is already resolved in a version. Added a screenshot what i have.

![image](https://cloud.githubusercontent.com/assets/17835373/19966656/0c3f0fee-a1cd-11e6-8de1-ccd54cfc2bd4.png)

@user378230 , still a enhancement (maybe i'm missing something here) or can we close it?

user378230:
Not sure, this is a old issue though, so we'll close it. I'm sure someone will shout if they have a problem still.

mwpowellhtx:
This is still an issue, unless I missed a fix for it somewhere along the way. In my case I have a selectmenu within a modal dialog.

![image](https://cloud.githubusercontent.com/assets/1824187/22959823/3cac8afe-f307-11e6-8deb-bf6e6b903315.png)

gonza8888:
I have the same issue, when I want to select, it goes behind my next item.

![image](https://cloud.githubusercontent.com/assets/1005027/23561562/7b895af4-001d-11e7-8a8f-32abb1f211e7.png)

And heres the code:

```
    <div class="list">
		  <div class="item item-divider">
	      Completar los datos para estacionar
	      </div>

		  <div class="item" >
		    <!--<i class="icon ion-mic-a"></i>-->
		    Calle seleccionada
		    <span class="item-note">
		    {{estacionar.calle | json}}
		    </span>
		  </div>

		  <div class="item" >
		    Calle
		    <span class="item-note">
			  <ui-select ng-model="estacionar.calle" theme="selectize" ng-disabled="disabled" style="width: 300px;" title="Calles">
			      <ui-select-match placeholder="Escribir calle...">{{$select.selected.name}}</ui-select-match>
[...]

Issue #206, opened 2014-09-20
Changing position of dropdown


mattcasey:
One feature that I liekd from ui-select2 is the dropdown could be injected at the top-level of the DOM, so it would appear outside of DIV's with overflow:hidden, it also could switch between dropping down and dropping up.

Any roadmap to support this, or ideas to implement with CSS? I wouldn't mind working on a patch myself when I have time either.

leandro-freitas-softdevelop:
+1
<ui-select  class="btn-group dropup"> </ui-select> for Bootstrap.

leandro-freitas-softdevelop:
+1
> \<ui-select  class="btn-group dropup"\> </ui-select> for Bootstrap.

leandro-freitas-softdevelop:
+1
> \<ui-select  class="btn-group dropup"\> \</ui-select\> for Bootstrap.

eltoncezar:
+1

PR #591, opened 2015-01-16
move drop down to base of <body> to stop it from being cut off (drop.js)

aeharding:
Hi, guys! I'd like to open up the discussion on separating the drop down from the input field. This provides many benefits, including having it pop over any content areas that many have overflow hidden, scrolling, etc.

I did a proof-of-concept to demonstrate how this could work in this pull requests. I'm currently using [drop.js](https://github.com/HubSpot/drop) for tethering and whatnot. It works with the select2 style in the example I created, with a situation where a normal UI Select dropdown would be hidden past the containing div's red border.

This allows the dropdown to flow outside of the area it's 'contained' within.
(For example, modals and scrollable content areas.)

image
Plunker: http://plnkr.co/edit/EFQWcabsZBxC81Aa0wRC?p=preview

Related issues: #308, #234, #206, #41... And more! :-D

Please do not merge this commit in its current state. :-) Shit's broken and I tested exactly 0 use cases besides that in the example above. Again, this is a POC to spark discussion.


nickradford:
+1 I like the idea. Many other libraries do that sort of DOM manipulation in order to avoid document flow / overflow / scrolling issues.

vmattila:
This seems to be pretty good solution to this problem (or the problems mentioned on all the related issues). I had an issue with dropdown hiding outside the window when located near the bottom of the browser window, but after applying these modifications the dropdown opened automatically into a proper place (also above the search box).

However, I do not know if including other libraries is good idea (or accepted by the library maintainers' policy?)... so just thinking, would it be possible to separate this amendment to an external library, bridging drop.js and ui-select together?

aeharding:
@brianfeister (or other maintainers) any thoughts here? I would be willing to pull out functionality into this project if it is what's needed for an eventual merge. (However, testing the floating in this project will be fun.) Would probably just get the important parts of tether.js (which drop.js uses), unless there is a better base out there.

FFelling:
@aeharding Do you think this can work for the bootstrap theme?

aeharding:
@FFelling definitely, I just have to tweak some things. I didn't event check the bootstrap theme. ;)

aeharding:
@dimirc or @brianfeister Do you have any opinions on how this should be resolved? I'm willing to invest some time into making a much cleaner solution (with tests) if you give me an idea on the general direction (including 3rd parts deps or not).

santialbo:
Could this be made an optional attribute on the `ui-select-choices` directive, similar to the `tooltip-append-to-body="true"` from ui-bootstrap?

brianfeister:
One of @dimirc's goals with this project was it to be a native, lightweight alternative to select2.  As such, we want to minimize our dependency matrix.  I'm opposed to using an extra library like drop.js.  It's 12k which is non-trivial since that will be added on top of our existing footprint.  If you want to rewrite this in native Angular then I would be willing to consider a PR for this feature.  I agree with @santialbo about style.  Make it an option on the `ui-select-choices` directive, but do the work without drop.js.  

Also, your current PR is failing the Travis CI build, make sure that passes and make sure to write your own test coverage for your feature.

aeharding:
Cool, I'll get working on it! It'll be completely different code, since it won't be a hack.

brianfeister:
Awesome, ping me when you're done and I'll try to take a look.  Apologies in advance for communication latency.  Day jobs, money, and all that boring stuff.

aeharding:
FYI, the old stuff from this PR was moved to this branch:
https://github.com/aeharding/ui-select/tree/drop.js-hack

I am working on new stuff (with the latest from master, as of yesterday) here:
https://github.com/aeharding/ui-select/tree/tooltip-append-to-body

I will be incrementally committing. Please comment here with any suggestions -- I will reference this PR when creating a new PR for my new stuff. Eventually, I'll squash and conventionalize my commit(s).

adam-marshall:
What happened with this?

jmsunseri:
I also would like to know what happened to this.

sonphnt:
Have we had any good solution for this proble?

aeharding:
Please look at the appendToBody attribute in the documentation.

https://github.com/angular-ui/ui-select/wiki/ui-select

dannygoncalves:
@aeharding necro post but, this appendToBody doesn't work for me, when I use it (having the ui-select inside a modal-body), I cannot click on the input field anymore

Issue #453, opened 2014-11-29
Input box is hidden with search-enabled="false" with Bootstrap theme


cdjackson:
With the Bootstrap theme, if I add the search-enabled="false" attribute, when the dropdown is clicked, the input box is hidden. This seems strange behaviour - maybe it's intended, but if so, can it be disabled? I want to keep the input box, but just not accept input (similar to how the Selectize theme works).
Aside from being a nuisance in itself, it also causes the parent to resize when the dropdown is opened.

gen4sp:
+1 unexpected and unable to fix without fork
Please, remove it or make optional.

donbobka:
:+1:

chuckrea:
+1 Just happened to me as well. Very odd.

brentmiller:
I am experiencing the same issue but with the select2 theme. Thus the issue isn't limited to the bootstrap theme.

jedimindtrix9:
Yes, this is weird. I set a placeholder for default text ie (-select an option-). I am using the selectize theme: Two things I've noticed...1. If I switch the theme to bootstrap, the placeholder option will display. The other option was to set search-enabled to "false". Any help here would be greatly appreciated.

seinavco:
+1

Please fix this issue.

SamPlacette:
+1 This is affecting me as well. I am working around the problem for now by specifying placeholder text using :before pseudo-selector.

sagivf:
+1 It would also be desirable for this to work with the "selectize" theme.
```css
.ui-select-container[search-enabled=false] .ui-select-match.ng-hide {
  display: inline-block !important;
}
.ui-select-container[search-enabled=false] .ui-select-match .text-muted.ng-hide {
  display: inline !important;
}
.ui-select-container[search-enabled=false].ng-dirty .ui-select-match .text-muted.ng-hide {
  display: none !important;
}
```
Thanks @SamPlacette.

jedimindtrix9:
One solution that I found works well with AngularJS if you are familiar with it. The way I solved this is to simply leave the theme as selectize but target you scope in js by using a for loop like this

for(var i=0; i<$scope.yourDropDownName.length; i++){
                        if($scope.yourDropDownName[i].id=='US'){
                            $scope.model.yourScopeModelName=$scope.yourDropDownName[i];
                            break;
                        }
                    }

readme42:
+1

luklapp:
[...]

PR #595, opened 2015-01-20
Fix input box hidding

gen4sp:
https://github.com/angular-ui/ui-select/issues/453
Input box is hidden with search-enabled="false" with Bootstrap them

fixed


austin-sa-wang:
Could we get this merged for those who are waiting on issue #453?

readme42:
This works, merge it please!

wesleycho:
@user378230 may be worth merging this via `curl https://patch-diff.githubusercontent.com/raw/angular-ui/ui-select/pull/595.patch | git am -3` and resolving any conflicts/modifying the commit message locally via `git commit --amend`. This modification looks like it needs to be on all templates.

wesleycho:
Thoughts on this @user378230 ? I can merge this locally and fix up the conflict in this instance fairly easily if desired.

user378230:
@wesleycho  I'm not a 100% sure what it does, its a minor change though - merge for now, we can revert if need be. 

For reference #1543 has the same change.

CoperNick8:
This fix should also be applied to dist/select.js file in this line
`$templateCache.put("bootstrap/match.tpl.html","<div class=\"ui-select-match\" ng-hide=\"$select.open\"  ...`

PR #596, opened 2015-01-20
Hidding caret if search enabled in bootstrap theme

gen4sp:


dimirc:
Why will we want to hide it if search is enabled?

PR #600, opened 2015-01-22
Fixed undefined function

parisholley:
This is fixing the symptom to a real problem (unsure what that is), but the matchItem comes across as a TrustedValueHolderType vs a string and causes this to break.


dimirc:
Can you create a plunker reproducing it?

parisholley:
problem was pretty obscure, no clue how to reproduce at this point

dimirc:
I see that basically you are checking if there is a replace method on matchItem but not since we cannot reproduce it, maybe we should close this PR until having more info how will be actually usefull. Feel free to reopen the issue if you get the error back or a way to reproduce it. Thanks

PR #611, opened 2015-01-23
fix(ngModel) Allow uiSelect ngModel to populate from parent directive.

swalters:
This allows directives to include uiSelect and add parsers/formatters to ngModel

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -884,7 +884,7 @@
       },
       replace: true,
       transclude: true,
-      require: ['uiSelect', 'ngModel'],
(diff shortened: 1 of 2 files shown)


dimirc:
What could be a use case for this?

swalters:
If you want to use ui-select as a reusable directive, then you need the ability to use the ng-model controller that your wrapper component is using.  I'm specifically using ui-select in ui-grid 3.0 (ng-grid) and I also use custom drop-down directives throughout my application that use ui-select.

Issue #316, opened 2014-10-16
Pagination (far-future)


jimmywarting:
Thought this would be a discussion about how pagination should work.

For the scroll pagination solution i think it would be best if it depended on [UI-scroll](http://angular-ui.github.io/ui-utils/#/scroll), rather then reinventing the wheel. It's fun when everything is more modular so we can improve small bits instead.
It would work more as an optional enhancement drop-in "If it exist we will use it" kind of thing.

I also found this [Multi level menu](http://tympanus.net/Development/ResponsiveMultiLevelMenu/index.html) recently and I thought it would be cool to have a kind of optgroup thing working like that

Issue #221, opened 2014-09-24
Infinite scrolling with remote data available?


doranen:
Don't see a way to provide the data necessary to enable infinite scrolling using query

kiwipiet:
+1

ephraimm:
+1

yusin:
+1

NielsGregers:
+1

asm0dey:
:+1:

jpmckearin:
+1

tushariscoolster:
+1

arjundas:
+1

PavelKonon:
+1

Xepe:
:+1:

gouxlord:
+1

radyz:
+1

simkessy:
+1

zsapkagy:
:+1:

tadchristiansen:
+1

kodermax:
+1

rapmue:
+1

tushariscoolster:
Use this with ui select;
it works
https://github.com/tushariscoolster/angular-scroll-complete

eg

<ui-select  multiple data-ng-model="$parent.data.category" theme="bootstrap" data-ng-disabled="disabled" on-remove="data($item)">
                                <ui-select-match placeholder="Enter an pillar">{{$item.category}}</ui-select-match>
                                <ui-select-choices percent="70" when-scrolled="data($select.search,false)" repeat="category in categories" refresh="data($select.search,true)" refresh-delay="500">
                                    <div ng-bind-html="category.category"></div>
                                </ui-select-choices>
                            </ui-select>

simkessy:
@tushariscoolster  does it destroy the elements when you scroll back up? If I have a large list, it's good for a start but once a user begins to scroll down the list, if it just keeps adding to the dom it will begin to slow things down.

tushariscoolster:
@simkessy  yupps I agree, I too faced that problem, but dint got much time to enhance that
But for performance improvement u can try https://github.com/bimal1331/paperDom
Am not sure it work

nevcos:
+1

andrewboni:
+1 on this

mkamayd:
+1

chris-hanson:
+1

angeldeejay:
+1

hyzhak:
+1

pkreuzinger:
+1

[...]

PR #620, opened 2015-01-30
Added ui-scroll

domakas:
* Added ui-scroll
* Added infinite scroller
* Added loader

### DIFF
--- a/src/bootstrap/choices.tpl.html
+++ b/src/bootstrap/choices.tpl.html
@@ -1,10 +1,11 @@
 <ul class="ui-select-choices ui-select-choices-content dropdown-menu"
     role="listbox"
-    ng-show="$select.items.length > 0">
+    ng-show="$select.items.length > 0 || uiScrollLoading"
+    ng-class="{ 'ui-select-choices-loading-empty': $select.items.length === 0 && uiScrollLoading }">
   <li class="ui-select-choices-group" id="ui-select-choices-{{ $select.generatedId }}" >
     <div class="divider" ng-show="$select.isGrouped && $index > 0"></div>
     <div ng-show="$select.isGrouped" class="ui-select-choices-group-label dropdown-header" ng-bind-html="$group.name"></div>
-    <div id="ui-select-choices-row-{{ $select.generatedId }}-{{$index}}" class="ui-select-choices-row" 
+    <div id="ui-select-choices-row-{{ $select.generatedId }}-{{$index}}" class="ui-select-choices-row"
     ng-class="{active: $select.isActive(this), disabled: $select.isDisabled(this)}" role="option">
       <a href="javascript:void(0)" class="ui-select-choices-row-inner"></a>
     </div>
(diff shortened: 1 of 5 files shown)


Mironor:
Looks like a fix for #221

rahulmr999:
The ui-scroll is still not available right ? I don't see the changes in the latest build. Will this be fixed anytime soon ? I cannot migrate from ui-select2 because of this .:(

PavelKonon:
+1

scharfmedia:
+1, i  see this PR is a month old without feedback, any ETA for this?

domakas:
I will try to add a example page for this when I have time, maybe it will be merged faster then.

amcdnl:
Whats the premise behind adding ui-scroll vs just native?

rahulmr999:
@domakas DId u get time to look into this ? I am not able to migrate because of this . Please .

dimirc:
Is this adding a dependency to other library? we should avoid external dependency and maybe work on compatibility

amcdnl:
I second that

asm0dey:
If we should avoid dependency - maybe it's option to add this as separate bower module, that integrates with ui-scroll on one side and with ui-select on another?

domakas:
Still haven't time to make the demo page, but I hope to make it soon.

Regarding dependencies, I implemented ui-scroll just because of #316 suggestion. Either way, it's optional, you don't have to add ui-scroll to your project if you are not using this scrolling feature.

nlevari:
I wish I had a demo or a readme file instead of Digging the code for answers. Any ETA?

amcdnl:
@domakas I see and totally see the value there....but I don't think thats the typical scenario and users of that would typically does a AJAX search instead.

@dimirc thoughts?

rvoronyak:
Will this be merged into the master soon? ETA?

jonathanchris:
@domakas Thanks for the work on this, but I need some assistance with an example or at least just how this would look to get it working. Specifically with the infinite attribute, I can't seem to put anything in there  that works. So far i've pulled down your branch, built it, and put it in my source. Then tried to add it to an existing working ui-select example but i'm not getting any results.

PowerKiKi:
@scharfmedia, @nlevari, @rvoronyak without any tests nor documentation, the chances to get it merged seem pretty slim

hummorsa:
Hi, is his going to be included in the ui-select at all ? or is something was cancelled ?

amcdnl:
@hummorsa No, I do not believe so.  It doesn't fit in with the vision of ui-select.  I'd advise @domakas to make a public repo that includes this and I'll cite it in the docs.

mlibanori:
how use ui-select + ui-scroll ?? is there some example ??

filipe3oliveira:
@domakas how can we use this?

domakas:
Sorry for delayed response for all of you. After some time working with this, it's still buggy and probably shouldn't be used yet. This requires big time effort to make it work properly, which I don't have at the moment. I will try to dig down a demo page, so someone maybe could continue working on this.

Sorry for all the fuss.

Issue #623, opened 2015-01-31
Empty/blank tag added to list when pressing Enter with no option selected.


JeromeLam:
The cause of this issue may be similar to #520.  Here's a screenshot of what it looks like:
![image](https://cloud.githubusercontent.com/assets/1011816/5990314/9d8da2c0-a978-11e4-8b33-0d1213da2140.png)

To reproduce, just open the [tagging demo](https://github.com/angular-ui/ui-select/blob/master/examples/demo-tagging.html), focus on the input under *Simple String Tags
(Predictive Search Model / No Labels)* and press Enter.  I validated this against v0.9.6 and master.

Thanks for looking into this and keep up the great work!

Issue #520, opened 2014-12-16
Null tag added to list (race condition)


aeharding:
Hi guys,

I'm happy that tagging is being added to ui-select, however, it's still pretty rough around the edges. 0.9.5 fixed a lot of the things that made it unusable (couldn't select first thing, other breaking stuff), but I've encountered another smaller issue with 0.9.5.

http://plnkr.co/edit/Yx4I7EEM7jn7XVdKqsbL?p=preview
![image](https://cloud.githubusercontent.com/assets/2166114/5458339/fc009596-8516-11e4-8e99-b2b1f7ccc1ea.png)

Click inside the input area, and then type a comma and then hit enter very quickly (it doesn't occur when you press comma and then spacebar, for some reason). A null item will be added to the array of items, which is annoying since it causes many errors in the console, and cannot be removed by clicking the 'x' or using the delete button.

Also, why does the dropdown flicker when tagging is enabled? It's kinda annoying, and it'd be nice if that could be fixed.

I appreciate the work done here guys!

NJAldwin:
I'm seeing this as well, though it's not a race condition in this case.  Pressing the down arrow (to open the dropdown), then pressing the up arrow (to deselect anything), then pressing enter causes a null element to be added.  The null element does not appear in the list, so it probably shouldn't be able to be selected.

![dropdown with nothing selected](https://cloud.githubusercontent.com/assets/145036/5694748/4154e1c2-9946-11e4-8c07-76002975c3ff.png)

![dropdown with a null element](https://cloud.githubusercontent.com/assets/145036/5694746/117dbbd6-9946-11e4-9b3a-d4c0dc8d52f9.png)

Furthermore, this null element does not seem to be removable, instead throwing this:

```
TypeError: Cannot read property '_uiSelectChoiceLocked' of undefined
    at ctrl.removeChoice (http://run.plnkr.co/Agl9e65wEUyqtG1P/select.js:497:23)
    at getNewActiveMatchIndex (http://run.plnkr.co/Agl9e65wEUyqtG1P/select.js:609:20)
    at _handleMatchSelection (http://run.plnkr.co/Agl9e65wEUyqtG1P/select.js:625:18)
    at http://run.plnkr.co/Agl9e65wEUyqtG1P/select.js:648:23
    at Scope.$eval (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.18/angular.js:12556:28)
[...]

PR #626, opened 2015-02-02
Only allow highlighted dropdown item to be selected using ENTER.

JeromeLam:
This addresses issue #623, which is specific to tagging.  I verified the change against the basic demo.html file and there is no impact.

I also checked this change against the various Plunker in #520 and their respective issues are no longer there.


brondavies:
:+1:

brianfeister:
:+1:

PR #632, opened 2015-02-04
Feature request / Bug Fix.  Pre-populating multi-select with selections not in source

MarkRBM:
This change adds a flag called 'allow-custom-selected' to the ui-select directive which when set to true allows the form field to be pre-populated with selections that are not in the dropdown options source array as long as the selection object has the correct 'track by' parameter.

I can give a detailed use case if necessary.

I am also open to suggestions on what the flag should be called as I don't think 'allow-custom-selected' is very clear.

Finally I am happy to write the necessary tests but I tried to copy and adjust one of the existing tests (line 162 'should correctly render initial state with track by feature') but I don't think that test is working correctly as I could remove the track by portion of the ng-repeat and the test would still pass.  If you can give me some guidance on how to write a test for this feature that would be great.

### DIFF
--- a/examples/demo-multi-select.html
+++ b/examples/demo-multi-select.html
@@ -164,6 +164,20 @@
   <p>Selected: {{multipleDemo.selectedPeopleSimple}}</p>
 
   <hr>
+  <hr>
+  <h3>Array of objects tracked by name and with initial items not in source</h3>
+  <ui-select multiple ng-model="multipleDemo.selectedPeopleCustom" allow-custom-selected="true" theme="bootstrap" ng-disabled="disabled" close-on-select="false" style="width: 800px;" title="Choose a person">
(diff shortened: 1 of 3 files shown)


Ir1sh:
Any way to get feedback on this? did I do something wrong? not follow a process? I know everyone is busy so how do I get this put on the priorities list?

Ir1sh:
can I get an update on weather this issue has been fixed in another pull request or its still outstanding?

Ir1sh:
Can you let me know if there is anything I can do to get this merged in? we just tried to update to latest and this is still a problem

Jefiozie:
Hi @Ir1sh  i see that you are working on this issue for a long time. Could you help me out with a small recap and maybe a rebase to the lastest version? 

Happy to help you and i'm trying to figure out what the problem is and if we need to fix this in this libary. As you can see we have many bugs and little help.

Thanks

Issue #624, opened 2015-02-01
JS tagging error when using up arrow to highlight the option before the first in the list.


JeromeLam:
Here's a screenshot of the error in Chrome's developer tools:
![image](https://cloud.githubusercontent.com/assets/1011816/5993560/41288f8a-aa23-11e4-80d4-370eb672a3f5.png)

To reproduce, just open the tagging demo, focus on any input field, make sure at least one option is highlighted in the dropdown list and use the up arrow key until the highlight goes through the top of the dropdown. I validated this against v0.9.6 and master.  Note that there is no problem if you used the down arrow key to try and go through the bottom of the dropdown.  The highlight stays stuck on the last item.

Thanks for looking into this and keep up the great work!

JeromeLam:
I've identified the potential cause for this error here:  https://github.com/angular-ui/ui-select/blob/master/src/select.js#L554

I can't seem to figure out why the condition ```ctrl.search.length === 0 && ctrl.tagging.isActivated``` was added for KEY.UP in #63.  If I haven't typed anything in $select.search, this allows activeIndex to become negative, leading to the dropdown highlight disappearing.

If anyone has any insight, please provide some feedback.  I can then create a PR to rectify this.  Thanks!

brianfeister:
Thanks @JeromeLam this is related to a condition where `tagging-label="false"` suppresses the value in the dropdown so that you can do "free tagging" but have it feel more like "free search", omitting the `My Tag (new)` bit.  This is how I'm using the directive in a combo / additive filter context.  Index tracking is affected differently between this and the other iteration where the dropdown list includes the new value.  If you submit a PR (again, would be **awesome**), make sure you test with `tagging-label="{whatever}"` but also `tagging-label="false"`

JeromeLam:
Hey @brianfeister, I created PR #638.  I hope the code changes are aligned with your expectations of index tracking.

PR #638, opened 2015-02-06
Issue/#624 - Add extra handling for negative activeIndex.

JeromeLam:
This addresses issue #624, which is specific to tagging. I verified the change against the basic demo.html file and there is no impact.

To @brianfeister's comment in #624, I tested successfully ```tagging-label="{whatever}"``` and ```tagging-label="false"``` using the two first sections in demo-tagging.html.


brianfeister:
Thanks @JeromeLam - I'm in a meeting ATM, but first-glance is that this will merge without issue.  Thanks for the assist, you caught what I missed :+1:

PR #639, opened 2015-02-06
Workaround to close other ui-selects when current ui-select is opnened

### LINKED ISSUE #228: BUG: Dropdown doesn't close
Because the dropdown is set to close on document click. If the user clicks on an element that prevents propagation of the click event then the dropdown will not close and the value will not reset to the selected one.

The close functionality needs to be taken off of a document click handler and attached to the input's blur event. Then a mousedown listener on the items will set a property to prevent the blur event from closing it in this case (as blur fires before click).

davidrapin:
fixes #228 with a workaround (trigger a document.click before opening current dropdown)

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -72,6 +72,7 @@
     angular.element.prototype.closest = function( selector) {
       var elem = this[0];
       var matchesSelector = elem.matches || elem.webkitMatchesSelector || elem.mozMatchesSelector || elem.msMatchesSelector;
+      if (!matchesSelector) { return []; }
 
       while (elem) {
         if (matchesSelector.bind(elem)(selector)) {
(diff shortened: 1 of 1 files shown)


amoch-gannett:
tests on builds are failing. has there been any traction on this issue since Feb?

arnogeismar:
I worked around it by listening to an onclick event on the ui-select in a wrapper directive and then manually looping over the others and closing them. 

iElement.on('click', function(iElement){
                       var targetElement = iElement.currentTarget;
// isolate-select-behandelaar is the name of my wrapper directive which indexes each ui-select
                       var targetElementIndex = $(targetElement).attr('isolate-select-behandelaar');
// class on the ui-select elements
                       var allUiSelectInputBoxes = $('.behandelaarInput');
                        $.each(allUiSelectInputBoxes, function(key, value){
                            if($(value).attr('isolate-select-behandelaar') !== targetElementIndex){
                                angular.element($(value).children()[0]).scope().$select.open = false;
                            }
                        })

                    });

wesleycho:
Closing as having too many issues (too much what appears to be unnecessary use of timeouts), as well as breaking tests. Feel free to open a PR that fixes the referred issue without excessive use of timeouts & on top of updated `master`.

Issue #640, opened 2015-02-07
An exact dropdown match while "free tagging" (no labels) doesn't auto-select.


JeromeLam:
Here's a screenshot of what I mean:
![image](https://cloud.githubusercontent.com/assets/1011816/6092773/41a4134c-aebc-11e4-9507-5ddc02a9960e.png)
**Setup**:  *multiple tagging tagging-label="false"*

To reproduce, just open the [tagging demo](https://github.com/angular-ui/ui-select/blob/master/examples/demo-tagging.html), focus on the input under *Simple String Tags
(Predictive Search Model / No Labels)*, type "Green" and press Enter.  When I press Enter, I assume the "Green" option would be selected since it's an exact match, although it's not highlighted.  The behavior is that what I typed would get cleared and nothing gets selected.  I validated this against master.

In an unlikely edge case, a duplicate would get created, but since that's not what's happening, I figure the exact match should get auto-selected.  If not, it would mean that a user free typing an exact match can only tag that match while using the mouse or the arrow keys to select it from the dropdown.  Is this the desired user behavior?

user378230:
Closing as I believe this is resolved in more recent versions of `ui-select`. 

If you believe its still relevant please update with a plunkr (or similar) against the latest version of ui-select (currently v0.15.0)

Thanks!

PR #644, opened 2015-02-09
Auto-select an exact dropdown match when "free tagging" (no labels).

JeromeLam:
This addresses issue #640, which is specific to tagging. I verified the change against the basic demo.html file and there is no impact.

I feel that if a user typed an exact match and pressed Enter, the tag should still get created.  The user shouldn't be required to click on the selection in the dropdown.


dimirc:
@JeromeLam could you add a test for this?

JeromeLam:
@dimirc This is my first time writing tests, and I struggled all day trying to make it work.  Could you give me pointers?

In [createUiSelectMultiple()](https://github.com/angular-ui/ui-select/blob/master/test/select.spec.js#L1178), I added ```if (attrs.tagginglabel !== undefined) { attrsHtml += ' tagging-label="' + attrs.tagginglabel + '"'; }```.  After that, I duplicated the [following test](https://github.com/angular-ui/ui-select/blob/master/test/select.spec.js#L1689) in hopes of modifying it without much success.  I figure that just by adding *tagginglabel: "false"* to the options of *createUiSelectMultiple()* would make it work, but the test fails.

Can you advise me on how to make it work?  Your expertise is greatly appreciated!

dimirc:
@JeromeLam just to reconfirm that what I understood you is ok, when a user types something that has an exact match, then the match should be selected? is this what you are trying to do or something different?

JeromeLam:
@dimirc That's correct.  In free-tagging mode with no labels, if you typed an exact match in the dropdown, the exact match wouldn't be highlighted.  If you pressed Enter to commit the exact match being typed (not selecting it from the dropdown using the mouse), it didn't get added without the fix in this PR.  You can always refer to #640 for steps to reproduce the odd behavior (at least to me).

wesleycho:
Closing due to what appears to be strange history.

The code seems to have changed since then - for example, the beginning to the file in this PR is present in src/common.js. Feel free to open a new PR on top of current `master`.

PR #647, opened 2015-02-10
Fixed issues with multiple tag creation/selection

ravishivt:
For multiple selects, update logic to check for already selected items
to use angular.equals() instead of indexOf() since indexOf() will fail
on the automatically added angular properties such as $$hashkey.

Updated keyup logic to add the "(new)" option if and only if the new
item doesn’t match the value of a possible selection choice or an
already selected item.  The existing logic would always add the "(new)"
option even if that new item exactly matched an existing selection
choice since the new item would differ in the isTag attribute.


ravishivt:
This looks similar to #640 but is different in the both the problem and the fix.  The issue here is that the multiple tagging feature would always prefer to create a new tag instead of selecting an existing tag.

For a demo of the problem, see http://plnkr.co/edit/ztgZ1JA0S0fuO9eSVQ3i?p=preview.  Typing in "Adam" in the Object Tags input, the "(new)" option is the only choice and not the existing Adam object.  Now test http://plnkr.co/edit/gl9jE9VMrJlJXRegXJhv?p=preview.  Typing in "Adam" shows the existing option as the only option.  This plnkr contains the patched select.js.

For both of these plunkers, the tagging function in the controller determines what user input should pull up the existing object.  This is ideal since the matching decision is in the user logic and not in the core code.

ravishivt:
Can somebody take a look at this pull request?  I've updated the code to merge nicely with v0.13.2.  I've also updated the [broken plnkr](http://plnkr.co/edit/f1pvkGT0AiT3YtRu5htC?p=preview) and [working plnkr](http://plnkr.co/edit/XdpsoKP1uE3q1UzX2z9S?p=preview) with the latest proposed changes.

aaronroberson:
@ravishivt Can you remove the dist/ files?

ravishivt:
@aaronroberson  Please see #1473 for pull request without the dist/ files.

wesleycho:
Closing as superceded by #1473

PR #651, opened 2015-02-11
Allow sorting of multiple select

homerjam:


wmertens:
:beers: :tada: :+1::exclamation:

dimirc:
@homerjam LGTM

Only quick issue that found was when dragging all the wat to the right, the index is pasted in the input
image

Btw, why not just `sortable` as attribute instead of `sort-enabled`?

homerjam:
I'll try and fix that, I guess we can disable the input while dragging or
something.

I went for `sort-enabled` because it followed the other options more
closely I thought. I'm not wedded to either though. Anyone else have an
opinion on that? I can easily change my PR, so let me know.

homerjam:
@dimirc I've resolved that problem you noted - I've also changed the attribute to `sortable`(think it does make more sense actually)

dimirc:
Thanks @homerjam

stevematyas:
Can anyone advise on the proper use of sortable="true"? Reference: #974

cheesecase:
Hi, is there a documentation how to use the sorting (sortable="true") for multiple select?

More specifically, is there an option to change the sorting from horizontal to vertical without changing the code?

Thanks

PR #654, opened 2015-02-12
fix(templates): use ng-bind instead of ng-bind-html

just-boris:
ng-bind-html requires sanitization of passed conent. It is not always possible and desired.


just-boris:
related #329

just-boris:
@dimirc, when do you plan new release?

dimirc:
Release [v0.9.7](https://github.com/angular-ui/ui-select/releases/tag/v0.9.7), thnx for the PR!

PR #659, opened 2015-02-13
A way to force placeholder in input field in multi select using "force-placeholder" attribute

kesy85:


dimirc:
What will be a use case for this? Also can you remove /dist files?

wesleycho:
Closing due to history issues (should be no revert commit) & lack of activity - feel free to open a new PR on top of current `master` if desired.

PR #671, opened 2015-02-16
fix for single select with simple tags

brianfeister:


dimirc:
You shouldn't change /dist files but [/src](https://github.com/angular-ui/ui-select/tree/master/src)

brianfeister:
*whew*, good catch, my fault for not noticing they were in the wrong spot when I opened the PR.  @trickpattyFH20, great if you can push a commit out to your fork that removes this change from `/dist/select.js` and adds it to `/src/select.js`

trickpattyFH20:
Sorry! I moved the changes from dist to src. 

Looks like the build doesn't like "== null" and it wants me to use "=== null" but using triple equals does not work... still gets "Cannot read property 'toUpperCase' of undefined" unless you use just a double equals...

Can this still be merged?

dimirc:
Will be also better of you could squash commits to try keeping master as clean as possible

trickpattyFH20:
@brianfeister everything is moved out of dist and into src.
@dimirc I squashed the commits into one and it's ready to go.

dimirc:
@brianfeister I'm not familiar with all the tagging functionality. As you created this PR, can you confirm if its ok and merge it to be able to ship it with next release?

brianfeister:
Thanks @trickpattyFH20, it will be in the next release.  Looks like @dimirc is on a roll merging PR's and reviewing things, I'll leave it to him to determine the next release.

dimirc:
v0.9.9 released

PR #673, opened 2015-02-17
If tagging mode is activate, do not check values

### LINKED ISSUE #540: ui-select modifying ng-model object
With this snippet:
```
<div class="form-group">
			<label for="users" class="col-sm-3 control-label">Users</label>
			<div class="col-sm-8">
				<ui-select multiple search-enabled ng-model="editingDesk.users" on-remove="editingDesk.removedUsers.has($item) ? editingDesk.removedUsers.delete($item) : editingDesk.removedUsers.add($item)" on-select="editingDesk.newUsers.has($item) ? editingDesk.newUsers.delete($item) : editingDesk.newUsers.add($item)">
					<ui-select-match placeholder="Select users...">{{$item.userName}}</ui-select-match>
					<ui-select-choices repeat="user in users | filter: $select.search">
						{{user.userName}}
					</ui-select-choices>
				</ui-select>
			</div>
		</div>
```

When the controller instantiates, `editingDesk.users` is properly populated. Once the controller finishes instantiating, `editingDesk.users.length` is 0. If I comment out the HTML above, then I don't have this issue.

Asimov4:
Motivation: Fix #540
The tagging mode is designed to allow new values to be added to the model.
It does not make sense to check those values against the list of choices when going from model --> view.

### DIFF
--- a/dist/select.js
+++ b/dist/select.js
@@ -1004,6 +1004,7 @@
                 return false;
               };
               if (!inputValue) return resultMultiple; //If ngModel was undefined
+              if ($select.tagging.isActivated) return inputValue;
               for (var k = inputValue.length - 1; k >= 0; k--) {
                 if (!checkFnMultiple($select.selected, inputValue[k])){
                   checkFnMultiple(data, inputValue[k]);


dimirc:
Dont change /dist files but only /src. Also could you add a test for it?

Asimov4:
Good point. I created a new pull request: https://github.com/angular-ui/ui-select/pull/683

PR #663, opened 2015-02-14
Allow discard tag when tagging function returns null

alexcrown:
Issue #485

PR #678, opened 2015-02-18
fix(tagging): allow to decline tag creation

dimirc:
Rebase of @alexcrown PR https://github.com/angular-ui/ui-select/pull/663


LeleDev:
I've updated ui-select to 0.9.9 but, when the tagging function returns null, i get the following error: `Uncaught TypeError: Cannot set property 'isTag' of null`

dimirc:
@LeleDev Can you setup a plunker reproducing this to check it?

PR #683, opened 2015-02-18
If tagging mode is activated, do not check values

### LINKED ISSUE #540: ui-select modifying ng-model object

Asimov4:
Motivation: Fix #540
The tagging mode is designed to allow new values to be added to the model.
It does not make sense to check those values against the list of choices when going from model --> view.

### DIFF
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -1725,6 +1725,26 @@ describe('ui-select tests', function() {
 
       expect(el.scope().$select.multiple).toBe(true);
     });
+    
+    it('should preserve the model if tagging is enabled on select multiple', function() {
+      scope.selection.selectedMultiple = ["I am not on the list of choices"];
+
+      var el = compileTemplate(
+          '<ui-select multiple="multiple" tagging ng-model="selection.selectedMultiple" theme="bootstrap" style="width: 800px;"> \
+              <ui-select-match placeholder="Pick one...">{{$item.name}} &lt;{{$item.email}}&gt;</ui-select-match> \
+              <ui-select-choices repeat="person.email as person in people | filter: $select.search"> \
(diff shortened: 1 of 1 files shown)


evillemez:
:+1: I definitely need this, hope it lands soon!  Good work!

Asimov4:
Just checking, there is no action on my side, right?

evillemez:
@Asimov4 According to my view there are merge conflicts - you might need to rebase on the current master branch.

Asimov4:
I reapplied the change on top of the HEAD of master.

Asimov4:
I pulled the latest changes and the unit tests passed without my change so I guess this is fixed.
Leaving the unit test as a pull request to guarantee this stays fixed.

wesleycho:
@user378230 this looks good, just adds a test for desired behavior & may close out #540.

Issue #649, opened 2015-02-11
Split src/select.js in multiple files


dimirc:
The file is growing too much and harder to read, we might need to split to one file per directive/service

Will this affect too much the current PRs?
We should modify gulp to contact files back for dist version

swalters:
+1

Issue #186, opened 2014-09-10
Why multiple theme approach?


mpolichette:
Hey guys, I am glad to see a good attempt at creating a select2 and selectize angular component.  However I'm confused by the approach to have multiple themes/templates.  

Since some of the components functionality is stored in the templates (hide/show etc..) it seems like it would be a better approach to get a consistent template down and then add CSS to style to get the bootstrap/select2/selectize look. 

I would guess having multiple templates/themes the way it is currently done would add unnecessary complexity and bloat to the project.  This is especially the case with the bower release, where you are forced to send ALL themes to the browser because they're bundled with the code.

Is there a reason that approach was not tried or abandoned?

Anyway, thanks for the valiant efforts so far!

mpolichette:
My issue was answered.

Issue #529, opened 2014-12-18
Remove theme specific templates


amcdnl:
I'd like to propose to remove the theme specific templates and css from the main project.  I think they could easily be broken out into sub-modules inside the main repo.This would:
- make the core stronger and easier to maintain
- make the download package smaller
- reduce the noise of bug specific templates 
- open up opportunities for others to add templates

@brianfeister and @dimirc - what are your thoughts?

sijmenvos:
@clairity Do you have a semantic-ui theme for angular-ui-select?

Issue #438, opened 2014-11-25
Material design theme


jimmywarting:
Based on [angular material](https://github.com/angular/material) could we make a new theme that behaves and feels as the same as it would on android lollipop? basically you get a modal dialog with checkboxes with a<br> `<select multiple><optgroup ...` like this video I made https://dl.dropboxusercontent.com/u/3464804/ilos%20android%202014-11-25%2021-45-43.mp4 

And when its not multiple perhaps show a dialog anyway but with radio buttons instead of checkboxes?

Not only would it solve the overflow hidden (#41) problem but also make it responsive. 
I have made a concept here without any ui-select... not 100% complete, but anyway... https://dl.dropboxusercontent.com/u/3464804/material/index.html

tolemac:
+1 nice

photusenigma:
+1 for material design theme

iamryandrake:
Any updates on this for themes?

amcdnl:
This is not a priority item for the core team.  PR's welcome.

Angular material has a select and there is also this one: http://ui.lumapps.com/directives/selects

YousefMMS:
I just need to +1 the material theme :)

jacek213:
+1

fujuca510:
+1

user378230:
Please do not +1 issues, it adds uncessary noise to the issue. You can use github reactions to voice your support. 

Note as previously mentioned, unless this is contributed in via a PR its unlikely to become part of the library. Its not a major priority.

PR #684, opened 2015-02-19
chore(refactor): split code into multiple files

dimirc:
I think it's time to split the code into different files to get more organized. Basically I split each main directive and service into different files and did a small change on the gulpfile to concat them when building.

I guess that this split might also help to avoid people pushing changes to /dist/select.js file and do the PR on the specific /src files

This is how it will look:
image

Later we should also reorganize the themes folders (check https://github.com/angular-ui/ui-select/issues/186)

Let me know your opinion

Related to: https://github.com/angular-ui/ui-select/issues/649


dimirc:
@brianfeister @JeffGreat @starf @homerjam sorry to spam you, but as I see you active on this repo, can you give me your opinion on this PR? I know this will break most PRs that aren't merged but I think this will help to start organizing better all the project.

PD: This PR is failing because I merged https://github.com/angular-ui/ui-select/pull/651 (support for sort for multiselect) today and I should becase and create a new file uiSelectSortDirective.js

dimirc:
I rebased this PR to last master, with a new file for sort directive:
image

starf:
Don't count me as active, just a user trying to fix some of my issues while migrating from ui-select2. Looks to me to be much more manageable & readable like this.

homerjam:
I guess I'm newly active! I think splitting up the files will certainly make things easier

brianfeister:
This looks good to me @dimirc.  Would be a great time to think about how we generate distributions, whether the theme files can be broken out per the discussions in #175, #186, #529... obviously the scope of that is bigger, however.  My company just made the decision to move to Angular Material over Bootstrap, so it would be great to both include that as a new theme and also come up with a strategy for breaking out dependencies and having a way to reference different CSS and angular templates per theme #438

PR #686, opened 2015-02-19
Possibility to change path to template files from outside.

MartinMuzatko:
The problem is, that changing the theme to augment the path (e.g.: partials/ui-select/bootstrap) doesn't work, since the theme is used somewhere else. Which indicates, that there needs to be an additional path parameter, in order to adapt the library to a custom subfolder, especially for bigger frameworks.

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -89,7 +89,7 @@
 
         .constant('uiSelectConfig', {
             theme: 'bootstrap',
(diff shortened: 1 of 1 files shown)


wesleycho:
Closing due to age - feel free to open a new PR against current `master` with passing tests.

Issue #302, opened 2014-10-14
Provide option to add no-value choice


just-boris:
Angular.js [select directive](https://docs.angularjs.org/api/ng/directive/select) have a convenient way to define null-value option. If you put `<option>` tag without value inside `<select>`, it will be treated as null-option which sets model value to `null`.
It is useful sometimes to provide user ability to reset selected value. Now I have to implicitly add that item to other choices in repeat.

```js
$scope.persons = [{"id": null, "name": "No person"}].concat(otherPersons)
```
```html
<ui-select ng-model="form.person">
    <ui-select-match placeholder="Select person...">{{$select.selected.name}}</ui-select-match>
    <ui-select-choices repeat="person.id as person in persons">{{person.title}}</ui-select-choices>
</ui-select>
```
I don't want to do it and it would be good if you will have an option for provide this option through html, for example
```html
<ui-select ng-model="form.person">
    <ui-select-match placeholder="Select person...">{{$select.selected.name}}</ui-select-match>
    <ui-select-null-choice>No person</ui-select-null-choice>
    <ui-select-choices repeat="person.id as person in persons">{{person.title}}</ui-select-choices>
</ui-select>
```
It is similar to select directive logic and looks good for me. What do you think?

jasny:
+1

fantasia949:
+1

Anton-Nikishkin:
+1

akirich:
+1

veob:
+1

kunashir:
+1

Blarm1959:
+1

dlukez:
Guys, check this out and tell me if it suits your needs... http://plnkr.co/edit/7nmBRicsmH33E55sSomz

Reklino:
That solution looks great to me. In my opinion, it would be best if the null option is not filterable. That way there is always a way to set the box back to null.

akirich:
absolutely you are right
null option should be always selectable

dlukez:
That's how I had it working in the first place but the requirement for my project dictated that the option should be filterable...

dlukez:
Just checked ngOptions -> it's not filterable. Maybe we mimic that behaviour?

Reklino:
Maybe there should be an option for the null option to filterable or not? Something like 'null-filterable="true"'. That may be a bit much though...

dlukez:
[...]

PR #687, opened 2015-02-19
isEmpty also true when selected == 0

starf:
seems kind of logically
greetings

### DIFF
--- a/src/select.js
+++ b/src/select.js
@@ -195,7 +195,7 @@
     ctrl.$filter = $filter;
 
     ctrl.isEmpty = function() {
-      return angular.isUndefined(ctrl.selected) || ctrl.selected === null || ctrl.selected === '';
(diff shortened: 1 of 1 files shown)


src/select.js
```diff
@@ -195,7 +195,7 @@
     ctrl.$filter = $filter;
 
     ctrl.isEmpty = function() {
-      return angular.isUndefined(ctrl.selected) || ctrl.selected === null || ctrl.selected === '';
+      return angular.isUndefined(ctrl.selected) || ctrl.selected === null || ctrl.selected === '' || ctrl.selected === 0;
```
dimirc: Not sure if this will be ok since `0` could be a value, so it isn't actually empty. This function will make the placeholder appear even when the user select `0`as value

starf: I understand. Thanks. Probably not an issue anymore when #302 is implemented.

user378230:
I agree with dimirc comments. So closing.

PR #699, opened 2015-02-25
feat(choices): add null option support

### LINKED ISSUE #302: Provide option to add no-value choice

dlukez:
Needed this feature so decided to implement it. I toyed with a few different solutions and boiled it down to this (doesn't look like much, but it took me quite a while to get to this).

This solution requires that the "null" option is a typical item present in the array so that we have flexibility in filtering and displaying it without having to write a bunch of conditional statements. We then use the "null-option" directive on the "ui-select-choices" element to mark our null option. When the option is selected, the model is set to null (and vice versa).

An additional helper option is included - "loose-null" - which tells ui-select to treat undefined and null as the same. In this case, if the model value is undefined, the null option will still be displayed. Just add
"loose-null" on the "ui-select-choices" element.

See http://plnkr.co/edit/7nmBRicsmH33E55sSomz?p=preview.

Happy to discuss implementation specifics, thought I'd open this PR to get the ball rolling. Yet to write unit tests... Could anyone help with this?


dlukez:
Mm, may not have been as relevant to #302 as I thought after having another read. May take another look at this at some point. If we want to include the null-option in the filtering things might be tough since the filtering happens outside of ui-select itself.

dimirc:
If `nullOption` is used, then placeholder won't work or when will it be shown?

dlukez:
The placeholder would be shown if "loose-null" isn't used and the value is undefined. I agree that this relationship between placeholder/null-option should probably be thought about further.

dlukez:
Gonna close this for now to make sure it's not accidentally merged before it's ready. A discussion is still on going in #302 on this

PR #711, opened 2015-03-02
Patch to allow for complex trackBy expressions

voipme2:
This allows for complex trackBy expressions, such as `item.name + '/' + item.version`.  Previous versions would end up causing duplicates in the repeater directive, since the matches regex would end up as undefined.  This would effectively cause items to be tracked by only the first portion (in this case, `item.name`), resulting in duplicates.


wesleycho:
This needs squashing, then rebasing, and unit tests.

Jefiozie:
Closing it down as it is a old issue. Please create new issue if this is still a bug/ problem or you want this merged.

Thanks.

Issue #343, opened 2014-10-25
ui-select initialize fail when the ngModel value is over the limitTo filter


yixblog:
problem plunker:http://plnkr.co/edit/seYgCTj5okknDdl3RzBk?p=preview
when i set the ngModel value when initialize,and the ui-select repeat items has limitTo filter
if the initialize value is out of the top limit number then the name will not display

loliver:
If we add a limit property to the $select object (and controller etc), we can set it to an absurd number (I used 1,000,000) to remove the limit when doing ngModel.$formatters.unshift (alternatively, in 1.4.x we can set it to null and it'll return the full array).

In my testing the performance is still far better than when I was loading the ~1300 items in to the box and it functions correctly. I'll rewrite it in the src files and do a pull request and link here.

jordyvaneijk:
Hello I'm using v0.12.1 of UI-Select but this is still present. Is there a fix which i can use so i can use limitTo? because we have huge performance issues when having collections with over 500 items...

I hope someone can tell me more..

dilerarc:
+1 problem still exists

jordyvaneijk:
After some time  i decided to make my own limitTo... This is a very simple implementation. 
I used the standard limitTo coming witch angular and added my own items with a union before the limit was done. This returns list of items with you limit on it + the list with items specified. Which will be filtered out by UI-select control because they are already in the list.

dilerarc:
@jordyvaneijk Could you provide a piece of code please? I don't quite understand your solution.

dilerarc:
@jordyvaneijk Oh, i got it, thanks man!

dilerarc:
@jordyvaneijk But now I have an exception while searching items in my select
`[ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: o in $select.items, Duplicate key: undefined:undefined, Duplicate value: undefined`

wesleycho:
This is already doable by using the `refresh` functionality.

Going to close this as a usage issue.

kristerd:
@wesleycho I don't understand how this is doable using the refresh functionality, can you please elaborate?

ankur8007:
Can you please help how to resolve this issue, I am facing similar issue with limitTo ?

voohoo:
I would also like help on this issue.

PR #716, opened 2015-03-04
Allow limitTo filter to return results correctly

loliver:
Fixing issue #343 by declaring a limit property within the controller that can be specified as an attribute on the ui-select element.

By default the limit is set to 1,000,000 as until 1.4 limitTo can't be ignored if the filter is invoked on an array, so it has to be an absurdly high number.

When ngModel.$formatters.unshift is called within the directive, the specified limit value is ignored and 1,000,000 used instead, the same way that $search is overwritten.

An example implementation would be to put limit="10" on the ui-select element, then repeat="item.Id as item in items | filter:$select.search | limitTo:$select.limit".


hugomosh:
+1

naltatis:
+1 works like a charm

alpha-cast:
Gah, just spent most of a day working this one out, even had a plunkr ready to go for an issue before I managed to find a link here.

Any idea when this is going to get merged? It's kind of killer if you have massive lists to display!

I've managed a work around by manually watching things and making the model update myself via an intermediate variable, but boy is it ugly to have to do.

alpha-cast:
Actually I just realised this is only specifically for the limitTo filter right? If any other filter cuts the results down then it'll still have the same problem?

I don't suppose there's a more general way to solve this problem? I guess not since filters can do both filtering and mapping and there's no general way to work out which is doing what. :-(

Would be great if there was an optional way to say "here's the full list of things you contain, regardless of how I'm currently filtering your options list".

alpha-cast:
The more I think about it, there's actually 2 different concerns here:
1. Map a list of all available options. This lets the select track external changes to its ng-model.
2. Filter the list of available options down while the user is typing in the select box.

Finding a way to separate these concerns seems like it would fix this problem generally. I've hit 2 cases (on the same select) where the filters I'm applying both filter in the empty query case, so they'll both stuff up the current "get all the options" code.

Something like this maybe, where if ui-all-choices="" is present it is used to get all the options. In this case I'm just binding to an intermediate scope variable and using it again in the ui-select-choices[repeat] expression to avoid having to do run any mapping twice.
```html
<ui-select ng-model="ctrl.selectedPerson"
           ui-all-choices="person in mappedPeople = (ctrl.peopleList | doSomeMapping)">
  <ui-select-match>{{$select.selected.name}}</ui-select-match>
  <ui-select-choices repeat="person in mappedPeople | filter : $select.search">
    {{person.name}}
  </ui-select-choices>
</ui-select>
```

jordyvaneijk:
Hello I'm using v0.12.1 of UI-Select but this is still present. Is there a fix which i can use so i can use limitTo? because we have huge performance issues when having collections with over 500 items...

I hope someone can tell me more..

dilerarc:
+1 problem still exists

wesleycho:
Can you squash the commits, rebase, then add a unit test?

lukaszszymanski:
+1 sill there

Jefiozie:
Please do not "+1" issues, use github 👍 reactions if you really must. Comments should be reserved for constructive discussion/issue resolution. +1s merely add unnecessary noise and scrolling between other, more helpful, replies.

Other ways you can actually help include:

Submitting a pull request that resolves the issue/adds the requested feature
Researching possible implementations or causes of bugs
Providing reproduction plunkrs if one isn't already available

Thanks! 😄

Jefiozie:
Closing it down as it is a very old issue. Please create new issue if this is still a bug/ problem or want this merged.

Thanks.

Issue #713, opened 2015-03-03
Z-Index problem with dropdown menu


cmlenz:
We are using a layout that has a table-like list of items, each containing a ui-select field. The items in this list need to be relatively positioned, and they need a z-index. This layout causes the ui-select dropdown menu to be rendered below the subsequent items in the list.

I have created a minimal example for this problem here:

  http://plnkr.co/2K8HuGAJ8EzJWlDaQJSh

As far as I'm aware, this is not a problem that ui-select itself can fix. However, we do need a way to handle this situation ourselves in our application, by increasing the z-index of the item with an open ui-select dropdown menu.

My first thought would be to add `on-open` and `on-close` events to the `ui-select` directive. Is that something that would be accepted (I can create a PR)? Or is there a better way to achieve what we need that I'm overlooking?

Thanks!

cmlenz:
After some more research this issue would be addressed by adding an `append-to-body` option as suggested in #41.

PR #718, opened 2015-03-04
Add an append-to-body attribute for improved layering.

cmlenz:
Add an `append-to-body` attribute to the `<ui-select-choices>` directive that moves the dropdown element to the end of the document body before opening it, thereby solving problems with the dropdown being displayed below elements that follow the `<ui-select>` element in the document. This implementation is modeled after the `typeahead-append-to-body` support from UI Bootstrap. See #41 (and quite a few dupes).

I have only tested the Bootstrap theme.

Here's a Plunk showing an example:

  http://plnkr.co/cwVLBfI6Ebo5oLj15QWd

(Based on my Plunk in #713 which shows the broken behavior without `append-to-body`).


leipert:
+1

maxavi:
+ 1

src/bootstrap/choices.tpl.html
```diff
@@ -1,6 +1,7 @@
 <ul class="ui-select-choices ui-select-choices-content dropdown-menu"
+    style="display: block"
     role="listbox"
-    ng-show="$select.items.length > 0">
+    ng-show="$select.open">
```
brianfeister: Can you explain why this was changed? We have so many nuanced use cases for this project, I worry this could break the template for one of the many in our large array.

src/common.css
```diff
@@ -173,21 +173,21 @@
...
     color: #fff;
     text-decoration: none;
     outline: 0;
     background-color: #428bca;
 }
 
-.ui-select-bootstrap .ui-select-choices-row.disabled>a,
-.ui-select-bootstrap .ui-select-choices-row.active.disabled>a {
+.ui-select-choices-row.disabled>a,
+.ui-select-choices-row.active.disabled>a {
```
brianfeister: You mentioned in your PR that you only tested Bootstrap, removing the CSS class namespace would very likely have impact on the other themes, so please test them as well, seems likely this could cause a regression.

src/common.js
```diff
@@ -133,5 +133,25 @@ var uis = angular.module('ui.select', [])
...
+  return function(element) {
+    var boundingClientRect = element[0].getBoundingClientRect();
+    return {
+      width: boundingClientRect.width || element.prop('offsetWidth'),
+      height: boundingClientRect.height || element.prop('offsetHeight'),
+      top: boundingClientRect.top + ($window.pageYOffset || $document[0].documentElement.scrollTop),
+      left: boundingClientRect.left + ($window.pageXOffset || $document[0].documentElement.scrollLeft)
+    };
+  };
+}]);
```
brianfeister: thanks for keeping this lightweight :+1:

src/select2/choices.tpl.html
```diff
@@ -1,4 +1,5 @@
-<ul class="ui-select-choices ui-select-choices-content select2-results">
+<ul class="ui-select-choices ui-select-choices-content select2-results"
+    ng-style="{top: position.top+'px', left: position.left+'px'}">
```
brianfeister: I had the impression that you were making this directive "opt-in", meaning that it only impacts the directive if invoked.  This gives me the opposite impression, can you explain a bit for me?

src/selectize/choices.tpl.html
```diff
@@ -1,4 +1,5 @@
-<div ng-show="$select.open" class="ui-select-choices selectize-dropdown single">
+<div ng-show="$select.open" class="ui-select-choices selectize-dropdown single"
+     ng-style="{top: position.top+'px', left: position.left+'px'}">
```
brianfeister: same question.

brianfeister:
Thanks for this @cmlenz!

My main concern is that I've previously seen issues where select2, due to it's attaching elements to the `<body>`, will leave DOM artifacts when a user routes away (in a single page application) while the dropdown is open.  Can you take the demo from your plunk and add it as a static html demo in `/demos` for me (just copy one of the existing demos and add your code)?  Also, can you create a minimal (simple as you can manage) example of a link that I can click to route away from the current route and attempt to reproduce the bug I'm describing?  I'm thinking it won't be an issue with the code as you've written it, but I want to have it there permanently as means of verifying that we don't introduce that bug into this library.

src/bootstrap/choices.tpl.html
cmlenz: I'll back that out, it isn't needed, and results in empty dropdown appearing with the Bootstrap theme.

src/select2/choices.tpl.html
cmlenz: I'll back that out, it was a leftover from earlier experimentation. The styles are set in the `positionDropdown()` function.

cmlenz:
Thanks for the detailed review @brianfeister. Good point about the dropdown hanging around after the select is gone. For demonstration purposes, wrapping the select in an `ng-if` should do the trick. Here's a Plunk that shows that the current PR **does** have this problem: http://plnkr.co/cwVLBfI6Ebo5oLj15QWd

That should be easy to fix by removing the dropdown on `$destroy`. I'll work through this and the other issues and post an updated PR later today.

brianfeister:
Thanks! Also I meant /examples not /demos :P

cmlenz:
I've noticed that the chosen approach just does not work with the Select2 theme. The problem is that I expected the `append-to-body` option to only apply to the choices element, and thus moving it to the body would suffice. However, the Select2 theme displays the input, search field, and dropdown all as one unified element. Moving only the choices element to the body does not help there.

I'm currently playing with an alternative approach that moves the whole select element to the body when it is open.

cmlenz:
#736 has a simpler patch that also supports the Select2 theme.

PR #727, opened 2015-03-06
Update select.js

kcrabtree1225:
Add handling to close the dropdown when it is clicked, as opposed to toggling only when the user clicks on the caret.


dimirc:
I did this [change](https://github.com/angular-ui/ui-select/commit/ed610774a53bc295008b71eba9937fd5eacf9bb4) that should achieve the same.

Check [plunker](http://plnkr.co/edit/GG7YE90E2yZXO2FR9cFE?p=preview)

PR #728, opened 2015-03-07
feat(focus): allow to set focus when loaded or listening a scope event

dimirc:
We could have 2 ways to set focus:

1. Attribute `autofocus` to automatically get focus when loaded
2. Attribute `focus-on` to define a scope event name to listen (e.g. `focus-on='SomeEventName'`)

Demo [plunker](http://plnkr.co/edit/enUFAvPO4NKlXT8aJvnD?p=preview)

I wanted to have this implementation on directives but to use autofocus/focus-on names could create conflict with other directives since the names will be somehow generic, so I'm just checking for these attributes and append the corresponding directives (with more specific names) if detected on compile fn.

To understand better the changes, will be cleaner to check diff without whitespaces
https://github.com/angular-ui/ui-select/pull/728/files?w=1


kimardenmiller:
Functionality looks awesome :+1: 

Can't wait to have it.

dimirc:
I'm dropping the implementation via directives for now, and just adding to uiSelect link fn

New plunker [demo](http://plnkr.co/edit/iH5eGi0voWItvxkIvJkh?p=preview)

dimirc:
[v0.11.0](https://github.com/angular-ui/ui-select/releases/tag/v0.11.0) released

piernik:
In multi on on-focus event it sets focus on input but not open suggestions. Can You open suggestion (as on `click` event)?

kimardenmiller:
@dimirc working great. Our application has a small issue: We're not using any outline on focus:
    ```  .btn-default-focus {outline: none;}```
So the user can't tell it's focused. Other fields give a blinking curser on focus, but ```ui-select``` does not.

May be the same issue as @piernik's?

Does one need to trigger an activate() to get the curser? Is it doable with your code as written?

PR #731, opened 2015-03-09
chore(refactor): split code into multiple directives

dimirc:
Recently we started organizing the library on https://github.com/angular-ui/ui-select/pull/684 when code was separated into multiple files, but I still see too much  conditional code to check if the component is on single/multiple selection mode. So I think that will be better to split the code into 2 different directives, `uisSingle` and `uisMultiple` respectively that will internally be appended to the element automatically on main `uiSelect` compile fn, depending on the presence or absence of `multiple` attribute.

PD: the public API wont change, its just internal

New files created:
 * uisSingleDirective.js
 * uisMultipleDirective.js

### DIFF
--- a/src/bootstrap/match-multiple.tpl.html
+++ b/src/bootstrap/match-multiple.tpl.html
@@ -5,10 +5,10 @@
       tabindex="-1"
       type="button"
       ng-disabled="$select.disabled"
-      ng-click="$select.activeMatchIndex = $index;"
-      ng-class="{'btn-primary':$select.activeMatchIndex === $index, 'select-locked':$select.isLocked(this, $index)}"
+      ng-click="$selectMultiple.activeMatchIndex = $index;"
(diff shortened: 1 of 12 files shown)


dimirc:
I believe this PR is looking better. Maybe just a couple more changes and will be ready to merge, since I'd like to avoid people to submit new PRs under the current organization on master and then be forced to  rebase

src/uisMultipleDirective.js
```diff
@@ -0,0 +1,393 @@
+uis.directive('uisMultiple', ['uiSelectMinErr','$timeout', function(uiSelectMinErr, $timeout) {
```
onlywei: Is this really the directive name you want?

    <uis-multiple>

Aren't you afraid of the confusion it will cause by being in a totally different namespace from the other directives in this same library?

dimirc: The prefix `uis` is already used on some filenames/directives of this library to keep some names shorter, but we could also consider `uiSelectMultiple`and `uiSelectSingle` for this new directives if if thats cleaner.

PR #735, opened 2015-03-09
Fix for sizing bug introduced by #718

cmlenz:
The search input size was being calculated incorrectly for non-multiple select instances using the Bootstrap theme. Remedy by only calling the sizing function when the select is in multi mode, which is how this function is guarded on other call sites as well.

Check out the Bootstrap select field on the `demo.html` page to see the problem in action.

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -211,8 +211,10 @@ uis.directive('uiSelect',
       attrs.$observe('disabled', function() {
         // No need to use $eval() (thanks to ng-disabled) since we already get a boolean instead of a string
         $select.disabled = attrs.disabled !== undefined ? attrs.disabled : false;
-        // As the search input field may now become visible, it may be necessary to recompute its size
-        $select.sizeSearchInput();
+        if ($select.multiple) {
+          // As the search input field may now become visible, it may be necessary to recompute its size
+          $select.sizeSearchInput();
+        }
       });
 
       attrs.$observe('resetSearchInput', function() {


dimirc:
[v0.11.1](https://github.com/angular-ui/ui-select/releases/tag/v0.11.1) released

cmlenz:
Wow that was quick. Thanks!

igorlino:
Using the latest version from UI-Select shows a search input size problem. When opening for the 1st time, it takes the style="width:21px"
only after click over the icon, goes to a normal width, filling the dialog. (but again, using some magical number)

Following workaround code fixes for me at the moment:
   .select2-search > .ui-select-search.select2-input {
       width: 100% !important;
   }

cmlenz:
@igorlino can you provide an example where this can be reproduced?

igorlino:
@cmlenz that's a bit hard, I use ui-select in several places with different model bindings. I'm using the very latest versions. "bootstrap": "~3.3.2", "angular-ui-bootstrap-bower": "~0.12.1", "angular-ui-select": "~0.11.1", "select2-bootstrap-css": "~1.4.6".

The this sizing problem appeared only after 0.11.0. If I revert back the previous version, the problem is gone.

cmlenz:
@igorlino I'd really need a Plunk or something where the problem is reproducible. I've tried quite a few different scenarios and they've all worked for me. Also, what browsers are you seeing this on?

igorlino:
@cmlenz I understand, I will see what can I do to reproduce it, its bit of a problem since i'm under time pressure for other reasons. The browser is Version 40.0.2214.115 m

dimirc:
FYI release v0.11.1 look like it was pointing to incorrect commit (https://github.com/angular-ui/ui-select/issues/742)

PR #738, opened 2015-03-09
Hidden Input Element Not Resizing Correctly

djdaniels90:
The problem I was having was that when adding an item to a multiple select, the hidden input element was being pushed down to a new line with 100% width (100% of the container). This ultimately made the UI I was making look ugly.

I examined the source code and found the area in question, responsible for this feature: ctrl.sideSearchInput. I located the problem in the auto resize feature which is not being executed properly. I was unable to execute any of the code within the $timeout function which I noticed was missing a time argument. $timeout(function(){}) vs $timeout(function(){}, 250).

Looking at the $timeout documentation, $timeout without a time parameter executes with default time 0. Given the purpose of this function, 'Give tags time to render correctly', this was not in fact giving time for the tags to render nor was it ever being executed in my uses of ui-select. I have about 30-40 of these which all experienced this problem. This solution resolved them for me.  Ultimately in all my tests it was as if $timeout without a time argument was not being executed rather than being executed with time 0ms.

This fix worked for me and had a huge effect on the look and feel of the ui.


dimirc:
Can you ase a plunker with a failing example?

wesleycho:
Closing due to incorrect fix - feel free to open a new issue with details & reproduction on the exact problem so a better fix can be implemented.

Issue #739, opened 2015-03-10
null values in ng-model after changing ng-model programmatically on multiselect


KKrisu:
**version** 0.11.1

**Demo** http://plnkr.co/edit/kLsFtY?p=preview

**settings** multiselect, complex ui-select-choices objects

**Issue**
When ng-model is empty and I'm programmatically changing ng-model values, it changes ng-model to null. When ng-model is not empty and I'm changing values programmatically it works fine. That's why when you click on the button on my demo second time (or you choose any value manually using UI), the values are set correctly.

**FIX**
PR: https://github.com/angular-ui/ui-select/pull/740

renanlecaro:
I got into some trouble when modifying the model outside of ui-select too 
http://plnkr.co/edit/Wboo1fiePaLHKX0eHjxI?p=preview
It looks like it's more of a one way binding than a 2 ways : adding / removing elements with ui-select works well, but the digest cycle of the directive is not always triggered.

z3ws:
I have the same proble than Kris Kesy

nascosto:
+1 We are having the same problem. Spent quite a bit of time tracking this down.

clydetacoma:
+1

lianyi:
+1

VoltiSubito:
+1

olso:
+1

lisaq-leafnode:
We are also having the same issue. When is this pull request likely to be merged in?

dimirc:
Should be fixed with #748

PR #740, opened 2015-03-10
[issue 739] Changing ng-model programmatically on multiselect.

KKrisu:
This prevents setting null values when changing ng-model from empty value to set programmatically.

Fixes: https://github.com/angular-ui/ui-select/issues/739
(there is a demo of issue in plunker)


dimirc:
@KKrisu check https://github.com/angular-ui/ui-select/pull/748

PR #741, opened 2015-03-10
fix dragleave triggered on children

0xF013:


dimirc:
What error/issue was this causing?

0xF013:
@dimirc in some cases dragging a choice over another causes the whole drag process to cancel out. Should I prepare some examples?

dimirc:
One example will be helpful to check this

0xF013:
@dimirc I am sorry, I might have introduced some side effects with this fix, mainly dragleave not being consistently fired. The problem I actually had got easily fixed with 
```
.select2-choices li * {
  pointer-events: none;
}
```
which was either not set or overwritten by some other css.

dimirc:
Do you mean this commit should better be reverted then?

0xF013:
@dimirc yes please, I fixed the issue without looking into the cause

PR #744, opened 2015-03-11
Updated Readme.md

bes89:
added bower install command to copy&paste


dimirc:
We do have this info but at https://github.com/angular-ui/ui-select/wiki/Getting-Started#installing meybe we could somehow promote more the wiki to avoid having 2 places with same info

bes89:
Come on! If you are an experienced developer, know what select2 or selectize is and find this repo you just want to get started (especially because of the demos!)
Better you move "Development" to the docs and place and install command in front.

Btw thank you for this great project! :+1:

PR #746, opened 2015-03-11
Remove 'disabled' attribute from DOM if root element is not <select>

### LINKED ISSUE #726: disabled="disabled" attribute on root <div> causes scroll problems in IE
Ui-select adds ``disabled="disabled"`` attribute on root ``<div>`` of select widgets. ``disabled`` attribute is not valid on divs, and IE (tested in 11.0.15 including 10 and 9 emulation mode) has a [bug preventing scrolling of page when mouse is over such elements](https://connect.microsoft.com/IE/feedback/details/1161951/non-input-elements-with-set-disabled-attribute-and-overflow-hidden-prevent-scroll-of-page).

Steps to reproduce:
* Open ["Demo"](http://plnkr.co/edit/a3KlK8dKH3wwiiksDSn2?p=preview) page in IE
* Resize browser window vertically until page don't fit into viewport
* Press "Disable ui-select" button
* Put mouse pointer over second (Select2 theme) or third (Selectize theme) select widget
* Try to scroll with mouse wheel
* Page will not scroll

You can see ``disabled="disabled"`` attribute on root ``<div>`` of each select widget.

P.S. ``disabled`` does not appear on plain select2 disabled elements without angular-ui ([demo page for select2](https://select2.github.io/examples.html#disabled)).

image

Tested also in Chrome (40.0.2214.115 m) and Firefox (38.0a2), ``disabled="disabled"`` appears too but not causes any problems.

kolen:
Fixes #726.
Not sure if it will not break anything. If there's no code that reads 'disabled' attribute directly from DOM to determine if select widget is disabled, everything should be right.

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -153,6 +153,12 @@ uis.directive('uiSelect',
           });
         }
 
+        // <div>s and other non-form elements should not have 'disabled' attribute and it causes problems in IE
+        // so if current theme uses something other than <select>, remove 'disabled' attribute from root element in DOM
+        if (element.prop('tagName') != "select") {
+          element.removeAttr('disabled');
+        }
+
         function onDocumentClick(e) {
           if (!$select.open) return; //Skip it if dropdown is close
 
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -546,6 +546,28 @@ describe('ui-select tests', function() {
     expect(isDropdownOpened(el3)).toEqual(true);
   });
 
+  it('should not have disabled attribute on div', function() {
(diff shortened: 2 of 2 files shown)


wesleycho:
Can you squash your commits then rebase against current `master`?

kolen:
Squashed but now tests don't work, investigating cause of this.

kolen:
For some reason, function returned from compile always don't get 'disabled' attribute in 'element' parameter. But then 'disabled' appears back, always. I have no idea how it works. Feel free to close and cancel this pull request, I gave up. Looked into Angular documentation today and it feels like quantum physics is easier than Angular.

PR #747, opened 2015-03-11
Watch the allowClear attribute for changes

rwlogel:
Currently allowClear is only evaluated during the linking phase.  If you want to be able to control this dynamically it needs to be watched for changes.


dimirc:
Can you create a for test for this?

rwlogel:
There doesn't appear to be any existing unit tests for the allow-clear attribute so I'm not sure where I should put one.  Do you have any suggestions?   Could you suggest another unit test that would be a good example to start from.

rwlogel:
Created unit tests.

rwlogel:
Where did the changes to common.css come from?  They seem to be from another pull request?

cmlenz:
@rwlogel Not sure how that got in here, but that change is from #745

See also commit list at the top of this PR:

screen shot 2015-03-13 at 10 14 52

PR #748, opened 2015-03-12
fix(multiple): formatter binding

dimirc:
- [x] Prevent $select.selected with incorrect values from formatter 
Closes https://github.com/angular-ui/ui-select/issues/739
Fixed [plunker](http://plnkr.co/edit/0IDbtL?p=preview)

- [x] If item not found on current available lists, add it directly to selected
- [x] Change comparison to angular.equals()
Closes https://github.com/angular-ui/ui-select/pull/410

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -5,10 +5,27 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
 
     controller: ['$scope','$timeout', function($scope, $timeout){
 
-      var ctrl = this;
-      var $select = $scope.$select;
+      var ctrl = this,
+          $select = $scope.$select,
+          ngModel;
+
+      //Wait for link fn to inject it 
+      $scope.$evalAsync(function(){ ngModel = $scope.ngModel; });
 
       ctrl.activeMatchIndex = -1;
+
+      ctrl.updateModel = function(){
+        ngModel.$setViewValue(Date.now()); //Set timestamp as a unique string to force changes
+        ctrl.refreshComponent();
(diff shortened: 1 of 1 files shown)


dimirc:
@renanlecaro [commented](https://github.com/angular-ui/ui-select/issues/739#issuecomment-78388012)
>I got into some trouble when modifying the model outside of ui-select too 
http://plnkr.co/edit/Wboo1fiePaLHKX0eHjxI?p=preview
It looks like it's more of a one way binding than a 2 ways : adding / removing elements with ui-select works well, but the digest cycle of the directive is not always triggered.

Also this PR should solve it now. Updated [plunker](http://plnkr.co/edit/rnoGd1z7GwfWYiJ9OiK3?p=preview)

KKrisu:
Looks good for me.
I think you can include unit test I wrote in PR https://github.com/angular-ui/ui-select/pull/740 which covers issue #739

renanlecaro:
Okay this is weird, It does work on the plunkr but not in my app, and that's because of the angular version. With the latest angular version, the ui-select list updates every two adds 

http://plnkr.co/edit/s4M6U89aXYMCDfCUlk0x?p=preview

(i modified the plunks to add unique tags to the list)

dimirc:
@renanlecaro I did some additional changes. Check updated [plunker]  (http://plnkr.co/edit/upcWuSsUahlpQqV144pz?p=preview) with angular v1.3.14

KKrisu:
@dimirc I think would be good to add unit test for issue #739, which I wrote in closed PR
https://github.com/angular-ui/ui-select/pull/740/files#diff-26736c4d2d5d1e4c66b6e602618cb367R1694

PR #749, opened 2015-03-12
Fix input size calculation

ibigbug:
Since AngularJS 1.3.1 or higher version changed the default behavior of
attrs.$observe(). If we dont't have a `ngDisabled` on uiSelect, the
initial calculation won't work in attrs.$observe('disabled')

Link: https://github.com/angular/angular.js/commit/531a8de72c439d8ddd064874bf364c00cedabb11


ibigbug:
The default value is set to `ctrl` not `attrs`. The problem is:
```js
attr.$observe('sth', function() {
  // if `sth` is undefined
  // this function won't trigger.
})
```
See the link I provided.

PR #754, opened 2015-03-12
Fix for selects not working in IE8

Josh68:
https://github.com/angular-ui/ui-select/issues/158, based on an observed (but maybe not well understood) code change by @careywalker. Basically, just move the ng-click/ng-mouseenter bindings from the LI (choices-row) to the contained DIV (choices-row-inner). So far, I haven't seen an affect on behavior in any browser, but this is with light manual testing.

All automated tests pass with these changes in place:

>gulp test
[15:05:36] Using gulpfile C:\web\sites\ui-select\gulpfile.js
[15:05:36] Starting 'clean'...
[15:05:36] Finished 'clean' after 14 ms
[15:05:36] Starting 'scripts'...
[15:05:36] Starting 'styles'...
[15:05:36] Finished 'styles' after 61 ms
[15:05:37] Finished 'scripts' after 1.01 s
[15:05:37] Starting 'build'...
[15:05:37] Finished 'build' after 14 μs
[15:05:37] Starting 'karma'...
[15:05:38] Finished 'karma' after 880 ms
[15:05:38] Starting 'test'...
[15:05:38] Finished 'test' after 5.63 μs
INFO [karma]: Karma v0.12.31 server started at http://localhost:9876/
INFO [launcher]: Starting browser Chrome
INFO [Chrome 41.0.2272 (Windows 7)]: Connected on socket gaw3nxxZW2Omee4BBX-X with id 17666700
Chrome 41.0.2272 (Windows 7): Executed 110 of 110 SUCCESS (2.483 secs / 2.465 secs)

### DIFF
--- a/src/uiSelectChoicesDirective.js
+++ b/src/uiSelectChoicesDirective.js
@@ -1,6 +1,6 @@
 uis.directive('uiSelectChoices',
-  ['uiSelectConfig', 'uisRepeatParser', 'uiSelectMinErr', '$compile',
-  function(uiSelectConfig, RepeatParser, uiSelectMinErr, $compile) {
+  ['uiSelectConfig', 'uisRepeatParser', 'uiSelectMinErr', '$compile', '$window',
+  function(uiSelectConfig, RepeatParser, uiSelectMinErr, $compile, $window) {
 
   return {
     restrict: 'EA',
@@ -40,7 +40,7 @@ uis.directive('uiSelectChoices',
 
         choices.attr('ng-repeat', RepeatParser.getNgRepeatExpression($select.parserResult.itemName, '$select.items', $select.parserResult.trackByExp, groupByExp))
             .attr('ng-if', '$select.open'); //Prevent unnecessary watches when dropdown is closed
-        if (document.addEventListener) {  //crude way to exclude IE8, specifically, which also cannot capture events
+        if ($window.document.addEventListener) {  //crude way to exclude IE8, specifically, which also cannot capture events
           choices.attr('ng-mouseenter', '$select.setActiveItem('+$select.parserResult.itemName +')')
               .attr('ng-click', '$select.select(' + $select.parserResult.itemName + ',false,$event)');
         }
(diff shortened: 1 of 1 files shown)


Josh68:
I'm curious if there's any chance this request might make it into a build. Thanks.

dimirc:
I tried this on IE but when pressing arrow UP it isn't working, also on mouseenter is only working if the pointer is over a choice lower that current but not when going up.

PD: I tried this in IE10 with Browser Mode IE8

Josh68:
Hmmm, so do you have any ideas how to fix this issue for IE8? I can tell you that I can even move the mouseenter attribute back to the parent (.ui-select-choices-row), and click-to-select behavior still works again in IE8). I think the basic issue is IE8s assessment of whether the elements you're trying to bind the click event to are even visible, according to its interpretation. The original error I'm getting on click is "Can't move focus to the control because it is invisible, not enabled, or of a type that does not accept the focus." If I could figure out where to do it, adding an extremely small timeout (like 10ms) on the select might actually also fix things. (See, eg, https://github.com/jquery/jquery-mobile/issues/3505).

Also, do you know that the proposed code changes are causing the undesired behavior you're seeing? I'm not sure how to test arrow and mouseenter behavior, because they don't even seem to work for me with master code. In FF, for example, I'm seeing erratic key and mouseenter behavior - basically, it doesn't appear to work anyway. What I'm after in IE8 is just basic mouse click support. Any suggestions are appreciated. Thanks.

Josh68:
I have studied this issue a little more and here's my conclusion so far.

1) The issue appears to be that you're adding event bindings to an element that's actually up the DOM tree from the target, and IE8 does not do event catpuring

2) Inspecting in Chrome and putting a breakpoint on the beginning ctrl.select event handler (line 492-493), the actual event target is indeed the div.ui-select-choices-row-inner, not the parent .ui-select-choices-row

3) The handlers all work on the $select object passed into the link function within the uiSelectChoices directive, so I'm not sure how changing the event bindings would make a difference other than allowing IE8 to respond correctly to click and mouseenter events, unless the $event passed into the ctrl.select click-handler. Seems there could be a way to mitigate that in the handler.

Any comments or feedback are welcome, and thanks for looking into this.

Josh68:
I've come back to this request, reverting my previous changes and putting the same code in conditional blocks that check for document.addEventListener support. Basically, this should change functionality only for IE8 (or lower). I've tested in Chrome, FF, IE8, and IE11, plus IE11 in IE9 compatibility mode, and only in IE8 (so far) do the event binding attributes get put on the .ui-select-choices-row-inner, where at least clicks now register correctly (because it is the event target, not the target's parent node).

Thanks for continuing to consider this request.

Josh68:
Hello. Is there any chance a fix will be made for this bug, or is IE8 support being dropped?

PR #761, opened 2015-03-16
allow cancellation of selection

### LINKED ISSUE #760: cancel selection
How can I cancel item selection?

I have some items in my dropdown that I need to respond to, however I don't want them to change the currently selected value. 

Probably the best way would be to cancel selection when on-select callback returns false, would this be easy to add?

parliament718:
allow cancel selection by returning false from on-select callback closes #760

### DIFF
--- a/dist/select.js
+++ b/dist/select.js
@@ -537,19 +537,23 @@ uis.controller('uiSelectCtrl',
         var locals = {};
         locals[ctrl.parserResult.itemName] = item;
 
-        if(ctrl.multiple) {
-          ctrl.selected.push(item);
-          ctrl.sizeSearchInput();
-        } else {
-          ctrl.selected = item;
-        }
-
         $timeout(function(){
-          ctrl.onSelectCallback($scope, {
+          var setValue = ctrl.onSelectCallback($scope, {
             $item: item,
             $model: ctrl.parserResult.modelMapper($scope, locals)
           });
+          
+          if(setValue) {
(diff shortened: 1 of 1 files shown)


dimirc:
Pls dont change /dist files but only /src

parliament718:
sorry don't really know what I'm doing yet here. When I looked at src the code didn't even look the same?
https://github.com/angular-ui/ui-select/blob/master/src/uiSelectController.js#L272

Can this fix be incorporated somehow?

dimirc:
Maybe we could move the [broadcast](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectController.js#L267) after the onSelectCallback but inside the $timeout

parliament718:
Should I do that?

allenm:
I also need this , expect this PR to be merged .

qcmatsuoka:
+1

martinoss:
+1, need this too for tagging. I need to be able to cancel an insertion of an invalid value, either by the tagging function or on-selection.

wesleycho:
Remove the changes to `dist`, and make the changes to the appropriate files in `src`.

Jefiozie:
If i'm correct this PR can be closed as the original issue has been closed. If not please verify with latest version and make a new issue/PR.

Thanks

cyrilchapon:
This is not ok.
Issue has been closed without been resolved.

PR #772, opened 2015-03-18
Disable within fieldset

AdamAxtmann:
disable ui-selects when inside a disabled fieldset


src/uiSelectDirective.js
```diff
@@ -70,7 +70,18 @@ uis.directive('uiSelect',
...
+          $select.disabledFlag = attrs.disabled !== undefined ? attrs.disabled : false;
+          $select.disabled = $select.disabledFlag || $select.disabledSelector;
+        });
+
+        // When inside a disabled fieldset, disabledWatch element will be disabled
+        var disabledWatch = angular.element("<input type='hidden' class='ui-select-disable-watch' aria-label=''></input>");
+        element.append(disabledWatch);
+
+        // Watch the disabled flag on disabledWatch, and set $select.disabled.
+        scope.$watch(function () { return disabledWatch.is(":disabled"); }, function (value) {
```
dimirc: .is() isn't part of [angular.element](https://docs.angularjs.org/api/ng/function/angular.element) so we'll be creating a dependency on jquery to make this work which we should avoid

rwlogel:
What if we did this instead:

```javascript
if(!!disabledWatch.is) {
  return disabledWatch.is(':disabled');  // Use jquery if available
}
return !!element[0].querySelector('.ui-select-disabled-watch:disabled');
```

I tested in in Chrome and Firefox and it worked fine.  It doesn't seem to work in IE 11, but neither do fieldsets.

Also maybe the unit tests shouldn't use jquery if dependency on jquery should be avoided.

AdamAxtmann:
@rwlogel That seems to work. I ran the tests on my machine and updated the pull request with the change.

AdamAxtmann:
hello?

rwlogel:
Any ETA for this merge?

Sheeryada:
Could any one update on this merge please?

wesleycho:
This needs history cleaned up - there should be no revert commit.

cwagner22:
+1

katemihalikova:
What about merging this via squash merge?

luzrafaelf:
+1

Jefiozie:
Hi, I think people see what needs to be done to get this merged. Somebody maybe can make a new PR for this?

luzrafaelf:
+1

PR #773, opened 2015-03-18
Added directive to use custom compare function for taging

krmarien:
By default angular.equals is used, but sometimes properties of a newly created tag should be ignored. With a custom equals function this can be accomplished.


wesleycho:
Closing due to history needing work (there should be no merge commit).

In addition, this needs unit tests.

Issue #267, opened 2014-10-07
Support for adding separate templating to group heading elements


alsoscotland:
It would be nice to have the ability to independently template group headings,
or at least style group headers uniquely based on a class that is applied dynamically base on item property values

swissmanu:
i'm gonna try to contribute a solution to this issue. stay tuned :)

swissmanu:
alright. PR #796 is opened. hope you like the way how it's implemented.
otherwise any feedback is welcome :)

PR #796, opened 2015-03-24
Implemantation of Issue #267: Custom Template for Group Labels

swissmanu:
- new directive `uiSelectGroup` for customizing the item groups header template.
- if no custom group template is given, a default fallback to $group.name is used instead.
- see `examples/demo-groupby.html` for further explanations


swissmanu:
any interest in this?

Envek:
I have interest on this, thank you very much for your work — I currently pulled it on my fork and playing with it. IMO, this is must have feature.

I've found one issue: for some reason `ng-repeat` directive does not work inside of `ui-select-group` directive (yeah, I'm crazy and I need to have group labels to be huge). And I do not know why.

Envek:
Here is an example that illustrates problem with ng-repeat: http://plnkr.co/3fGeRhzjXl1CYHZhcDcO

By the way it's a nice example of custom templates for group labels, I think.

wesleycho:
This needs rebasing, as well as unit tests.

PR #799, opened 2015-03-25
Don't tab if 'TAB' is specified as a taggingToken

tmanderson:
Prevent default behavior of TAB key if TAB is a specified taggingToken. This may have to be enabled for multiple tagging select fields only.


wesleycho:
This needs a unit test.

PR #811, opened 2015-03-31
Fix(bootstrap): Prevents text overflow. Fixes #689.

### LINKED ISSUE #689: Bootstrap theme doesn't cut off text past border
Hello, I'm using the ui-select with the bootstrap theme, but when I select an option that has an text past the input box size, the text doesn't cut off like the one in select2 theme.

cutoff

kgeorgiou:
Fix related to Issue #689.


rapmue:
thanks, saved my day.

kgeorgiou:
No problem. I closed the pull request though because I realized it doesn't fully work when you have some inline elements like ``<span>foo</span><span>goo</span>`` as the contents of the ``ui-select-match``. I'll find a solution and reopen.

PR #814, opened 2015-04-01
Fix for no element selected.

priyanshuch:


wesleycho:
I believe this has been superceded and fixed in `master` - closing as addressed.

PR #821, opened 2015-04-02
Fix cleaning selected data on enter

gregurco:
I used this examples and found that if you select something by press enter (not by mouse) , then ng-click from trash button is called, because by default the type of the button is "submit". I think, that this small fix can save time of people, who will use this examples.


brianfeister:
Please provide a plunkr / codepen demo of the incorrect behavior you're describing.

gregurco:
http://plnkr.co/edit/kOwtJVAZFTIQiNve6dHW?p=preview - there you can find two forms (first with issue and the second without). Tru to select option in the selector with keyboard + enter.

wesleycho:
Can you squash these commits? (`git rebase -i HEAD~3` and mark all with `f`)

Issue #356, opened 2014-10-30
Set sort order with group by?


lapidus:
I want to group countries by "type" but it doesn't sort these option groups alphabetically.
How can I set a sort order of the groups created?

```
<ui-select-choices group-by="'type'" repeat="country in countryList | propsFilter: {name: $select.search}">
```

Thanks!

burzum:
+1

netstyler:
+1

Vitaliy-Yarovuy:
+1

Vitaliy-Yarovuy:
i have create pull request for that
https://github.com/angular-ui/ui-select/pull/840

brianfeister:
This was merged at #969

PR #840, opened 2015-04-10
add "group filter" feature

Vitaliy-Yarovuy:
i my case i need to change default order of groups.
In this realisation order can be set by array of groupName or function "groupsFilter" ( work like group-by)


Vitaliy-Yarovuy:
resolve the same issues as #356

brianfeister:
Can you please add an example somewhere in `/examples`?

Vitaliy-Yarovuy:
done

brianfeister:
Ah, your changes are in the `/dist` directory.  These files are only changed when we push a new release so we'll need to make sure there are no changes in `/dist`

Vitaliy-Yarovuy:
but if i not change dist, the demo will not work,
so i need to revert changes in dist ?

brianfeister:
Yup, it will work after we merge, I will build dist.  Sorry for the workflow, it's confusing, I agree and I know.

brianfeister:
This was merged at #969

Issue #828, opened 2015-04-06
Is there a standard approach to adding Select or Clear All options?


brycepj:
In the list of choices, is there a way to add an option that adds/removes all options to/from the selected group? Has anyone implemented something like this? 

I suppose I could probably hack around this with `ng-change` -- I'm just wondering if there's a better way.

user378230:
Clear all has been implemented in more recent versions of the library. 

To simulate select all, simply add a button to add all the options to your ng-model array.

PR #844, opened 2015-04-13
Implement 'select all' and 'reset all' functionality

b1rdex:
I need a way to select all choices and reset all selected choices for each multiple select to be available.
Same was requested in #828.
No pulls are found, so I did one myself.


b1rdex:
This pull request doesn't contain `dist`, so you must `gulp build` first to use it.

b1rdex:
https://github.com/b1rdex/ui-select/tree/features/select-and-reset-all-with-dist have built version to use (with bower for example).

b1rdex:
This thing is buggy, dunno what's wrong.

PR #861, opened 2015-04-16
When search is disable, the select, if cliccked, must display the selected item

dometec:


wesleycho:
This seems related to #453, and PR #595.

Issue #872, opened 2015-04-21
Wrong input focus when angular DebugInfo is disabled


piebe:
The plunker below contains the standard ui-select example, but with angular 1.3 and DebugInfo disabled, ie

$compileProvider.debugInfoEnabled(false);

Use the mouse to select the first selectbox, then select the second selectbox and start typing. The result will appear in the first select instead of the second. 

http://plnkr.co/edit/aRORZjU8VKhKuJW1pTxx

Most likely cause line 149 in uiSelectDirective:
var targetScope = angular.element(e.target).scope(); //To check if target is other ui-select
            
.scope() is not available when DebugInfo is turned off...

ckniffen:
+1

awerlang:
+1

PR #873, opened 2015-04-21
fix Wrong input focus when angular DebugInfo is disabled

piebe:
fix for Wrong input focus when angular DebugInfo is disabled issue #872


wesleycho:
Closing as this looks already fixed in `master`: https://github.com/angular-ui/ui-select/blob/d5e30fb1227678348f7435d9864356ed0f75f690/src/uiSelectDirective.js#L174-L175

PR #878, opened 2015-04-22
Fix width calculation to allow for different padding

cdjackson:
This PR uses the offsetLeft parameter to calculate the input size. It makes the assumption that offsetLeft (which is an available member) is the same as the right side, and then adds some margin.

This allows for different padding to be used, which is what the bootstrap themes do.

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -344,7 +344,7 @@ uis.controller('uiSelectCtrl',
           if (containerWidth === 0) {
             return false;
           }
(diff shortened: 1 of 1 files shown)


wesleycho:
This needs history fixed - there should not be any merge commits.

Can you fix the history and package it into one commit?

cdjackson:
Tidied up into single commit...

user378230:
An extra semi-colon is failing the build :disappointed:

Issue #709, opened 2015-03-02
Track by with single property binding fails


AntiFreez:
http://plnkr.co/edit/kp1QFUFCCgdvoVfrLxVv?p=preview

kether667:
+1 it seems this was introduced in 0.10.0 or 0.9.9

adematte:
It works fine in the version I am using: 0.9.0.

abobwhite:
This did work but seems to not work anymore - I am using 0.9.9. This is a big issue.

chrislovecnm:
:+1:  One this. I have been fighting this and already selected values in the select boxes ... Time to roll back to previous version.  This is broken in the latest version :frowning:

bbreuil-harris:
+1

adematte:
This is still broken in 0.13.1. Any updates on a fix/workaround?

wesleycho:
Closing this as a duplicate of a couple of issues.

PR #882, opened 2015-04-23
Fix multiple select issue with single property binding and track by expression 

AntiFreez:
This fixes https://github.com/angular-ui/ui-select/issues/709


brianfeister:
Thanks, need to take a look.

brianfeister:
Sorry for the extremely late review but there is a disconnect here.  I'm hesitant to remove this because this directive is doing so many things it must be there for a reason.  Re-assigning to @dimirc as he's more likely to have prior knowledge about why the `$select.parserResult.trackByExp` was in there.

ckniffen:
This is also a feature I need.

rafasoares:
Any updates on this?

AntiFreez:
Unfortunately, no.

user378230:
Closing as something [similar](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectMultipleDirective.js#L119) appears to have made its way in to master

PR #895, opened 2015-04-29
Remove confusing ng-disabled="disabled" in demos

caillou:
For some reason, a lot of the code in the demos contains the following attrubute:

```
<ui-select ng-disabled="disabled" …>
```

`ng-disabled="disabled"` doesn't have any effect, as it expects a `boolean`. This makes the demos confusing.

### DIFF
--- a/examples/demo-append-to-body.html
+++ b/examples/demo-append-to-body.html
@@ -83,7 +83,6 @@
     <p>Selected: {{address.selected.formatted_address}}</p>
     <ui-select ng-model="address.selected"
                theme="bootstrap"
-               ng-disabled="disabled"
                reset-search-input="false"
                style="width: 300px;"
                title="Choose an address"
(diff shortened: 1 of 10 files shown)


brianfeister:
Thanks for helping clean up. I didn't add the `ng-disabled="disabled"` bit in there. Can you give me a codepen / plunkr that demonstrates that there is indeed no difference between the two?

joedavison:
It is semi-confusing, but I think it needs to be there to support the "Enable ui-select" and "Disable ui-select" buttons, e.g.:

index.html:

```
  <button class="btn btn-default btn-xs" ng-click="enable()">Enable ui-select</button>
  <button class="btn btn-default btn-xs" ng-click="disable()">Disable ui-select</button>
```

demo.js: 

```
app.controller('DemoCtrl', function($scope, $http) {
  $scope.disabled = undefined;

  $scope.enable = function() {
    $scope.disabled = false;
  };

  $scope.disable = function() {
    $scope.disabled = true;
  };

...
}
```

joedavison:
It looks like this might be a holdover from the "Demo" linked from the README:

http://plnkr.co/edit/a3KlK8dKH3wwiiksDSn2?p=preview

brianfeister:
Thanks @joedavison, that sounds right. I'm going to close this in light of your comment. @caillou feel free to re-open with demo / further explanation, etc.

Issue #877, opened 2015-04-22
Typing fast in firefox skips the second character


piebe:
Go to the example pluncker. Tab to the second field. Start typing fast. If you can type fast enough the second character will not be there. 

This is only the case the first time you start typing in the field. After tabbing away and back again, sometimes the effect van be reproduced.

piebe:
See pull request #896

PR #896, opened 2015-04-30
Fix: issue #877 Typing fast in firefox skips the second character

piebe:
Changing the focus on elements does not always go fast enough in firefox to reliably catch all keystrokes. This fix changes the order of events, and keeps catching keystrokes after the first until the focus has shifted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+<a name="0.19.1"></a>
+## [0.19.1](https://github.com/angular-ui/ui-select/compare/v0.19.0...v0.19.1) (2016-08-09)
+
+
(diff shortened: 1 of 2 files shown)


wesleycho:
Rebase this on top of current `master`.

Plus3531:
I still experience this issue. Is is already fixed in the master code?

piebe:
Pull request is rebased... ready for review...

user378230:
Think you broke commit history, there should just be one commit with your changes. Also there shouldn't be any changes to the `/dist` files.

piebe:
Seems I made a mess of things. Another attempt: #1785

Issue #890, opened 2015-04-28
Tagging without multiple with new tags doesn't works when using angularjs 1.3.2


larrybolt:
When using angularjs 1.3.2 tagging without multiple doesn't work for new items that are not added to the array.
I've also tried editing `demo-tagging.html`, by replacing the angularjs libraries with following:
```
<script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.3.2/angular.js"></script>
<script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.3.2/angular-sanitize.js"></script>
```
And it seems to confirm something changed in the 1.3.2 to break this feature.

I'm really hoping there's an easy fix for this :disappointed: as this feature was exactly what I needed.

larrybolt:
And just trying to use angularjs 1.3.0 on the app I'm working on does run the tagging function, but doesn't seem to pass any arguments such as the new object.
``` Javascript
$scope.tagTransform = function (newTag) {
    console.log("new tag", newTag, arguments);
    // ...
  };
```

grebett:
+1

grebett:
I've written a quick workaround you can find [there](http://pastebin.com/djfdj4R1).
Not sure it's good enough for a pull request but may do the job for the while.

emreavsar:
+1 same issue here.

nutanabhiram:
I am facing the same issue , tagging is not enabling with following code using angularjs 1.3.2 
<ui-select tagging tagging-label="('new')" ng-model="singleDemo.color" theme="bootstrap" title="Choose a color">
    <ui-select-match placeholder="Select color...">{{$select.selected}}</ui-select-match>
    <ui-select-choices repeat="color in availableColors | filter: $select.search">
      {{color}}
    </ui-select-choices>
</ui-select>

karptonite:
I actually see this same problem in 1.3.1 as well. Works fine in 1.3.0.

guilhermerocha:
I'm facing the same problem.

ThomasAlber:
same problem here
you can reproduce the problem in the official example

http://plnkr.co/edit/m1SQXUxftBLQtitng1f0?p=preview

by changing the angular version in line 21 from
```<script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.3.0/angular.js"></script>```
to
```<script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.3.1/angular.js"></script>```
[...]

Issue #770, opened 2015-03-17
Unable to create tag with default tokens (,|ENTER) when tagging-label="false"


sime:
Since 0.11.2

If `taggging-label="false"` is used, the default tokens ',' and 'ENTER' do not create a tag.

Attempt to create a tag by pressing ',' or 'ENTER' on the following Plunkr.

http://plnkr.co/edit/RK7a6NgAAFGTEm48UjGn?p=preview

dmitryevseev:
+1

dmitryevseev:
As a workaround I've made results dropdown hidden:
```.select2-drop { display: none; }```

(in my case user does not need to select existing tags) and set tagging-tokens attribute:
```<ui-select multiple tagging ng-model="keywords" tagging-tokens="ENTER|TAB|,">```

It allows to have tokens working without ability to select existing values.

cesar-oyarzun-m:
hey @dmitryevseev I try your solution but I get console error 

```TypeError: Cannot assign to read only property '_uiSelectChoiceDisabled' of aaaa undefined```

How can I fix this?

dmitryevseev:
@cesar-oyarzun-m This solution worked on 0.11.2. You're probably using a newer version

cesar-oyarzun-m:
yeah I'm usigin 0.17.1 but I have the same problem

PR #899, opened 2015-04-30
fix for #890, #770

kshutkin:


brianfeister:
Thanks for this, need to take a look.

brianfeister:
Looks good, verified it works with this Plunkr: http://plnkr.co/edit/BZwOdzBKYedMkP9eMfo3?p=preview

LeleDev:
Single select tagging seems to work, but the new tags doesn't show up among the others in the dropdown (as instead it does with multiple select)

kshutkin:
I only fixed cases with tagging enabled but no labels (tagging-label="false").

LeleDev:
OK... maybe this pull request could help with other cases? https://github.com/angular-ui/ui-select/pull/972

Issue #848, opened 2015-04-14
Uncaught TypeError: Cannot read property 'clipboardData' of undefined when pasting


ruionwriting:
When pasting anything - all modes - the console throws the presented error.

bobs619:
I ended up changing the spot where it errors to this:

```javascript
var t = e.originalEvent.clipboardData ? e.originalEvent.clipboardData.getData("text/plain") : window.clipboardData.getData("text");
```

I tested it on IE 11 and Chrome.  Works!

rysilva:
The issue is that originalEvent is jquery.  Since jquery is supposedly not required for ui-select, this reference should be removed completely for the fix.

sdir456:
+1

shali3:
+1 Please fix

cfey:
hey guys, do you have any progress on this?

shali3:
I created a pull request #1391 to fix this yesterday.

cfey:
ah, thanks
+1

erikdubbelboer:
There are many pull requests for this such as #1276 , But the project isn't maintained anymore so nobody merges them.

cfey:
Is there a way to keep this going?
Would be sad if this project gets abandoned :confused:

shali3:
I opened #1398 
You are welcome to comment there

aaronroberson:
Please confirm fix in v0.14.7 and close this issue.

wesleycho:
Closing this as fixed - feel free to open a new issue if it is still a problem.

Issue #789, opened 2015-03-23
Error when pasting text in IE


BKlerx:
The code for the event handler that is called when pasting some text into the input control throws an error because 'e.originalEvent.clipboardData' is undefined in Internet Explorer.

I found a solution on http://stackoverflow.com/questions/27738155/unable-to-get-property-getdata-of-undefined-or-null-reference-in-ie-but-not

This code applies the fix and works fine for me:
  _searchInput.on('paste', function (e) {
    var data;
    if (window.clipboardData && window.clipboardData.getData) { // IE
      data = window.clipboardData.getData('Text');
    }
    else if (e.originalEvent.clipboardData && e.originalEvent.clipboardData.getData) { // other browsers
      data = e.originalEvent.clipboardData.getData('text/plain');
    }
...

I am using jQuery 2.1.3 and AngularJS 1.3.14

rohitrpaswan:
![image](https://cloud.githubusercontent.com/assets/4832153/14165093/df6f1d56-f722-11e5-8674-8a08aa2e77a0.png)

Issue #704, opened 2015-02-26
pasting issues


jordydejong:
I'm having issues with pasting a set of values (e.g "foo,bar"). I expect 2 items to be added.

The pasting logic is only executed if I define a tagging funtion

select.js:

      _searchInput.on('paste', function (e) {
            var data = e.originalEvent.clipboardData.getData('text/plain');
            if (data && data.length > 0 && ctrl.taggingTokens.isActivated && ctrl.tagging.fct) {

However in ctrl.select, is sends the search value (which is empty at that point) to my tagging function, instead of the item which was provided to the method

      if ( ctrl.tagging.fct !== undefined && typeof item === 'string' ) {
            item = ctrl.tagging.fct(ctrl.search);

Any ideas on what I am doing wrong? Or is this a bug? If I change the line above to the following, everything seems to work fine

            item = ctrl.tagging.fct(item);

PR #910, opened 2015-05-07
fixed pasting issue with generic ClipboardEvent in tagging mode

pyu10055:
fixed pasting issue with generic ClipboardEvent in tagging mode, and allow pasting when the tagging function is not defined.

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -441,12 +441,12 @@ uis.controller('uiSelectCtrl',
 
   // If tagging try to split by tokens and add items
   ctrl.searchInput.on('paste', function (e) {
-    var data = e.originalEvent.clipboardData.getData('text/plain');
-    if (data && data.length > 0 && ctrl.taggingTokens.isActivated && ctrl.tagging.fct) {
+    var data = (e.originalEvent || e).clipboardData.getData('text/plain');
+    if (data && data.length > 0 && ctrl.taggingTokens.isActivated) {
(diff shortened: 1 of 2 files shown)


pyu10055:
fixing bugs #704, #789, #848

faisalferoz:
+1 any chance of this getting merged?

pyu10055:
bump, can someone take look at this fix?

dmitry-dedukhin:
Actually, https://github.com/angular-ui/ui-select/issues/704 is not fixed by this change.
Paste handler immediately calls ctrl.select in loop and in case tagging-label="false" ctrl.select tries to parse text from ctrl.search which is not filled in because e.preventDefault() was called in paste handler.

To fix this we could add timeout before calling ctrl.select from paste handler, and remove e.preventDefault() from it something like this:

```javascript
ctrl.searchInput.on('paste', function (e) {
    var data = e.originalEvent.clipboardData.getData('text/plain');
    if (data && data.length > 0 && ctrl.taggingTokens.isActivated && ctrl.tagging.fct) {
        var items = data.split(ctrl.taggingTokens.tokens[0]); // split by first token only
        if (items && items.length > 0) {
            $timeout(function () {
                angular.forEach(items, function (item) {
                    var newItem = ctrl.tagging.fct(item);
                    if (newItem) {
                        ctrl.select(newItem, true);
                    }
                });
            });
            e.stopPropagation();
        }
    }
});
```

PR #939, opened 2015-05-18
Improved README guide on installation steps.

magic890:
If you try to: bower install ui-select
choosing the name from the repo bower.json name (as you can do for npm), you'll get a different repository (git://github.com/brettshollenberger/ui-select.git), so I think it's better to clarify this installation step.


user378230:
Superseded by #1550

PR #940, opened 2015-05-18
Prevent whole page scroll when scrolling ui-choices

emmanueldaher:
Prevent the browser's annoying default behavior of scroll the whole page
when scrolling in ui-choices options, useful when it has many
options, changeable in uiSelectConfig globals and preventPageScroll attribute.

### DIFF
--- a/src/uiSelectChoicesDirective.js
+++ b/src/uiSelectChoicesDirective.js
@@ -47,6 +47,18 @@ uis.directive('uiSelectChoices',
         if (rowsInner.length !== 1) throw uiSelectMinErr('rows', "Expected 1 .ui-select-choices-row-inner but got '{0}'.", rowsInner.length);
         rowsInner.attr('uis-transclude-append', ''); //Adding uisTranscludeAppend directive to row element after choices element has ngRepeat
 
+        if ($select.preventPageScroll) {// Prevent the whole page for scrolling when the choices ul reaches it's scroll limits.
+          element.bind('mousewheel', function(event) {
+            var heightDif = this.offsetHeight - this.clientHeight,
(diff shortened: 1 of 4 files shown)


src/uiSelectChoicesDirective.js
```diff
@@ -47,6 +47,18 @@ uis.directive('uiSelectChoices',
         if (rowsInner.length !== 1) throw uiSelectMinErr('rows', "Expected 1 .ui-select-choices-row-inner but got '{0}'.", rowsInner.length);
         rowsInner.attr('uis-transclude-append', ''); //Adding uisTranscludeAppend directive to row element after choices element has ngRepeat
 
+        if ($select.preventPageScroll) {// Prevent the whole page for scrolling when the choices ul reaches it's scroll limits.
+          element.bind('mousewheel', function(event) {
```
wesleycho: Change to `element.on`

src/uiSelectDirective.js
```diff
@@ -41,6 +41,14 @@ uis.directive('uiSelect',
           }
         }();
 
+        $select.preventPageScroll = function() {
+          if (angular.isDefined(attrs.preventPageScroll)) {
+            return $parse(attrs.preventPageScroll)();
```
wesleycho: Is it desired to parse this without a $scope passed into this?

emmanueldaher: I used the same $parse function from the others.
it only needs to parse string to boolean.

dorrotman:
While this was not merged, I used this code in an angular directive that applies by the class name of the dropdown list container, and it helped me a lot. 
Thanks!

PR #941, opened 2015-05-19
Cannot select options in ie8

vera0119:
Cannot select options in ie8.
when I open the dropdown, the cursor jump to the last item in the list.


orloffv:
+1

vera0119:
It has been resolved in the latest version(0.19.8).

Issue #938, opened 2015-05-18
Tagging Label isn't implemented in uiSelectSingle


viniciusdacal:
There's a way to make Tagging Label works in uiSelect Single?

Pavek:
+1

KylePDavis:
Same here.

Realized it wasn't just me when I noticed that the [tagging demos without multiple](https://github.com/angular-ui/ui-select/blob/master/examples/demo-tagging.html#L119) aren't working either.

In the meantime though I have been able to work around this by using the `v0.9.9` release with `$select.taggingLabel` in my markup.

e.g.,
```html
<i ng-if="item.isTag">{{$select.taggingLabel}}</i>
```

Hope that helps.

MartinNuc:
+1

LeleDev:
pretty serious issue, tagging is still broken for single-select, please fix it soon...

fferraris:
:+1:

Krisa:
Indeed, the example is bugged :(

@KylePDavis Could you please show (e.g. in a plunker) how you used the taggingLabel as workaround?

KylePDavis:
I used it in my [mongodb-workbench](https://github.com/KylePDavis/mongodb-workbench) project.

Here, specifically: https://github.com/KylePDavis/mongodb-workbench/blob/v0.0.1/src/web/components/connection/connection.html#L38

Krisa:
I see thank you. Actually, it's just to keep the value of the taggingLabel in a dynamic way but not changing much the issue. 

But my problem is actually even different. The call to the tagging function (stored in the attribute tagging) is just not working in my case. It works when I specify the attribute "multiple" but as soon as I go into a single picklist, my function is not called anymore and then the rest fails obviously... I'm using controllerAs, I fear it might be something not supported yet. Finally, my workaround is to do a "multiple" for a single picklist, and I have added limit=1...

viniciusdacal:
@Krisa I fix this bug in my fork https://github.com/viniciusdacal/ui-select
and there is a pull request opened #972

performatric:
@viniciusdacal your fork works in angular ^1.4?

viniciusdacal:
@performatric yes, currently I'm using with angular 1.4.9 and it's working pretty fine.
In that fork i just fixed this problem with tagging.
I created another repository that we are currently maintaining and adding new features https://github.com/viniciusdacal/angular-select
I really don't like that fragmentation, but for this case I only see this way

[...]

PR #942, opened 2015-05-19
Move keyup eventHandler to controller, this way we can use taggingLabel

viniciusdacal:
Move keyup eventHandler to controller, this way we can use taggingLabel in multipleSelect and singleSelect.
Related to issue #938


jashmenn:
This is great. Tested and it works for me. Hopefully it can be merged to master soon.

viniciusdacal:
Pull request closed in favor to #972

PR #943, opened 2015-05-19
fix: appendToBody doesn't work with jQuery < 1.10.0

AlexFess:
.relpaceWith in jQuery (version less then 1.10) doesn't work properly
and as so, ui-select with 'appendToBody=true' doesn't display selected
value.
Use .after() and .remove() instead of .replaceWith() to avoid this
issue.


AlexFess:
Add a fix for https://github.com/angular-ui/ui-select/issues/944 (remove target element check)

wesleycho:
This needs squashing, then rebasing.

PR #945, opened 2015-05-19
Use npm to install required dependencies to run the tests.

yagoferrer:
Angular, Angular Sanitize, Angular Mocks and JQuery.
Removed: post-install bower script. No longer need to call bower from npm.
This will help when you guys decide to publish the package in NPM.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -9,10 +9,10 @@ module.exports = function(config) {
 
     // List of files / patterns to load in the browser
     files: [
-      'bower_components/jquery/dist/jquery.js',
-      'bower_components/angular/angular.js',
-      'bower_components/angular-sanitize/angular-sanitize.js',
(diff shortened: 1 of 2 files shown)


JasonMore:
+1 from me. I was just about to post a PR with not using bower at all, since I'm using UI-Select with webpack as a faux npm module.

```
  "dependencies": {
    // cruft removed
    "ui-select": "git://github.com/angular-ui/ui-select"
}
```

Anahkiasen:
Any news on this one?

user378230:
This doesn't seem like a bad idea...

@aaronroberson thoughts?

aaronroberson:
@yagoferrer and @JasonMore You can install ui-select from npm currently. The official npm package can be found here: https://www.npmjs.com/package/ui-select

With that said, I'm going to merge this PR in because it is a good idea :D

PR #947, opened 2015-05-21
ui optimize

billypon:
1. bootstrap/match-multiple.tpl.html: remove "&nbsp;" and use "margin-left" instead
2. theme select2: searchInput's height should be the same as the item's
3. searchInput's width should fill the rest space, and min-width is too wide

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -18,6 +18,9 @@
 }
 
 /* Select2 theme */
+.select2-input {
+    height: 27px !important;
+}
 
 /* Mark invalid Select2 */
 .ng-dirty.ng-invalid > a.select2-choice {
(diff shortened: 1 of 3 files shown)


src/uiSelectController.js
```diff
@@ -344,13 +345,17 @@ uis.controller('uiSelectCtrl',
...
           }
-          var inputWidth = containerWidth - input.offsetLeft - 10;
-          if (inputWidth < 50) inputWidth = containerWidth;
+          var inputWidth = containerWidth - input.offsetLeft;
+          if (theme == 'bootstrap')
+              inputWidth -= 1;
           ctrl.searchInput.css('width', inputWidth+'px');
           return true;
         };
+    if (theme != 'bootstrap')
```
wesleycho: Use !==

src/common.css
```diff
@@ -18,6 +18,9 @@
 }
 
 /* Select2 theme */
+.select2-input {
+    height: 27px !important;
```
wesleycho: This is a harmful breaking change for users - is there no way around this?

src/uiSelectController.js
billypon: I have modified

src/common.css
billypon: Because the select2's css use "height: auto !import", it will be overwrited.
But now my css rule doesn't work, it will be overwrited too.

PR #948, opened 2015-05-21
Fix demo clear button

bobkilla:
Fix the clear button on the demo page not working.


user378230:
Going to close as this would need rebasing and rewriting anyway, If anyone wants to submit a new PR for this the code is currently around [/docs/assets/demo.js#L70](https://github.com/angular-ui/ui-select/blob/fe0c0c12522d48d677fff8ca21d366ba37d8f90d/docs/assets/demo.js#L70)

Issue #806, opened 2015-03-30
Track by in multiple select


i4got10:
If specify __multiple__ and __track by__ simultaneously I recieve the error "Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed."

Demo. Just select 2 items in select, and you will get the error.
http://plnkr.co/edit/taoUPj40sWnjuRD5zVCh?p=preview

chuyik:
Same here.

chuyik:
Use the forked version and everything works great with Angular 1.3.

Try it: https://github.com/richardallen/ui-select

joaquinrivero:
@chuyik But, won't load a selected values.

eranimo:
Can this be fixed? Currently multiple entries are unsupported because of this.

segalr3:
same here, using angular 1.2.25, ui-select latest version

Issue #366, opened 2014-11-03
[ngRepeat:dupes] Duplicates in repeater are not allowed. Broken after angular/ui/select upgrade


davisford:
I have what should amount to a fairly simple use of ui-select, in fact I've been running it in production for months now with a pinned version of angular-ui and angular b/c things have been a bit dicey trying to get angular-ui working with angular `1.3.2`.  Well, today, I'm trying to update the dependencies and work through some of these issues, and this ui-select broke in the process.  Here's output from bower on the versions:

```
├── angular#1.3.2-build.3514+sha.d906ed3
   /* removed other angular modules, e.g. angular-route, etc. */
├─┬ angular-ui-bootstrap-bower#0.11.2
├─┬ angular-ui-select#0.8.3
```

This is the error being thrown after the upgrade:

```
Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: $item in $select.selected, Duplicate key: string:u, Duplicate value: "u"
http://errors.angularjs.org/1.3.2-build.3514+sha.d906ed3/ngRepeat/dupes?p0=%24item%20in%20%24select.selected&p1=string%3Au&p2=%22u%22
    at http://localhost:3000/angular/angular.js:80:12
    at ngRepeatAction (http://localhost:3000/angular/angular.js:24073:21)
    at Object.$watchCollectionAction [as fn] (http://localhost:3000/angular/angular.js:13819:13)
    at Scope.$digest (http://localhost:3000/angular/angular.js:13952:29)
    at Scope.$apply (http://localhost:3000/angular/angular.js:14214:24)
    at done (http://localhost:3000/angular/angular.js:9474:47)
    at completeRequest (http://localhost:3000/angular/angular.js:9659:7)
    at XMLHttpRequest.requestLoaded (http://localhost:3000/angular/angular.js:9602:9) angular.js:11339(anonymous function) angular.js:11339(anonymous function) angular.js:8415Scope.$digest angular.js:13970Scope.$apply angular.js:14214done angular.js:9474completeRequest angular.js:9659requestLoaded
```

Obviously, this is a [well-known error, and nicely documented, to boot](https://docs.angularjs.org/error/ngRepeat/dupes)...the problem is `track by` doesn't seem to resolve the issue, and besides, I'm binding to an array of strings in the `repeat`, which shouldn't need `track by` AFAIK, because it ought to just use the array index (I thought).

Here's what the code looks like:

[...]

PR #953, opened 2015-05-24
Add taggingLabel for single and multiple select

viniciusdacal:
Add taggingLabel for single and multiple select.
Fix "track by" issue.
When we tried to use track by with objects, we got an error like this: issue #806 and issue #366


brianfeister:
@viniciusdacal PR's should not contain any changes to `/dist` files.  It's confusing, I know. Sorry.  Also, can you update this plunkr with your fix so we have verification that the fix addresses the code from the #806 bug report.

http://plnkr.co/edit/taoUPj40sWnjuRD5zVCh?p=preview

viniciusdacal:
My bad, I didn't see the instructions.
I create a new branch removing `dist` files and I sent a new pull request #972.
About #806, I was getting the same type of error, but in a different case. So, this pull request does not fix that issue.

PR #957, opened 2015-05-26
remove match-multiple directly when press backspace

billypon:
remove match-multiple directly when press backspace on searchInput


wesleycho:
This needs test fixed.

user378230:
Closing due to age, lack of tests and because I think requiring double backspace helps guard against accidental deletion.

PR #959, opened 2015-05-26
enable match-multiple to press left/right/delete/backspace

billypon:
enable match-multiple to press left/right/delete/backspace

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -168,7 +168,7 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
         if (oldValue && !newValue) $select.sizeSearchInput();
       });
 
-      $select.searchInput.on('keydown', function(e) {
+      var _handleKeyDownEvent = function(e) {
         var key = e.which;
         scope.$apply(function() {
           var processed = false;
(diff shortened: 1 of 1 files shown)


src/uiSelectMultipleDirective.js
```diff
@@ -183,6 +183,11 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
             e.stopPropagation();
           }
         });
+      };
+      $select.searchInput.on('keydown', _handleKeyDownEvent);
+      $timeout(function() {
```
wesleycho: This seems very fragile - is the $timeout necessary?

billypon:
@wesleycho `element.parent().querySelectorAll(".ui-select-match")` may be empty

wesleycho:
The reason I make the comment is that we should be working to reduce $timeout usage as much as possible instead of adding more.

Also, shouldn't it be `element.parent()[0].querySelectorAll('.ui-select-match')`?

billypon:
@wesleycho I'm sorry, my reply is wrong before.

Actually, `element.parent().querySelectorAll(".ui-select-match")` is not empty, but the "keydown" event handler will not be invoked. So I use "$timeout".

`element.parent().querySelectorAll` is and function from angular, another is from native.

PR #982, opened 2015-06-02
Do not run bower after install automatically

vperron:
This should be an user action, not something automatically run on install.
We're using ui-select in a browserify project, and we don't need/have bower installed. Furthermore it pollutes the development folder with an unused bower_components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "karma-coverage": "~0.2"
   },
   "scripts": {
-    "postinstall": "bower install",
     "test": "gulp test"
   },
   "license": "MIT"


vperron:
If someone sees this, please tell us if merging is an option.

k-lange:
@brianfeister can we please get this merged? If you want to keep the ```bower install``` for development purposes you can add it to ```prepublish```, then it's installed when you do a local ```npm install```.

vperron:
Not even k-lange... NPM should _not_ rely on bower, it's a malpractice on the long term ! As if apt-get was relying on yum or pacman...

vperron:
Please please please can you finally merge that one ???

visnup:
+10

visnup:
💖

vperron:
PRAISE THA LORD

shlensky:
:+1:

PR #989, opened 2015-06-03
fix incorrect  up direction class - document scroll position

sagivf:
I think the "html" element doesn't have scrollTop, the "body" does


stulentsev:
We need this merged.

rysilva:
This change works in chrome/safari but then breaks firefox.

Here's what I did to fix both.  Haven't tested IE yet though: https://github.com/rysilva/ui-select/commit/38122679cd262294552d9bb7a10deae2a62f126a

dimirc:
Should be solved with https://github.com/angular-ui/ui-select/pull/1212

Issue #994, opened 2015-06-04
Clicking outside of the Select should always close the Select


jmckinney88:
In commit [aef5a8088b749b6c3d5c5d3d12998d7a30e8bdfd](https://github.com/angular-ui/ui-select/commit/aef5a8088b749b6c3d5c5d3d12998d7a30e8bdfd), support was added for close-on-select. However, these changes also included the addition of the 'clickTriggeredSelect ' property on the controller. The 'clickTriggeredSelect' seems to serve no purpose in support of the close-on-select attribute and is preventing the first click outside of the Select from closing the Select if any item has been selected with a mouse click prior to the external click.

*Reproduction Steps w/ close-on-select=true:*
1. Open a Select, single or multi, by any means (click or enter)
2. Using the mouse, select an item.
3. Re-open the Select.
4. Using the mouse, click anywhere outside of the Select.
5. Clicking again outside the Select will close the Select.

Expected Behavior after step 4: The Select should be closed.
Actual Behavior after step 4: The Select is still open.

Jaco-Pretorius:
:+1: Getting the same issue

JonWallsten:
+ 1

nhim175:
1.5 year has passed, the bug is still there... looking forward to the fix ^^

user378230:
@nhim feel free to put up a PR :) 

A plunkr might go also help any fixers

Jefiozie:
Had a look at this issue and I think the clickTriggerSelect is not reset correctly. 

On [Line 385](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectController.js#L385) the clickTriggerSelect is set to true but never back to false.
At the [onDocumentClick](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectDirective.js#L162) there is a check if this is true or false. When the if statement is not true it will set the clickTriggerSelect to false, that is the reason why it will close the second time.

Have made a [plunker](http://plnkr.co/edit/3E8ARt3vbM5DzuXrz8Ku?p=preview) to see if this will resolve the issue. 

Let me know then I will commit the changes. :smile:

Jefiozie:
@user378230 , do you think I need to submit this as a PR?

Jefiozie:
@jmckinney88  Did you find some time to test this? Thanks

jmckinney88:
@jefiozie I have not yet and am away from my desktop at the moment, but I can take a look tomorrow. Thanks for the tag.

[...]

PR #995, opened 2015-06-04
Issue 994: Change guard against closing the Select when in a multisel…

jmckinney88:
…ect with close-on-select(false) to prevent failing to close when mixing interaction between keyboard and mouse.

Changes are proposed to fix the issue detailed in [Issue 994](https://github.com/angular-ui/ui-select/issues/994)

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -155,8 +155,13 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
       };
 
       scope.$on('uis:select', function (event, item) {
-        $select.selected.push(item);
-        $selectMultiple.updateModel();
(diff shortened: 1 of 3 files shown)


src/uiSelectMultipleDirective.js
```diff
@@ -155,8 +155,13 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
       };
 
       scope.$on('uis:select', function (event, item) {
-        $select.selected.push(item);
-        $selectMultiple.updateModel();
+        // The model should be updated so that selected items may be removed, however
+        // this should be done after the onDocumentClick handler has run so that the
+        // item's list item is still in the DOM.
+        $timeout(function () {
```
wesleycho: This seems fragile - why is this necessary?

PR #996, opened 2015-06-05
Ensure $select.selected exists prior to testing length.

sime:
I have a use case where following exception is being thrown.  To be honest I do not totally understand the execution path, but handling this exception seems obvious enough.

```TypeError: Cannot read property 'length' of undefined
    at uis.directive.controller.ctrl.getPlaceholder```


user378230:
Appears to be in place already https://github.com/angular-ui/ui-select/blob/331f819ce0f549e0836bd4724bba67c65a6d957a/src/uiSelectMultipleDirective.js#L62

We should probably use `isEmpty` really though when I come to think about it.

PR #836, opened 2015-04-08
Issue #148 - Add up direction to dropdown rendering.

JeromeLam:
This addresses issue #148, which relates to how the dropdown displays in the event there isn't enough space to render it fully.  An up direction has been added.

I verified the change against demo.html, demo-multi-select.html and the three theme html files.

PR #998, opened 2015-06-05
Fix to recalculate top position of dropdown when shown upward

ischenkodv:
When dropdown is opened in upward direction and I begin to type something so that there are just a couple of elements visible then dropdown appears high above the field. Here is the screnshot:
up_direction_height_bug

And plunker to see the bug: http://plnkr.co/edit/283bzy?p=preview (type something in the bottom field but note that it could depend on window size, ensure that element is on the bottom).
Here is the plunkr with my fixed version of select.js: http://plnkr.co/edit/0rdvSM?p=preview

It seems that this bug was introduced in #836 

I setup more $watch callbacks to recalculate top position of dropdown when `$select.search` and `$select.selected` variables are changed.


arambert:
:+1:

ManuelPutallaz:
(Y) thanks

+1 to this feature in the new version

cdjackson:
Just for information, I hadn't seen this PR, but have also submitted a fix in #1123. This fix is to reduce code and add the changes directly into the css classes which I think is a better approach than adding watches and recalculating the sizes.

PR #972, opened 2015-05-28
Add taggingLabel for single and multiple select

viniciusdacal:
issue #938

PR #999, opened 2015-06-08
Fix #954

### LINKED ISSUE #954: Cannot clear first option (single value tagging)
I have probably found bug in combination of using tagging feature without multiple.
In such situation with allow-clear=true I am not able to clear value when first one from the list is selected.

As fix that is working for me i suggest to add  ctrl.activeIndex = -1; at the first line of clear function.

Please verify this bug and also suggested solution and if it is correct apply it in your code.

thanks

kshutkin:
Clear functionality was broken for single select mode with tagging (no labels, because with labels tagging broken completly for single select mode).

I think this fix must be tested again after https://github.com/angular-ui/ui-select/pull/972 merged.


wesleycho:
Closing due to broken history (there should be no merge & revert commits).

Feel free to open a new PR against current `master` with clean history.

PR #1004, opened 2015-06-11
Fix an error in multiple mode

MNBuyskih:
In multiple give an arror on startup, when there no selected items:
"TypeError: Cannot read property 'length' of undefined"

Sorry, my Sublime remove some trailing spaces. Actual changes on lines 58-60


ckniffen:
+1

nikolai-b:
`$select.selected` can be an empty array in which case it should also `return`.
I think #996 is a slightly better solution.

cdjackson:
Probable duplicate of #956 and #996

wesleycho:
Closing due to fix having already made it in `master`.

PR #1008, opened 2015-06-12
fix(highlight filter): bug in highlight if number type

zhangyaochun:
if number type, it has no replace method, so it will run error


user378230:
#1402 merged so closing this.

PR #1017, opened 2015-06-16
Attach uiSelectDirective hide to ontouchstart if defined

posgarou:
ui-select currently does not work properly in touch-enabled browsers/devices when ngTouch is included.

This pull request fixes this problem by attaching uiSelectDirective's auto-hide functionality to the document's touchstart event, if defined, rather than the click event.


wesleycho:
ngTouch has been deprecated.

I would recommend against this, as it has ramifications when other solutions like fastclick or hammer.js are used.

user378230:
Agreed, closing.

PR #1022, opened 2015-06-18
Adds the containerCssClass and dropdownCssClass options from Select2 …

RyanDale:
…to allow custom CSS classes to easily be applied to ui-selects

### DIFF
--- a/src/bootstrap/choices.tpl.html
+++ b/src/bootstrap/choices.tpl.html
@@ -1,4 +1,4 @@
-<ul class="ui-select-choices ui-select-choices-content ui-select-dropdown dropdown-menu"
+<ul class="ui-select-choices ui-select-choices-content ui-select-dropdown dropdown-menu {{ $select.dropdownCssClass }}"
     role="listbox"
     ng-show="$select.items.length > 0">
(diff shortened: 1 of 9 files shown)


Balrog30:
I was looking to add the very same functionality myself.  There's a color scheme we use for status labels on our site, and I wanted to use ui-select for an item that can have multiple status flags.

Had I written it myself I would have tried to integrate it with ng-class so that you can pass multiple classes (say one that applies text rules and one that applies box shape or color rules, for a contrived example).  It could be done by managing the classes in the ui-select-match directive's link function instead of in the template.

But hey, if this gets merged I'd use this just as well.

user378230:
This could probably be accomplished by #1531 when someone does that. 

Couldn't you just do:

```html
<div class="my-class">
  <ui-select...
    ...
  </ui-select...
```
```css
.my-class .ui-select-dropdown {
  ...
  ...
}
```
I just feel like this might be a little bloaty and I think that most would expect them to be watched and updated.

@wesleycho Any thoughts on this? I'm edging towards closing.

wesleycho:
If the library isn't using `replace: true`, then there is no reason for this feature.

Issue #914, opened 2015-05-11
required or ng-required not working with ui-select 'multiple' option


vinayagarwal:
required or ng-required=true is working fine with ui-select when the control is expected to select single value. However, with 'multiple' option, it is not working. The model presumes the value as array [], and marks it as valid=true.

Am I the only one facing this issue? Is this issue already addressed in the next version?

imaa:
I found this solution here 

<code>
angular.module('common').directive('uiSelectRequired', function () {
    return {
        require: 'ngModel',
        link: function (scope, elm, attrs, ctrl) {
            ctrl.$validators.uiSelectRequired = function (modelValue, viewValue) {

                var determineVal;
                if (angular.isArray(modelValue)) {
                    determineVal = modelValue;
                } else if (angular.isArray(viewValue)) {
                    determineVal = viewValue;
                } else {
                    return false;
                }

                return determineVal.length > 0;
            };
        }
    };
});
</code>

add  ui-select-required

vinayagarwal:
Thanks imaa! It is working well. The error msg also started showing up when I change the condition from:
ng-show="abcForm.tasks.$error.required"
to
ng-show="abcForm.tasks.$invalid"

Wondering, can this be merged with the main repository or with ng-required itself with more sophistication to handle $error.required as well?

frekele:
ng-required or only required not work witk "<ui-select multiple tagging" for me too.
Componente in form validation not detect state @$scope.myform.$valid or @$scope.myform.$invalid.

evdoks:
I confirm the behaviour - when `multiple` is set `required` is not working.

jcolemorrison:
I just submitted a PR to fix this issue.  It will allow you to just put `required` directly on the multiple tag and receive the expected functionality.

PR Link:  https://github.com/angular-ui/ui-select/pull/1025

vinayagarwal:
That's great Cole. Thank you.

improvisio:
The solution above works for me as well. Thanks!

GilbertChan:
@improvisio how did you get the solution above working? I added the module then added ui-select-required in the ui-select tag but its not working.

improvisio:
@GilbertChan I first added the directive as shown above: https://github.com/angular-ui/ui-select/issues/914#issuecomment-100962567
Then it worked. ui-select-required is not part of the ui-select package.

japharr:
Thanks, @imaa it works!

[...]

Issue #850, opened 2015-04-14
required tag


brackmayhall:
Is it possible to set a required tag for this control? I am not getting a prompt when the required tag is set and it is embedded in a form element. Thanks.

viveleroi:
Are you using `ng-required`? I was just looking into this myself and noticed that it does work.

brackmayhall:
I have tried both with no success. Any idea what I could be doing wrong here?
```
                                <ui-select id="vendorNameSelect" ng-model="item.vendorId" theme="select2" ng-disabled="!new"  title="Select Vendor" ng-required="true">
                                    <ui-select-match placeholder="Select Vendor">{{$select.selected.name || $select.selected}}</ui-select-match>
                                    <ui-select-choices repeat="vendor.vendorId as vendor in vendors | propsFilter: {name: $select.search}">
                                        <div ng-bind-html="vendor.name | highlight: $select.search"></div>
                                        <small>
                                            phone: {{vendor.phone}}
                                        </small>
                                    </ui-select-choices>
                                </ui-select>
```

viveleroi:
Nothing jumps out at me, though I have no what might conflict with the validation. Maybe the disabled state?

Here's my code for reference. When the drop down is an empty value, the field container is properly marked as `ng-invalid`, and `ng-valid` when I choose something.

    <ui-select ng-model="model.vessel" append-to-body="true" search-enabled="false" ng-required="true">
        <ui-select-match>{{$select.selected}}</ui-select-match>
        <ui-select-choices repeat="vessel in vessels">
            <span ng-bind-html="vessel"></span>
        </ui-select-choices>
    </ui-select>

brackmayhall:
I'm still having issues. Going to try to break it down using the examples.

brackmayhall:
I stripped down the demo.html example and gave this a shot. The input prompts holds true for the required tag but the select does not.

'''
<form name="testForm">
  <input name="firstInput" ng-model="data.first" required>

  <p>Selected: {{person.selected}}</p>
  <ui-select ng-model="person.selected" theme="select2" ng-disabled="disabled" style="min-width: 300px;" title="Choose a person" required>
    <ui-select-match placeholder="Select a person in the list or search his name/age...">{{$select.selected.name}}</ui-select-match>
    <ui-select-choices repeat="person in people | propsFilter: {name: $select.search, age: $select.search}">
      <div ng-bind-html="person.name | highlight: $select.search"></div>
      <small>
[...]

PR #1025, opened 2015-06-18
Bug/required multiple select

jcolemorrison:
Closes the following:

https://github.com/angular-ui/ui-select/issues/850

and

https://github.com/angular-ui/ui-select/issues/914

The fix adds a custom validator to multiple selects that enables the `required` tag.  Custom validators are not "my" own customized way of doing things, they are the official angular way of writing validators as outlined in the Custom Validators section here https://docs.angularjs.org/guide/forms

The problem was that required checks the ngModel's viewModel, which when empty on a ui-select-multiple is an empty array...which is technically not empty.  Now it checks properly without interfering with the standard required validator.

Also fixed a test bug where $item.name wasn't being used in the multiple example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 /node_modules
 /.idea
 /.tmp
-/dist
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 /node_modules
 /.idea
 /.tmp
+/dist
 .DS_Store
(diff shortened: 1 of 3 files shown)


jcolemorrison:
I can remove the Git ignore on the /dist folder if need be.

guidoffm:
I have found this problem: When I also have the ng-change attribute pointing to a function and when I remove the last element from the multiple selection then the model is set to undefined.

user378230:
I believe #1492 fixes this and has been merged, so closing. If you disagree please let us know.

PR #1026, opened 2015-06-19
Fix validating multiselect fields with 'required' attribute

### LINKED ISSUE #258: ng-required not validating form on submit

slawekkolodziej:
This PR makes 'required' attribute working with multiselect fields and fixes #258, #850 and #914. It overrides `ngModel.$isEmpty`, but unlike PR https://github.com/angular-ui/ui-select/pull/422 it handles the following scenario (https://github.com/angular-ui/ui-select/issues/258#issuecomment-64919605):

> 1. without selecting any thing ng-required work fine
> 2. select few element
> 3. removed all element from drop down
> 4. you see ng-required is not working any more now.

(See unit tests)

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -92,6 +92,10 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
         return resultMultiple;
       });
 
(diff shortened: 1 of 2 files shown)


user378230:
#1492 merged

PR #1028, opened 2015-06-19
Displaying 'no results' message

slawekkolodziej:
This PR adds support for displaying 'no results' message (when defined). Although I implemented this feature only for bootstrap theme.


user378230:
I believe this has been superseded by #1011

PR #1030, opened 2015-06-19
Making sure atleast one items is highlighted before selecting on TAB

jintoppy:
Current behaviour is, when user press TAB, ui-select is just triggering select() without checking if there is any activeIndex. If I do a clear, then there shouldn't be any highlighted item and it should not select when I press TAB. 

Please correct me if I am wrong


wesleycho:
Reverted this since there's a test it breaks when merged with current `master` - this needs updating.

PR #1035, opened 2015-06-23
Fix tolerance for null or undefined values

### LINKED ISSUE #967: Possible error with ngModel.$render with null/undefined values
I was having an issue with a multiple select item when the provided model was undefined and noticed that ngModel.$render first checks if ngModel.$viewValue is null, and if so, set  $select.selected to an empty array.

However, immediately after, $select.selected is set to ngModel.$viewValue again, causing it to become undefined again. I suppose a fix would be to set ngModel.$viewValue = [] rather than $select.selected = [].

Original Code:

      ngModel.$render = function() {
        // Make sure that model value is array
        if(!angular.isArray(ngModel.$viewValue)){
          // Have tolerance for null or undefined values
          if(angular.isUndefined(ngModel.$viewValue) || ngModel.$viewValue === null){
            $select.selected = [];
          } else {
            throw uiSelectMinErr('multiarr', "Expected model value to be array but got '{0}'", ngModel.$viewValue);
          }
        }
        $select.selected = ngModel.$viewValue;
        scope.$evalAsync(); //To force $digest
      };

Possible fix:

      ngModel.$render = function() {
        // Make sure that model value is array
        if(!angular.isArray(ngModel.$viewValue)){
          // Have tolerance for null or undefined values
          if(angular.isUndefined(ngModel.$viewValue) || ngModel.$viewValue === null){
            ngModel.$viewValue = [];
          } else {
            throw uiSelectMinErr('multiarr', "Expected model value to be array but got '{0}'", ngModel.$viewValue);
          }
        }
        $select.selected = ngModel.$viewValue;
        scope.$evalAsync(); //To force $digest
      };

homerjam:
fixes #967


homerjam:
Any chance of merging this?

wesleycho:
Changes to `dist` need to be removed, and this needs to be squashed and rebased.

homerjam:
Hi @wesleycho is there any chance you could help/explain how to go about this? I'm a bit of a casual git user :blush:  (generally I use sourcetree - in case that helps?)

wesleycho:
In this case, `git reset --hard a926f70; git pull --rebase origin master; git push -f` should do the trick - do those commands separately though so you can step through things.

homerjam:
Thanks. Does the above look right?

wesleycho:
Looks like something was busted in the process - might be better to start fresh by killing the branch locally, updating master with current master, branch off and redo changes then force push up.

homerjam:
I made a mistake and deleted my branch - have resubmitted [here](https://github.com/angular-ui/ui-select/pull/1529)

PR #1036, opened 2015-06-23
Update model on sort (fixes #974)

### LINKED ISSUE #974: sortable="true" does not update the ng-model
Reference: #209

After drag n' drop of options selected (re-ordering), the selected items (`<ui-selelct ... ng-model="mycontrollerscopeattr".. >`) is not updated. 

Try it out here: http://plnkr.co/edit/eiwaOrkK4V9ZGwfGE5IV?p=preview

If this is not a bug (and developer error) then, can you make the example work? I'll gladly update any documentation.

homerjam:


faisalferoz:
+1

pgrm:
would be great if this could be merged, any updates?

homerjam:
bump

mrak1990:
bump, without this fix sortable option not useful

dmitry-dedukhin:
Although exists a workaround https://github.com/angular-ui/ui-select/issues/974#issuecomment-106801452 when we have more than one ui-select we can't use uiSelectSort:change event since we don't know which ui-select has triggered event.

wesleycho:
Remove changes to `dist`, squash and rebase.

homerjam:
Hi @wesleycho sorry I have no idea how to do what you just said. I've reset to the previous commit (without changes to dist) and force pushed. Hopefully this works?

src/uiSelectSortDirective.js
```diff
@@ -92,6 +92,8 @@ uis.directive('uiSelectSort', ['$timeout', 'uiSelectConfig', 'uiSelectMinErr', f
         }
 
         move.apply(theList, [droppedItemIndex, newIndex]);
+        
+        scope.$parent.$selectMultiple.updateModel();
```
user378230: Instead of going through `$parent` can't we `require: ['^uiSelect', '?^uiSelectMultiple']` and access via a `ctrls[]` param?

homerjam: I've just tried this and the 2nd controller is `undefined`? I'm not quite sure what to try...

qraynaud: I think you can require ^ngModel and not ^uiSelectMultiple and call ngModel.$setViewValue(Date.now());

Is it not the only useful thing that updateModel does here? Everything else seems like overhead in this function for what you are trying to achieve.

qraynaud: I tried it and it is working fine.

```js
require: ['^uiSelect', '^ngModel'],
link: function(scope, element, attrs, ctrls) {
  var $select = ctrls[0];
  var ngModel = ctrls[1];

  // ...

  $ngModel.$setViewValue(Date.now());

  // ...
}
```

homerjam:
@user378230 updated with @qraynaud 's suggestions - can confirm it works

qraynaud:
@homerjam : you have conflicts to fix, hope it will me merged (& released) soon. I need this working too :+1:

user378230:
@homerjam This LGTM :+1:

Will merge once conflicts resolved.

homerjam:
Sorry guys I have no idea what I'm doing, literally. Will this do?

qraynaud:
@homerjam : you removed the double `^` on `uiSelect`. I don't think this is a good idea.

homerjam:
Nuts. Ok. What does that do?

user378230:
@homerjam can you confirm that the fix works with the change back to  `^^`? 

This needs a squash to a single commit but I can do that as part of the merge. (See https://help.github.com/articles/about-git-rebase/#further-reading if you want to have a go yourself)

homerjam:
Hi @user378230 

I can confirm that it works and have since read up on the `^^` and seems a good idea.

I've tried doing the rebase several times but I keep getting conflicts and going round in circles 😞 . Thanks for the link but in this case please can you take over in the interest of getting this done!

Thanks

Issue #1016, opened 2015-06-16
Dropdown direction issue on webkit


pulse14:
The problem is on this line (in uiselect directive), on Chrome (43) ```$document[0].documentElement.scrollTop``` return 0

```javascript
 else if (offset.top + offset.height + offsetDropdown.height > $document[0].documentElement.scrollTop + $document[0].documentElement.clientHeight)
```
It works if i replace by 
```javascript
 else if (offset.top + offset.height + offsetDropdown.height > $document[0].body.scrollTop + $document[0].documentElement.clientHeight)
```
But it doesn't work anymore on Firefox now...

Anyone have this problem ?

EDIT :
```window.pageYOffset``` work on both

Thank you

hengun348:
I can confirm this issue as well

acetrand:
Im seeing this also in Chrome. Please fix asap.

pulse14:
I will make a pull req. 
It works fine after my fix.

acetrand:
@pulse14 Please do. I didnt follow where you put the `window.pageYOffset` so I'm holding my breath for your PR :wink:

sricc:
Seeing this as well.

jonasflesch:
I confirm that this fixed the problem for me.

tjsail33:
+1. This is a pretty serious bug and the PR needs to be accepted.

tinybigideas:
+1 for the PR to be accepted. Great fix @pulse14. I've accepted this into my build and it has corrected the issue. Hopefully this is accepted into the master sooner rather than later!

Iku-turso:
+1.

evillemez:
+1 I can also confirm the fix by @pulse14 works.  Thank you!

dimirc:
Should be solved with https://github.com/angular-ui/ui-select/pull/1212

PR #1047, opened 2015-06-26
Fix dropdown direction issue on webkit

pulse14:
#1016


dimirc:
Should be solved with https://github.com/angular-ui/ui-select/pull/1212

Issue #874, opened 2015-04-21
Demo doesn't block browser auto-complete


AgDude:
Using Chrome 41, the browser allowed me to select one of my stored addresses, which was not an allowed option on the scope. Wrapping the ui-select in a ```<form autocomplete="off">``` tag, solves the issue.

This is a widely discussed problem of browsers selectively respecting the autocomplete attribute.

realrunner:
What do you mean by "a tag" ?

AgDude:
Sorry, I entered invalid markdown, the comment should make more sense now.

jquinter:
tried this under chrome 42, osx 10.10.2, last version stable of ui-select. Didn't solve the issue.

MikeMatusz:
Seems like in the latest versions of Chrome/FF, I have to actually modify the template to put autocomplete="false" on the input element.  "off" doesn't seem to work anymore.

jgesser:
Hi folks, the solution made by hugobessa worked for me.
Does any one knows when it will be released and this issue closed?
thanks!

dimirc:
Should be solved soon with https://github.com/angular-ui/ui-select/pull/1210

PR #1052, opened 2015-07-01
#874 changes autocomplete="off" to autocomplete="false"

hugobessa:
changes `autocomplete="off"` to `autocomplete="false"` for not letting the browser autocomplete in chrome and firefox new versions


dimirc:
Will be solved at https://github.com/angular-ui/ui-select/pull/1210 Thanks!

PR #1068, opened 2015-07-07
Using cross-browser document scrollTop

goranprijic:
This is a fix for small bug that was noticed in Chrome: dropdown is always opened at the top of the element


skullcrasher:
+1 Could also use this fix.

itamarki:
+1. Thanks you!

sars:
+1 !

bilal414:
Thank you!

dimirc:
How can we reproduce the problem?

wesleycho:
This needs rebasing - this otherwise looks like a good fix.

longzheng:
+1 vote for this to be fixed too

Issue #1067, opened 2015-07-07
Programmatically reset search field


Sjors:
I'm using `reset-search-input="false"` in a [project](https://github.com/blockchain/My-Wallet-HD-Frontend/blob/5615b12c80b73b152de5ecee2205a3a5c015e55d/app/partials/send.jade#L44-L51) so that when the user types a search query, it is not cleared when the user focusses on a different form field. This is useful because the user needs to enter a bitcoin address (fairly long) and might make a typo.

However sometimes I do need to reset the field, for example when the user clicks a "Reset Fields" button.

I'll take a stab at this myself. The approach I am thinking of is as follows:

ui-select has an attribute `focus-on`. This allows the controller to broadcast a message `$scope.$broadcast('SetFocus')` and that then causes the field to receive focus.

I could add an attribute `reset-search-input-on` that removes the cached search upon `$scope.$broadcast('ResetSearch')`.

Does that make sense?

meribald:
Maybe I understood your problem wrongly but I wanted to tell what I do for this.
I just make the model undefined when I want to reset the ui-select. Let's say there is a ui-select like this:

<ui-select ng-model="myScopeModel.selected" dropdownvisible="true" reset-search-input="false">
....
...
</ui-select>

Setting myScopeModel.selected to undefined resets the ui-select.

PR #1076, opened 2015-07-09
Added reset-search-input-on attribute with demo.

Sjors:
This allows one to programmatically reset the search field. See #1067.


src/uiSelectController.js
```diff
@@ -49,8 +49,8 @@ uis.controller('uiSelectCtrl',
   };
 
   // Most of the time the user does not want to empty the search input when in typeahead mode
-  function _resetSearchInput() {
-    if (ctrl.resetSearchInput || (ctrl.resetSearchInput === undefined && uiSelectConfig.resetSearchInput)) {
+  function _resetSearchInput(force) {
+    if (force | ctrl.resetSearchInput || (ctrl.resetSearchInput === undefined && uiSelectConfig.resetSearchInput)) {
```
wesleycho: Did you mean `||`?

src/uiSelectDirective.js
```diff
@@ -132,6 +132,15 @@ uis.directive('uiSelect',
               });
           });
         }
+        
+        //Reset form based on scope event name (e.g. reset-search-input-on='SomeEventName')
+        if (angular.isDefined(attrs.resetSearchInputOn)){
+          scope.$on(attrs.resetSearchInputOn, function() {
+              $timeout(function(){
```
wesleycho: This seems fishy - why not just do `scope.$apply(function() { $select.clearSearchInput(); });`?

wesleycho:
This needs tests.

user378230:
Closing as don't think this would be more bloat than use. 

Willing to accept a PR to expose a way to reset the search input using $select if necessary. User's can then extend on this with a custom directive ([Accessing $select](https://github.com/angular-ui/ui-select/wiki/Accessing-%24select))

Issue #577, opened 2015-01-12
Tagging label problem with group-by option


idanb11:
Hi @dimirc ,
Great work...

Please have a look at the [demo](http://plnkr.co/edit/letVEHZG80WjIvDtoWP9?p=preview)

If you try to use ui-select with combination of tagging and group-by 
when you start to add a new custom tag the drop-down open with empty content and no tagging label are shown.
Everything else works fine besides that.

Idanb11

noam3127:
I've been experiencing the same issue. Any help would be appreciated.

angusoid:
Has anyone found any fixes for this issue?

PR #1081, opened 2015-07-11
Fix tagging with group-by (and unit-tests)

lukaw3d:
This pull-request includes:
* Fixed tagging to work with group-by (and added a unit test) (fixes https://github.com/angular-ui/ui-select/issues/577)
* Fixed a unit test `should allow decline tags when tagging function returns null` to catch an error and fixed that error (`Cannot set property 'isTag' of null`)
* Removed needlessly calling the tagging function (and added a unit test)
* Fixed a console error in [demo-tagging example](http://rawgit.com/angular-ui/ui-select/master/examples/demo-tagging.html). When entering `a` into Object Tags input, highlight filter replaced `<small>` with `<sm<span class="ui-select-highlight">a</span>ll>`

### DIFF
--- a/examples/demo-tagging.html
+++ b/examples/demo-tagging.html
@@ -86,10 +86,10 @@
   <p>Selected: {{multipleDemo.colors2}}</p>
   <hr>
 
-  <h3>Object Tags</h3>
+  <h3>Object Tags (with groupBy)</h3>
   <ui-select multiple tagging="tagTransform" ng-model="multipleDemo.selectedPeople" theme="bootstrap" ng-disabled="disabled" style="width: 800px;" title="Choose a person">
     <ui-select-match placeholder="Select person...">{{$item.name}} &lt;{{$item.email}}&gt;</ui-select-match>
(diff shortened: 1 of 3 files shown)


lukaw3d:
An example how tagging currently doesn't work with group-by:
http://rawgit.com/lukaw3d/ui-select/example-groupby-tagging/examples/demo-tagging.html
(`Object Tags (with groupBy)` heading)

An example how tagging works with group-by after these fixes:
http://rawgit.com/lukaw3d/ui-select/master/examples/demo-tagging.html
(`Object Tags (with groupBy)` heading)

lukaw3d:
I have rebased the commits to make it mergable again.

wesleycho:
This needs squashing and another rebase.

lukaw3d:
@wesleycho Rebased

Are you sure that commits should be squashed? Fixes have quite distinct scopes

user378230:
> Are you sure that commits should be squashed? Fixes have quite distinct scopes

I think you're right there, this LGTM once @wesleycho has reviewed I'll merge.

wesleycho:
Squashing makes rebasing easier if there's a conflict.

This LGTM to me too.

PR #1106, opened 2015-07-24
Fix bug: If the taggingtoken is ',', the ',' should not be shown on the tagging label

wangzhen-inf:
If the taggingtoken is ',', the ',' should not be shown on the tagging label http://plnkr.co/edit/EK1CA5ANtVVtzwkfFGtk


user378230:
Is there tests we can do for this? I'm thinking the initial problem should have been flagged by a failing test to begin with.

PR #1109, opened 2015-07-26
Don't hide input box when SearchEnable is false

cdjackson:
Fixes https://github.com/angular-ui/ui-select/issues/453

### DIFF
--- a/examples/bootstrap.html
+++ b/examples/bootstrap.html
@@ -42,20 +42,35 @@
     <fieldset>
       <legend>ui-select inside a Bootstrap form</legend>
 
-      <div class="form-group">
-        <label class="col-sm-3 control-label">Default</label>
-        <div class="col-sm-6">
+        <div class="form-group">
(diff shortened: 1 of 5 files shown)


dimirc:
Up and down key don't work

cdjackson:
Thanks - I’ll take a look…

cdjackson:
I've modified the concept slightly to allow the keys to work correctly. The problem was that by hiding the search box, we loose focus on the input. I've now changed this to instead add the 'ui-select-offscreen' class to the searchbox so that it still provides keyboard input, but is also hidden from the user.

I also added closing the select list if the list is open, and the user clicks on the match box again when in searchEnabled=false mode. Previously you had to click off the ui-select completely which seems wrong.

wesleycho:
History needs to be cleaned up here - there should not be any merge commits. It may be better to open a new PR with clean history.

cdjackson:
Squashed into a single commit.

PR #1110, opened 2015-07-26
Limit the maximum number of selections allowed in multiple mode

cdjackson:
This adds a ```limit``` attribute to the directive to limit the number of selections allowed in multi select mode.


dimirc:
Can you add a test?

cdjackson:
Added a test to ensure that we limit selection to 2. We start with one selection, add another and check we have 2, then add another and check we still only have 2 selections.

cambridgemike:
@dimirc Is this going out in the next maintenance release? If so, any ideas when that might be?

mbajur:
@dimirc i've made a fork with build for that feature: https://github.com/mbajur/ui-select (release: `0.12.1-mb`)

ganeshvelanki:
Is single select feature available ?

Issue #1059, opened 2015-07-03
Up and down arrow keys not working on select2 theme (version 0.12.0)


tolemac:
Can see this plunkr: 
http://plnkr.co/edit/q8fNxmGb1hqi3SuGdmeU?p=preview

Arrow keys work fine in Bootstrap and selectize but not in select2.

RyanDale:
The bug is happening because the dropdown is not focused. If you first click the search box to focus it, then the up/down arrow keys will work. I tested locally, and this issue (also referenced in #1053) should be fixed by #1040.

tolemac:
Then if you are in a form with several fileds and you are using the keyboard to enter data and using TAB to jump between fields ... when you arrive to ui-select ... have you take the mouse, point and click in search text box? It seems uncomfortable ...

RyanDale:
I wasn't suggesting that you should have to focus the select before using the arrow keys, just that the issue stemmed from the input not being correctly focused. Once the pull request I mentioned is merged, the dropdown will be correctly focused and the up/down keys should work as expected.

tolemac:
Ops! sorry, I misunderstood ... :(
Thanks

cbryer:
this happens when using search-enabled="false" attribute on the ui-select directive as well.  in that case there is no search box to focus and up/down never works.

dimirc:
Should be fixed with https://github.com/angular-ui/ui-select/pull/1040

cbryer:
unfortunately this is not fixed with #1040 because there is an ng-show="$select.searchEnabled" on the search input.  when searchEnabled == false the input has a style of display:none and can not get focus and the keyboard keys stop working.

PR #1114, opened 2015-07-29
solution for issue #1059 when searchEnabled is false.

cbryer:
when the search-enabled option on the ui-select directive is false the up, down and tab keys dont work.  this was because there was an ng-show="$select.searchEnabled" on the search input.  This change removes the ng-show and applies a css class to hide the search box using opacity and height styles.  Karma tests were updated and the following plunks demonstrate the changes.

previous:
http://plnkr.co/edit/zO8VFUUJTPnTRRh7u8kU?p=preview

with my changes:
http://plnkr.co/edit/3UpySUyaqi8fCbwB2OMi?p=preview


dimirc:
Could you remove /dist files from this commit?
Also can you add changes from https://github.com/angular-ui/ui-select/pull/1109 to avoid bootstrap theme to look weird?

cdjackson:
Just FYI, I've done a similar change in #1109 for the bootstrap theme. I used the 'offscreen' class to move  the search box off screen but still allow the focus to be set.

cbryer:
I merged #1109 into this pull request.  I liked the idea of using offscreen class however selectize didnt render nicely and select2 theme needed the class applied to a containing div instead of the search input so i opted to keep the new css class and remove the code from the controller in #1109 that added the class directly to the input.

cbryer:
is there anything else i should do for this pull request to get accepted?

Issue #973, opened 2015-05-28
Problem in rendering the dropdown when it renders in the up direction.


DhruvPrakash:
When the dropdown renders upwards and you search for something in the second half of the list, the result just floats in the air. You can replicate this issue by opening the bootstrap.html example and repeating the form group which contains the default implementation a bunch of times. Tested in chrome 41 and firefox 35.
 Please refer to the attached image.
![upward-render-issue](https://cloud.githubusercontent.com/assets/7776451/7861345/2ed145b6-056d-11e5-913c-00384770d255.png)

dcominottim:
+1

ramdog:
:+1: 

Also seeing this issue:
![image](https://cloud.githubusercontent.com/assets/5456178/9054069/1b6be798-3a32-11e5-9e26-0acc44112dd9.png)

cdjackson:
This is a css issue - I'll create a PR

PR #1123, opened 2015-08-04
Fix baseline when direction-up positioning is used

cdjackson:
When direction-up positioning is used, there was an issue where the select box would stay aligned at the top, and the height would reduce, meaning the bottom moved away from the top of the search box (see #973).

This PR removes the hardcoded top added in the code, and instead uses CSS positioning to position based on the absolute position at the bottom of the select.


jtrussell:
:+1:

Iku-turso:
+1.

stakach:
Works for me

jocelynduc:
Works for me too

jocelynduc:
Bug is fixed in current version of angular-ui-select (0.13.1)

wesleycho:
Is this already fixed?

This looks like an elegant fix otherwise.

yogeshgadge:
Any movement on this ?

andrew-stewart:
+1

andreiho:
This issue is still happening in the latest version.

PR #1132, opened 2015-08-07
fix(uiSelect): directives being created 2 times on ui-select-choices

### LINKED ISSUE #1000: Third party directives added to ui-select-choices is created 2 times
Seems `ui-select-choices` does extra work. Here is example. third party directive in any place is created once but in `ui-select-choices` it happens twice.
http://plnkr.co/edit/4nmGd7DxkDO6JuvONJ4D?p=preview

awerlang:
Fixes #1000


src/uiSelectChoicesDirective.js
```diff
@@ -5,6 +5,8 @@ uis.directive('uiSelectChoices',
   return {
     restrict: 'EA',
     require: '^uiSelect',
+    priority: 1,
```
wesleycho: This is a bad hack - one should never have to do priority: 1 to work around issues.

wesleycho:
Closing, as there are better fixes in open PRs.

Issue #1108, opened 2015-07-25
Is this project still being maintained?


cdjackson:
Activity from the maintainers seems very low and there are hundreds of PRs and issues.

Is this project still being maintained, or can someone suggest an alternative?

wald-tq:
I think this project is the best bet for a angular native implementation of a select2 field.

There have to be many developers which are using this directive and they know the way around the hard edges (including me).

Another alternative could be Angular Material with the nice chips: https://material.angularjs.org/HEAD/#/demo/material.components.chips

cdjackson:
Thanks. I agree - this is still a good project, I'm just concerned that it's not really being maintained and therefor it will quite soon become not such a good project :(

> There have to be many developers which are using this directive and they know the way around the hard edges (including me).

This is what concerns me :) I'm also making some modifications and have submitted a few PRs over the past few months, but there are now around 150 open PRs, and 470 or so open issues. It would be great if the maintainers no longer have the time to support it if others were able to...

The longer PRs are left unprocessed, the more difficult it is to merge them as there will be conflicts...

wald-tq:
I'm also a bit concerned. I use it throughout my project and I don't have a better choice for now.

Also see this discussion: https://github.com/angular-ui/ui-select/issues/1096

ProLoser:
Are either of you interested  in taking this project over?

sars:
@ProLoser , I'm interested also in development this project. Ready to help

dimirc:
@sars one way to help could be to try checking duplicated issues and posible PRs

cdjackson:
> Are either of you interested in taking this project over?

I would be happy to help with the project. I think since this is such a popular project that it needs a few people to help maintain it as it needs to be regularly maintained. Currently, we go for months without any activity from the maintainers, and hence we end up with a totally unmanageable number of PRs and Issues... I don't mean to criticise, people have a lot of demands on time these days, but the project can't thrive if nothing changes for months on end...

I'm not currently hugely familiar with the source, but would be willing to help maintain it. I would be happy to initially spend a few days going over the issues found in #1096. From there I would suggest it might be easier to close "all" old issues (say, older than 4 months) - or at least request that it is still valid. Likewise for PRs - many already have conflicts and need to be updated against the last codebase. This could allow the project to be brought back into a manageable situation where a few people should be able to keep on top of things every day or three (and I would be happy to support this too)...

[...]

PR #1134, opened 2015-08-07
Refactor branch

ProLoser:
Continuation of #1108 refactor. Will try to give feedback here.


ProLoser:
This diff is way too large to review. @cdjackson I think you may have to just develop on a branch in here

cdjackson:
Fixes #848

cdjackson:
@ProLoser currently I've changed the callbacks to use controller stubs (it possibly needs more modularising yet, but...). The reason I went down this route is that for some parts of the selection where overriding the default implementation is expected, I think it's useful to support promises where async operations is likely (eg the beforeDropdownOpen where the dev might want to populate the list from a server or something).

Thus, if the dev wants to override these callbacks for basic callback functions, it's very simple, and there's a clean interface that provides a good level of functionality. If they want to do something fancier, then they can override a lower level method, but then their a little more on their own as this won't be documented.

I've also split the tagging out into a separate directive - I wouldn't call it refined, but it functions ok from my initial tests (it needs some more work).

I had to add the KEY.MAP back into the code to cover tagging - I could have put it into the tagging directive since this is the only place it's currently needed, but it means other directives wouldn't be able to use it... Previously this was in the global space which I didn't like. I might revisit this as a trawl through the issues list tonight revealed a problem in this area with keycode mapping...

There's still some work to do on this, but I thought I'd throw out an update... I'm not sure how you want to proceed? It needs more tests, and it needs docs, and probably more of a tidy, It would be good if we could get a few more people commenting/testing this - my view is it would be good to get it merged, close all outstanding issues and request people resubmit if needed (as we head toward 500 issues and 180 PRs, it's unmanageable).

ProLoser:
@cdjackson I understand a desire to support promises, but that arrises from treating the list of items as something private and very precisely controlled. If you make these items public, then the user can modify this collection anytime they wish, they can do ajax and just update the collection anytime they want, the support for promises because superfluous.

Again, I still recommend not using before and after callbacks and simply demonstrating how to override extremely simplified functions.

ProLoser:
For now to make the code easier to review, can you remove the dist folder? It's just bloating this PR tremendously, we can deal with it after all the src is reviewed.

ProLoser:
I also feel that the repeater logic needs to be removed. It just bloats and complicates everything. Instead of allowing people to define the template for one row and handling the repeating for them, just move the logic up one level and let people define the entire dropdown container.

cdjackson:
> If you make these items public, then the user can modify this collection anytime they wish

True, but they might still want this loading to be triggered when the dropdown is opened...  Anyway, I appreciate there's many ways to achieve this so I'll take another look - it might be another few days (other things to do and all that :) ).

I'll remove the dist files now though.

ProLoser:
If you look at how I do it in ui-mention, you can override the 'findChoices' method and set your own loading flag and use that alongside the list of choice,s

src/addons/uiSelectTaggingDirective.js
```diff
@@ -0,0 +1,190 @@
+angular.module('ui.select.tagging', ['ui.select'])
+    .directive('uiSelectTagging',
+    ['$parse', '$timeout', function () {
```
ProLoser: You shouldn't be using this injection syntax.

Issue #788, opened 2015-03-23
How to validate the select input?


burzum:
I've already spent some time on trying to figure out how to do this without much success:

How can I validate an input to be not empty using the standard angular form validation mechanism?

Sylox:
+1

JobaDiniz:
+1.

I've tried to change the code to also pass `name`parameter, however nothing worked:
Angular 1.2.28

**select template**
```
<div class="ui-select-container ui-select-bootstrap dropdown" ng-class="{open: $select.open}">
    <div class="ui-select-match"></div>
    <input type="text" autocomplete="off" tabindex="-1" id="{{$select.id}}" name="{{$select.name}}"
           aria-expanded="true"
           aria-label="{{ $select.baseTitle }}"
           aria-owns="ui-select-choices-{{ $select.generatedId }}"
           aria-activedescendant="ui-select-choices-row-{{ $select.generatedId }}-{{ $select.activeIndex }}"
           class="ui-select-search"
           placeholder="{{$select.placeholder}}"
           ng-model="$select.search"
           ng-show="$select.searchEnabled && $select.open">
    <div class="ui-select-choices"></div>
</div>

```
**uiSelectMatch**
```
directive('uiSelectMatch', ['uiSelectConfig', function(uiSelectConfig) {
            return {
                restrict: 'EA',
                require: '^uiSelect',
                replace: true,
                transclude: true,
                templateUrl: function(tElement) {
                    // Gets theme attribute from parent (ui-select)
                    var theme = tElement.parent().attr('theme') || uiSelectConfig.theme;
                    var multi = tElement.parent().attr('multiple');
                    return theme + (multi ? '/match-multiple.tpl.html' : '/match.tpl.html');
                },
                link: function(scope, element, attrs, $select) {
                    $select.lockChoiceExpression = attrs.uiLockChoice;
                    attrs.$observe('placeholder', function(placeholder) {
                        $select.placeholder = placeholder !== undefined ? placeholder : uiSelectConfig.placeholder;
                    });
                    attrs.$observe('name', function(name) {
                        $select.name = name !== undefined ? name : uiSelectConfig.name;
                    });
                    attrs.$observe('id', function(id) {
                        $select.id = id !== undefined ? id : uiSelectConfig.id;
                    });

                    $select.allowClear = (angular.isDefined(attrs.allowClear)) ? (attrs.allowClear === '') ? true : (attrs.allowClear.toLowerCase() === 'true') : false;

[...]

PR #1141, opened 2015-08-11
[WIP] Add name attribute

crearc:
Name attribute is required to do validation as part of a form.

I'm still playing with it.

fixes https://github.com/angular-ui/ui-select/issues/788


awerlang:
Don't think this is gonna work.
Inner `input` have a separate ngModel. If you attach the same name, it would overwrite `ui-select` ngModel in form scope. That could work, but I think the original ngModel will stop publishing validation errors under `form.control`. Not quite sure though.

Have you noticed any problem with that approach?

MohammedSiddiqui:
simply giving a 'name' attribute on the <ui-select> tag still doesn't work. Upon inspecting my element, I can that the name attribute is attached to a 'div' instead.

PR #1142, opened 2015-08-12
Fix focus when append to body is true

ivanzotov:
When append-to-body is true for multiple select it takes time to replace element back, so we lose caret position and focus. I think we can just set timeout for that.


wesleycho:
Closing due to changing the timeout this way is not a good fix - unnecessary addition of time in the delay is a typical sign that what is happening is not understood, and will encourage other fixes to be done this way, having a cascading effect.

Feel free to do a further investigation and craft a more proper fix.

PR #1139, opened 2015-08-11
Significant speed improvement on opening typeahead

timgl:

PR #1149, opened 2015-08-17
Significant speed improvement on opening typeahead

timgl:


wesleycho:
Remove changes to `dist`, then squash commits then rebase.

user378230:
I don't see any differences between this and #1139 (which seems to be in better shape anyway) so I'm going to close this.

PR #1163, opened 2015-08-24
Fix direction-up in webkit

littlebigbot:
Like this: https://github.com/angular-ui/ui-select/pull/989, but confirmed to work in Firefox


src/uiSelectDirective.js
```diff
@@ -273,7 +273,7 @@ uis.directive('uiSelect',
               var offsetDropdown = uisOffset(dropdown);
 
               // Determine if the direction of the dropdown needs to be changed.
-              if (offset.top + offset.height + offsetDropdown.height > $document[0].documentElement.scrollTop + $document[0].documentElement.clientHeight) {
+              if (offset.top + offset.height + offsetDropdown.height > $document.scrollTop() + $document[0].documentElement.clientHeight) {
```
littlebigbot: This is the only relevant change, the rest is the result of running `gulp`

Stevebel:
This fixed the issue for me, please bring into master

danielmouraa:
+1

dimirc:
Can you remove /dist files from commit and just leave the /src files changes?

dimirc:
Btw, did you tried on IE also?

dimirc:
@littlebigbot `$document.scrollTop()` isn't supported by [jqLite](https://docs.angularjs.org/api/ng/function/angular.element) and we dont want to have dependeny on jQuery

Anyway thanks for checking this, I just pushed a PR that should solved https://github.com/angular-ui/ui-select/pull/1212 this problem

PR #1168, opened 2015-08-28
On collection change update position of upward dropdown (fixes #1167)

### LINKED ISSUE #1167: Wrong top position of choices when edit value
Hi,

I´m using your ui-slect and it is a great component.
I´m using it a I belive to find a bug. 
When ui-slect shows the choices on top of the input field an I enter some value, the box with choises reduces its hight, but not its position. So the choices are somwhere over the top of my component.
Here is my code:
 
ui-select-full
ui-select-edit
```
<div class="form-group required">
    <label for="countryCB" class="control-label">{{'PRODUCT_COUNTRY_SELECT' | translate}}</label>
    <ui-select id="countryCB" ng-model="vm.selectedCountry" ng-disabled="disabled" title="{{'PRODUCT_COUNTRY_SELECT' | translate}}">
        <ui-select-match placeholder="{{'PRODUCT_COUNTRY_SELECT' | translate}}">
            <span ng-bind-html="$select.selected.country"></span>
            <small ng-bind-html="$select.selected.countryIso"></small>
        </ui-select-match>
        <ui-select-choices repeat="country in vm.countries | filter:{country:$select.search}">
            <span ng-bind-html="country.country | highlight: $select.search"></span>
            <small ng-bind-html="country.countryIso | highlight: $select.search"></small>
        </ui-select-choices>
    </ui-select>
</div>
```
Is it a bug, or am I wrong?

gajdusek-egston:
Hi,

This commit should fix issue #1167.

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -256,6 +256,21 @@ uis.directive('uiSelect',
         var dropdown = null,
             directionUpClassName = 'direction-up';
 
+        var repositionUpwardDropdown = function () {
+          if (!$select.open || !element.hasClass(directionUpClassName)) {
+            return;
+          }
+
+          if ($select.open) {
+            dropdown = angular.element(element).querySelectorAll('.ui-select-dropdown');
+            if (dropdown === null) {
+              return;
+            }
+            var offsetDropdown = uisOffset(dropdown);
+            dropdown[0].style.top = offsetDropdown.height * -1 + 'px';
+          }
+        };
+
         // Support changing the direction of the dropdown if there isn't enough space to render it.
         scope.$watch('$select.open', function(isOpen) {
           if (isOpen) {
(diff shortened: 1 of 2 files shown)


grahamlittle:
hi, 
i am sorry if i am asking stupid question but, issue says Open so does this mean it has not yet been merged into a release? 
Many thanks

src/uiSelectDirective.js
```diff
@@ -274,9 +289,14 @@ uis.directive('uiSelect',
...
               // Determine if the direction of the dropdown needs to be changed.
               if (offset.top + offset.height + offsetDropdown.height > $document[0].documentElement.scrollTop + $document[0].documentElement.clientHeight) {
+                element.addClass(directionUpClassName);
+                offsetDropdown = uisOffset(dropdown); // recalculate after class changed
                 dropdown[0].style.position = 'absolute';
                 dropdown[0].style.top = (offsetDropdown.height * -1) + 'px';
-                element.addClass(directionUpClassName);
+                // Reposition upward dropdown each time the collection of items changes
+                scope.$watchCollection('$select.items', function() {
+                  $timeout(repositionUpwardDropdown);
```
wesleycho: There is a big optimization that can be done here. Instead of doing a $timeout everytime it changes, scheduling an operation with `scope.$$postDigest` would batch it. This requires the addition of a boolean check inside the $watchCollection callback, and resetting the flag when the function is executed.

Something like
```js
var foo = false;
scope.$watchCollection('$select.items', function() {
  if (!foo) {
    scope.$$postDigest(function() {
      foo = false;
      repositionUpwardDropdown();
    });
  }
});
```

wesleycho:
This needs rebasing.

longzheng:
+1 vote for this to be fixed too

tbugge-amplero:
+ 1 this has been an issue for us for quite a while!

PR #1169, opened 2015-08-28
Do not hide placeholder when multiple=true and some items are selected.

stulentsev:
For some reason, placeholder was hidden in scenario described in the title. It seems like an oversight.

##### Step 1
ui-select1

##### Step 2
ui-select2

Now UX is a bit better, users don't have to wonder what is that big white empty area below selected items. :)


stulentsev:
turns out that search field should not be on a new line in this case. Debugging this problem now.

pierregaste:
I need to have the same behavior.
Maybe it could be an option ```disable-placeholder: [true | false]```?

Issue #1007, opened 2015-06-12
Wrong calculation of dropdown position


StarpTech:
There you can find the bug: https://github.com/angular-ui/ui-select/blob/2d1422d109dc1cd64213271e59506b14f74d356f/dist/select.js#L1032

My dropdown is show at the top even though I have enough space. When I use autocomplete the result is reduced, however the dropdown position is worng:

![issues](https://cloud.githubusercontent.com/assets/1764424/8123198/a2522f4c-10c5-11e5-8415-79aa1489b32c.png)

skullcrasher:
Could be resolved with PR #1068

marazt:
Hi, what about the solution? There is/was a pull request for it - #1170

SpencerKaiser:
I experienced this exact issue after including Angular Animate as a project dependency... once I removed it, this issue was resolved. If anyone has this same issue in the future, please search for ui-select issues mentioning angular animate.

PR #1170, opened 2015-08-28
Fix for when dropdown/search is on top

SamHChowdhury:
Cleaned up the interaction for when the dropdown/search is on top of the button.

At the moment, the size of the dropdown list shrinks after it is set on top of the button. This is an issue because it is not repositioned and ends up floating. This is taken care of with this pull. (Possibly related to issue #1007)


Magador:
:+1:

m0ppers:
:+1: :tada:

SamHChowdhury:
No point in keeping pull request open if it most likely won't be addressed

PR #1182, opened 2015-09-08
Avoid inline execution of javascript in choices template.

Sjors:
If a server sets the `content-security-policy` header to `script-src 'self' 'unsafe-inline';` all works well, but without `unsafe-unline`, the browser will throw an error: `Refused to execute JavaScript URL because it violates the following Content Security Policy directive:`.

There is no stack trace, but it turns out this is caused by `href="javascript:void(0)"` in the choices template. 

Pull request #395 proposed removing that piece Javascript altogether, but apparently there were side effects.

I'm replacing it with `ng-click="$event.preventDefault()"`, based on a similar fix in the [ui.bootstrap](https://github.com/angular-ui/bootstrap/commit/fb302c6086543420b3561be3ed5c49ac3974c243) project. This makes the CSP error go away.

All tests pass. I tried all examples in Safari and couldn't see anything obviously wrong, but I'm not sure what to look for.


Sjors:
Is there something I should improve in this PR to make it more likely to get merged?

PR #1183, opened 2015-09-09
fix(multiple): text input not resized on window resize

romario333:
Text input within multiple select is not resized correctly on window resize. Open this plunkr and try to resize window. Notice that text input (with green background) is not resized:

http://embed.plnkr.co/cqsb8nOIunj2rMU36z5H/preview

Here's the demo with this fix applied: http://embed.plnkr.co/pPdeELl80hgCOVZKCs6a/preview

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -1,4 +1,4 @@
-uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelectMinErr, $timeout) {
+uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', '$window', function(uiSelectMinErr, $timeout, $window) {
   return {
     restrict: 'EA',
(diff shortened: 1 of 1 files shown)


dangoodman:
I have just made a PR fixing the same issue, just with a more general approach — https://github.com/angular-ui/ui-select/pull/1252 .

src/uiSelectMultipleDirective.js
```diff
@@ -399,6 +399,14 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
         });
       });
 
+      function onWindowResize() {
+        $select.sizeSearchInput();
+      }
+
+      angular.element($window).on('resize', onWindowResize);
```
wesleycho: This needs to be debounced.

wesleycho:
This needs tests.

user378230:
This appears to have been implemented somewhere else and subsequently improved on by #1536 and #1538. 

Closing.

PR #1188, opened 2015-09-10
Fixed determination of dropdown direction

iKBAHT:
Now dropdown open to down if there enough place that dropdown will be entirely visible. If no it open to up.


wesleycho:
This should remove changes to `dist` & rebasing after.

Issue #1192, opened 2015-09-13
resolving simple values in uiSelectMultipleDirective.js 


noblecraft:
Currently uiSelectMultipleDirective, which is used if multiple is set, either uses the ngModel array (simple values), or the $parsed 'data' array for ngModel parsing / formatting (to/from view).

This means the 'data' array needs to be populated. 

This is a problem if I have pre-existing (simple) values selected for ngModel. I don't want to populate the entire data array (e.g. if my data array is huge). I can't just populate the data array with the corresponding complex values, since I don't want the user to be able to select duplicates.

I think we need to add an attribute e.g. `<ui-select dict='someDictionaryPromise'>`, and add another conditional branch to lookup / parse value by `dict`?

noblecraft:
created a PR: https://github.com/angular-ui/ui-select/pull/1193

this is not production ready code, feedback sought about the idea.

thanks

noblecraft:
This might be a duplicate of #404 

I think comment by https://github.com/sporkd hit nail on its head there.

cmichal:
+1

PR #1193, opened 2015-09-13
Noblecraft/1192

noblecraft:
this is a proposed fix for issue #1192 

sorry its a bit messy, first time contributing to ui-select.


src/uiSelectDirective.js
```diff
@@ -19,7 +19,7 @@ uis.directive('uiSelect',
 
       //Multiple or Single depending if multiple attribute presence
       if (angular.isDefined(tAttrs.multiple))
-        tElement.append("<ui-select-multiple/>").removeAttr('multiple');
+        tElement.append("<ui-select-multiple dict=\"" + tAttrs.dict + "\"/>").removeAttr('multiple');
```
wesleycho: Should switch to single quote strings to avoid the awkward escaping of the double quote.

wesleycho:
This needs changes to dist reverted, rebasing, and tests added for this new feature.

user378230:
Closing as I don't think this is a particular good solution to the problem. Let's see if anyone else is able to come up with something else.

PR #1199, opened 2015-09-18
stop on-select trigger when the limit is passed

sn0opr:
```
if ($element.attr('limit') == undefined || $element.attr('limit') > ctrl.selected.length) {
                  $scope.$broadcast('uis:select', item);

                  var locals = {};
                  locals[ctrl.parserResult.itemName] = item;

                  $timeout(function(){
                    ctrl.onSelectCallback($scope, {
                      $item: item,
                      $model: ctrl.parserResult.modelMapper($scope, locals)
                    });
                  });
```


wesleycho:
Going to close this as this has changes that pollute dist, bower.json and package.json, and there are other PRs that tackle this open.

zacyang:
May I ask which PR fix this issue, I have the same problem with the attr.limit  behaviours.

PR #1208, opened 2015-09-27
feat(choices): use object as source iterating over its properties

dimirc:
We'll be able to use an object as the source for the choices (iterating over its properties)

Demo [plunker](http://plnkr.co/edit/9DsLidTZR7RJDEq6PW0H?p=preview])

Pending:
- [x] Need to add a watch to update choices when source object changes

### DIFF
--- a/examples/demo-object-as-source.html
+++ b/examples/demo-object-as-source.html
@@ -1,0 +1,107 @@
+<!DOCTYPE html>
+<html lang="en" ng-app="demo">
+<head>
+  <meta charset="utf-8">
+  <title>AngularJS ui-select</title>
+
+  <!--
+    IE8 support, see AngularJS Internet Explorer Compatibility http://docs.angularjs.org/guide/ie
+    For Firefox 3.6, you will also need to include jQuery and ECMAScript 5 shim
+  -->
+  <!--[if lt IE 9]>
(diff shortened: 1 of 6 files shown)


amcdnl:
Added docs here - https://github.com/angular-ui/ui-select/wiki/ui-select-choices

amcdnl:
@dimirc It also does not work with 'nested' objects. The following setup:

		<ui-select ng-model="actionCtrl.model.action.delimeterType" style="width:100%">
			<ui-select-match>{{$select.selected}}</ui-select-match>
			<ui-select-choices repeat="type as (type, name) in actionCtrl.delimiterTypes | filter: $select.search">
			  <div ng-bind-html="name | highlight: $select.search"></div>
			</ui-select-choices>
		</ui-select>

tries to bind all the properties of `actionCtrl`. http://www.screencast.com/t/jMN4BZiI

amcdnl:
@pdegenhardt @dimirc I have resolved the issues and release 0.13.2

Avien:
Can you please check out issue #1255  I opened yesterday?
I have the exact same problem you described and upgraded to v0.13.2 but still not working
Thanks

samuel-weber:
Hi amcdnl
Your Change #1208 broke the track by expression
Example: n in [42, 42, 43, 43] track by $index

https://github.com/angular-ui/ui-select/pull/1208

Greets Samuel

PR #1210, opened 2015-09-28
fix(templates): autocomplete off to false

dimirc:


empirecoder:
autocomplete="false“ not working on Internet Explorer 11, "off" works

Issue #88, opened 2014-06-25
Performance issues


filipedeschamps:
Guys, we are suffering some serious performance issues. Does anyone know what is this **$select.isActive(this)** ?

![Performance problem](http://i.imgur.com/L9Q868J.png?1)

gjoliver:
yeah guys, this is really bad.
I have only 1 select with maybe 500 entries. And it's super slow.
any idea where the problem may be?

gjoliver:
ok I worked around this by doing repeat="...... | limitTo: 100", so we don't show a huge list of items.

oaprograms:
@gjoliver 
limitTo solves the speed issue, but it would be even nicer if remaining items could be appended automatically when the list is scrolled to bottom.

rogier21:
Any update on this? This is kind of a dealbreaker for me. I am migrating from Kendo, and have a dropdown with over 5000 items. In Kendo it opens, loads and filters in ms, in ui-select it's sluggish and often freezes the browser.

aturkelson:
I'm having the same issue with around 500 items.  Sometimes the performance gets so bad that the browser actually crashes!

KevinPHughes:
Also a problem for me. I can't do the limitTo: 'myNumber' trick because it doesn't allow previously selected items to be pre-populated unless they're within the first 'myNumber' of my source.

mareczek:
Very recently (~ week ago) performance dropped even further. I have problems with a list of 103 items.

adamkidder:
I recently switched from ui-select to angular-selectize due to performance. I'm dealing with lists of ~500 entries each and typing to filter is unacceptably slow.

buremba:
@mareczek +1

vitto32:
what about "suspending" the watchers instead of using ngIf (which stresses DOM on long lists)?
something like: https://coderwall.com/p/d_aisq/speeding-up-angularjs-s-digest-loop

igorlino:
+1

pitieu:
The performance issues are due to the ng-repeat mechanism in the ui-select-choices directive.  Like askhogan said the dropdown list are getting processed immediately. 
Also i added a custom filter and whenever i hover over a dropdown list item, it calls filter for every item in the dropdown list and every time i hover over another item it renders the filter again for every single item in the dropdown list. (at 20 items it already starts to get slow)

vitto32:
[...]

PR #1211, opened 2015-09-28
refactor(choices): avoid setting activeItem on mouseenter

dimirc:
We were using `ng-mouseenter` to be able to set the hovered row as the active item, but this caused some lagging as `ng-mouseenter` was calling for a full $digest.

Could be better just to show a simple hover background change (using css) and avoid doing a $digest.

Demo [plunker](http://plnkr.co/edit/1Nl9SBsc1yoxO3xDSZQk?p=preview)

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -17,6 +17,11 @@
   top: 0px !important;
 }
 
+
+.ui-select-choices-row:hover {
+  background-color: #f5f5f5;
+}
+
 /* Select2 theme */
 
 /* Mark invalid Select2 */
--- a/src/uiSelectChoicesDirective.js
+++ b/src/uiSelectChoicesDirective.js
@@ -41,7 +41,6 @@ uis.directive('uiSelectChoices',
 
         choices.attr('ng-repeat', RepeatParser.getNgRepeatExpression($select.parserResult.itemName, '$select.items', $select.parserResult.trackByExp, groupByExp))
             .attr('ng-if', '$select.open') //Prevent unnecessary watches when dropdown is closed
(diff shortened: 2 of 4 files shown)


src/common.css
```diff
@@ -17,6 +17,11 @@
   top: 0px !important;
 }
 
+
+.ui-select-choices-row:hover{
```
jtheoof: code consistency: add space before `{`

dimirc: @jtheoof done

jtheoof:
I might be wrong, but after a quick `git-grep` it looks like the only place where `setActiveItem` was being called was from the `ng-mouseenter` event. Perhaps we can get rid of it in `uiSelectChoicesDirective.js`.

dimirc:
@jtheoof yes, its not needed anymore. I just removed it

jtheoof:
:+1: Not sure what went wrong with the build.

dimirc:
Y just rebuilt it and Travis is fine now

jtheoof:
Cool, I think it's good to go. Might help #88 but I do not think it's the main performance issue.

sonphnt:
Hi,

When my ng-model is null at first with placeholder "Please select". Do we have any option to turn off active blue background on the first item of the list. Because that make people confused when user click to open the dropdown and see the first item selected with blue background.

The blue background on item should only be highlighted if ng-model has value when user select one item. Otherwise, we should leave all items in the dropdown list with white background as normal.

Thanks

kaufi07:
+1

PR #1212, opened 2015-09-28
fix(dropdown): cross browser dropdown position

dimirc:
To avoid incorrect dropdown position

Wrong [Demo](http://plnkr.co/edit/oZCudVaPNc42B7hLYB7j?p=preview)

Corrected [Demo](http://plnkr.co/edit/vElyF6gmKYXA6jiM1K1k?p=preview)

*To reproduce, scroll down to red field*

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -142,6 +142,7 @@ uis.controller('uiSelectCtrl',
     ctrl.itemProperty = ctrl.parserResult.itemName;
 
     ctrl.refreshItems = function (data){
+      $scope.calculateDropdownPos();
       data = data || ctrl.parserResult.source($scope);
       var selectedItems = ctrl.selected;
       //TODO should implement for single mode removeSelected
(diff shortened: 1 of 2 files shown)


ramdog:
Hey, nice work.  Indeed this looks good in the case where the dropdown has room below:
image


However, it seems the case where the dropdown is opening above, this is still an issue (to reproduce, scroll down so that the red field is just above the bottom then type in `admin2`):
image

dimirc:
@ramdog I cannot reproduce it. I just scrolled until red field is at bottom and then tried to open dropdown but I see it showing above correctly. In what browser did your tried? Can you explain me better how to reproduce it?

Chrome
image

Firefox
image

ramdog:
Hey @dimirc, it does open correctly, but from your screenshots it looks like you may have missed the last step in testing.  

1.  Do exactly what you did (scroll down just enough to force the dropdown to open up).
2.  Then type out `admin2` - the whole thing.  Notice the **`Admin2`** option is indeed selected but still floating way up. Instead, it should be just above the search input.

dimirc:
@ramdog got it. Pls recheck the corrected demo and it should be working correctly now

ramdog:
Looks great, nice work!
:+1:

estevaolucas:
Nice. It's working.

GitHubUserNameX:
Nice work. Is their also the intention to make this work in the directions left and right?

PR #1213, opened 2015-09-29
feat(choices): allow to set default dropdown position

dimirc:
We should be able to set dropdown position explicitly so I added a new attribute called `position` for the `<ui-select-choices>` tag, that will accept 3 values: *up*, *down* and *auto*

Sometimes we might not want the dropdown to go automatically up even when there isn't enough space, so now we can force position to always go to an specific direction.

Also you can specify globally at `uiSelectConfig` using `dropdownPosition`

Demo [plunker](http://plnkr.co/edit/J8plv8vnnrzFV61Pkpls?p=preview)

### DIFF
--- a/examples/demo-dropdown-position.html
+++ b/examples/demo-dropdown-position.html
@@ -1,0 +1,168 @@
+<!DOCTYPE html>
+<html lang="en" ng-app="demo">
+<head>
+  <meta charset="utf-8">
+  <title>AngularJS ui-select</title>
+
+  <!--
+    IE8 support, see AngularJS Internet Explorer Compatibility http://docs.angularjs.org/guide/ie
+    For Firefox 3.6, you will also need to include jQuery and ECMAScript 5 shim
+  -->
+  <!--[if lt IE 9]>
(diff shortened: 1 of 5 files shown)


isimonov:
Thanks a lot! My example works correctly http://plnkr.co/edit/Tzcp1q2WV7HZCp5meftC?p=preview

karptonite:
This is actually not working properly for me.
http://plnkr.co/edit/NuSM3OfJUgbxCRYoln8z?p=preview
The role selector is set to position="true" and the select options go down at first, but then reverses if you type a letter, the delete it.
http://quick.as/wy2mc3zva

dimirc:
@karptonite I found the bug and will push a new patch release today

karptonite:
@dimirc Thanks! Your work on this is certainly appreciated!

dimirc:
@karptonite v0.13.1 released and now your issue should be fixed. Check updated demo http://plnkr.co/edit/H6ZSPuZ6qnEEJAF4dFrA?p=preview

Issue #1195, opened 2015-09-15
Name, label and for attribute


cyrilchaponeverysens:
I'm struggling on how to emulate this behavior with *ui-select* component ? I mean the fact that clicking on label focuses the select ?

```
<label for="testInput">Test input</label>
<select name="testInput">
  <option>Test option</option>
</select>
```

PR #1220, opened 2015-09-30
add input id to search <input> if present on ui-select directive as input-id

arkarkark:
fixes https://github.com/angular-ui/ui-select/issues/1195
fixes https://github.com/angular-ui/ui-select/issues/236

### DIFF
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -78,7 +78,7 @@ describe('ui-select tests', function() {
       { name: 'Nicole',    email: 'nicole@email.com',    group: 'bar', age: 43 },
       { name: 'Natasha',   email: 'natasha@email.com',   group: 'Baz', age: 54 }
     ];
-  
+
     scope.peopleObj = {
       '1' : { name: 'Adam',      email: 'adam@email.com',      age: 12, country: 'United States' },
       '2' : { name: 'Amalie',    email: 'amalie@email.com',    age: 12, country: 'Argentina' },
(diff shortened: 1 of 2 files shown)


arkarkark:
basically a copy of @georeith 's #237 #238

dimirc:
Can you add one simple test to it so we can merge asap?

arkarkark:
tests added!

(sorry for the whitespace cleanup, my editor couldn't help itself).

test/select.spec.js
```diff
@@ -2093,6 +2095,20 @@ describe('ui-select tests', function() {
...
+
+    it('should add an id to the search input field', function () {
+      var el = createUiSelectMultiple({inputId: 'inid'});
+      var searchEl = $(el).find('input.ui-select-search');
+      expect(searchEl.length).toEqual(1);
+      expect(searchEl[0].id).toEqual('inid');
+    });
+  });
+
+  it('should add an id to the search input field', function () {
```
dimirc: Looks like same test is repeated twice

test/select.spec.js
```diff
@@ -127,6 +127,7 @@ describe('ui-select tests', function() {
       if (attrs.title !== undefined) { attrsHtml += ' title="' + attrs.title + '"'; }
       if (attrs.appendToBody !== undefined) { attrsHtml += ' append-to-body="' + attrs.appendToBody + '"'; }
       if (attrs.allowClear !== undefined) { matchAttrsHtml += ' allow-clear="' + attrs.allowClear + '"';}
+      if (attrs.inputId !== undefined) { attrsHtml += ' input-id="' + attrs.inputId + '"'; }
```
dimirc: Instead of `inputId` what about only `id`?

arkarkark: You can't have two elements with the same id and I felt weird about removing the id from the ui-select because people might be using that to select their css rules.

test/select.spec.js
arkarkark: I wanted to test both single and multiple because they are very different beasts (even though they both have the same text input name right now). The `it` name is the same because the `describe` names will disambiguate them. I'm happy to remove one if you want.
I'd prefer to keep the multiple test because that's the one I care about most (and that search field is always on display so much more likely to be used for `<label for="...">`.

PR #1236, opened 2015-10-07
Fixes for absolute positioning respective to offset parent

ivi-hamiti:
Issues when the in modals and scrolling as the choices are appended to the body


ivi-hamiti:
after some checkups in the project append to body was set always to true and that was the issue the dropdown was not positioned correctly

PR #1238, opened 2015-10-08
Modified gulp tasks to generate sourcemaps for min.css and min.js

### LINKED ISSUE #1107: Add map files
Please add map files for min.js,min.css files in distributive

anilanar:
Fixes #1107.

## Additional Notes
- `gulp-rename` dependency is removed because it didn't work well with `gulp-sourcemaps` even though it is supported. `gulp-concat` is used instead.

- Upgraded `gulp-concat` to the latest version: So I needed to upgrade this due to a bug in previous versions.

- ~~Alphabetically sorted dependencies in `package.json`, which is the default behaviour when `npm install` is used with `--save` or `--save-dev`.~~


wesleycho:
This needs rebasing - this seems like a useful feature though.

anilanar:
OK, I'll work on rebasing which shouldn't be much trouble.

anilanar:
@wesleycho Done rebasing, tested, seems to work fine.

user378230:
@aaronroberson the builds are your side of things so I'll leave this one to you :smiley:

wesleycho:
This LGTM.

PR #1276, opened 2015-10-27
Add proper paste support

erikdubbelboer:
Add paste="expression" to allow custom paste handling.

Allow pasting in tagging mode when the tagging function is not defined.

In IE use window.clipboardData so jQuery is not required.

Fixes #910, #704, #789, #848, #429

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -16,6 +16,7 @@ uis.controller('uiSelectCtrl',
   ctrl.searchEnabled = uiSelectConfig.searchEnabled;
   ctrl.sortable = uiSelectConfig.sortable;
   ctrl.refreshDelay = uiSelectConfig.refreshDelay;
+  ctrl.paste = uiSelectConfig.paste;
 
   ctrl.removeSelected = false; //If selected item(s) should be removed from dropdown list
   ctrl.closeOnSelect = true; //Initialized inside uiSelect directive link function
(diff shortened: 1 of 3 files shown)


logannc:
Is there a reason that `ctrl.searchInput.value` is not used instead of the paste buffer? If not, you might miss things that are half-typed, half-pasted.

ErikDubbelboer:
@logannc you are correct, I changed it so that the current input is prepended to the pasted data before parsing.

danielgrippi:
bump. can we get this merged?

aaronroberson:
Published in v0.14.7 if you would like to confirm.

Hypercubed:
This is a useful feature and should be be documented.  I suggest the following added to the wiki:

| option | description | value | default |
|---|---|---|---|
| `paste` | Accepts a string which is a scope function. The function will be passed the new item as a string and should return an object which is the transformed value to be pushed to the items array. Returning false (or undefined) skips pushing to the items array.  | string ($scope function)| undefined |

Hypercubed:
Actually, looking again it appears the return value is ignored..

user378230:
@Hypercubed you should be able to update the wiki yourself, your contribution would be welcome! :smiley:

Hypercubed:
Sure, was going to do that first... but wanted to make sure I understood the API.  My first assumtion was not correct.

PR #1279, opened 2015-10-28
Add inForm attribute

oslo10e:
When in-form="true" is set in ui-select, then when pressing on ENTER key and if dropdown list is not open, it does not open dropdown list and event propagation is not stopped, so parent form "submit" event is triggered as in any input other field.
See example of use in demo-in-form.html


aaronroberson:
@oslo10e Can you achieve this functionality without introducing a new attribute?

wesleycho:
This could also probably use a unit test.

This also needs to be rebased.

PR #1293, opened 2015-11-06
Including conditional to don't break remove choice when there is no i…

vipper:
…tems for the index receive


wesleycho:
Closing, as this change was made in `dist` instead of the appropriate `src` files.

Feel free to open a new PR against `master` with the appropriate modifications.

PR #1301, opened 2015-11-10
fix(uiSelectSort): fix dependency not found error

homerjam:
Also a bit of tidying


homerjam:
Yay! Thanks @aaronroberson

PR #1310, opened 2015-11-19
Fix error in multiple choice select

lbolla:
`ngModel.$viewValue` can be null on multiple choice select and "$on('uis:select')" below can fail if so.


wesleycho:
This needs changes to `dist` undone and rebasing.

lbolla:
Should I not have changed files in `dist/`? I think they get rebuild when running `gulp test`, anyway.

lbolla:
By the way, strangely enough `gulp test` does not fail on my machine.

wesleycho:
The changes to `dist` should never be committed - part of the problem is the whole release process with the library I believe, but until that is addressed to be easier on users, this must be followed.

wesleycho:
As to the failures, please rebase off of current `master` after removing the changes to `dist`

user378230:
Isn't this superseded by #1529? (Which appears to have a better fix for the issue)

wesleycho:
Could be - I just have been doing basic triaging yesterday and today, so you're probably in a better position to judge than me.

user378230:
No problem. Any help is good help! :smile: 

I'm going to close as looks to be the same as #1529

PR #1323, opened 2015-11-25
Feature - header & footer

plong0:
allows ui-select-header and ui-select-footer directives to be embedded in the select dropdown.

Example Structure with header and footer:

```
<ui-select>
    <ui-select-match></ui-select-match>
    <ui-select-header>Top of the list!</ui-select-header>
    <ui-select-choices><ui-select-choices>
    <ui-select-footer>Bottom of the list.</ui-select-footer>
</ui-select>
```

### DIFF
--- a/docs/examples/demo-header-and-footer.html
+++ b/docs/examples/demo-header-and-footer.html
@@ -1,0 +1,17 @@
+<h3>Basic Header and Footer</h3>
+<p>Selected: {{ctrl.address.selected.formatted_address}}</p>
+<ui-select ng-model="ctrl.address.selected"
+           theme="bootstrap"
+           ng-disabled="ctrl.disabled"
+           reset-search-input="false"
+           style="width: 300px;"
+           title="Choose an address">
+  <ui-select-match placeholder="Enter an address...">{{$select.selected.formatted_address}}</ui-select-match>
+  <ui-select-header><h4>Simple Header</h4></ui-select-header>
+  <ui-select-choices repeat="address in ctrl.addresses track by $index"
+           refresh="refreshAddresses($select.search)"
+           refresh-delay="0">
+    <div ng-bind-html="address.formatted_address | highlight: $select.search"></div>
+  </ui-select-choices>
+  <ui-select-footer><strong>Simple Footer</strong></ui-select-footer>
+</ui-select>
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -227,6 +227,33 @@ uis.directive('uiSelect',
           if (transcludedNoChoice.length == 1) {
             element.querySelectorAll('.ui-select-no-choice').replaceWith(transcludedNoChoice);
           }
+
+          var transcludedHeader = transcluded.querySelectorAll('.ui-select-header');
+          var transcludedFooter = transcluded.querySelectorAll('.ui-select-footer');
+          if((transcludedHeader && transcludedHeader.length) || (transcludedFooter && transcludedFooter.length)){
+            $timeout(function(){
+              var selectChoices = element.querySelectorAll('.ui-select-choices');
+              
+              if(transcludedHeader && transcludedHeader.length){
+                transcludedHeader.removeAttr('ui-select-header'); // To avoid loop in case directive as attr
+                transcludedHeader.removeAttr('ng-transclude'); // Content has already been transcluded
+                transcludedHeader.removeAttr('data-ui-select-header'); // Properly handle HTML5 data-attributes
+                transcludedHeader.removeAttr('data-ng-transclude');
+                selectChoices.prepend(transcludedHeader);
+              }
+              
+              if(transcludedFooter && transcludedFooter.length){
+                transcludedFooter.removeAttr('ui-select-footer'); // To avoid loop in case directive as attr
+                transcludedFooter.removeAttr('ng-transclude'); // Content has already been transcluded
+                transcludedFooter.removeAttr('data-ui-select-footer'); // Properly handle HTML5 data-attributes
+                transcludedFooter.removeAttr('data-ng-transclude');
+                selectChoices.append(transcludedFooter);
+              }
+              
+              // re-compile selectChoices in case header or footer requires one of ui-select-choice's directives
+              $compile(selectChoices)(scope);
+            });
+          }
         });
 
         // Support for appending the select field to the body when its open
--- a/src/uiSelectFooterDirective.js
+++ b/src/uiSelectFooterDirective.js
@@ -1,0 +1,8 @@
+uis.directive('uiSelectFooter', function(){
+  return {
+    template: '<li class="ui-select-footer" ng-transclude></li>',
+    restrict: 'EA',
+    transclude: true,
+    replace: true
+  };
+});
--- a/src/uiSelectHeaderDirective.js
+++ b/src/uiSelectHeaderDirective.js
@@ -1,0 +1,8 @@
+uis.directive('uiSelectHeader', function(){
+  return {
+    template: '<li class="ui-select-header" ng-transclude></li>',
+    restrict: 'EA',
+    transclude: true,
+    replace: true
+  };
+});


plong0:
Not sure what's up with travis tests failing, all karma tests pass when I run local gulp.  Looks like most recent PR's are failing their travis tests too.

plong0:
*Bonus side effect* - using [ui-scrollpoint](https://github.com/angular-ui/ui-scrollpoint) inside a ui-select-footer, I was easily able to add infinite scrolling to my ui-select choices.

joshrickert:
The infinite scrolling benefit is pretty cool.

Since you're prepending/appending the header and footer to the `ul` element, does that mean the header and footer scroll into/out of view? I think the folks over in #496 were hoping for a static overlay above and below the scrolling area.

plong0:
Oh yes, you are right.  They do scroll into and out of view.  I think you could might be able to have them stay in place with css?  I will need to test that.  If not, I will think about adding something like a fixed="true" attribute.

joshrickert:
I suspected a CSS solution might be applicable too. A [quick search](http://stackoverflow.com/questions/11261270/css-position-element-fixed-inside-scrolling-container) around Stack Overflow indicates it might be a bit tricky, though.

That said, I'm not sure there is much that can be done without seriously reorganizing the templates since the top-level element of `choices.tpl.html` is an `ul` element. I needed a header for my project, so I did an add-on directive to implement the header functionality, and the approach was much the same as you took.

plong0:
Ok, well I just read this article: http://webdesign.tutsplus.com/tutorials/sticky-positioning-with-nothing-but-css--cms-24042

Looks like it *should* be as easy as `postion: sticky;`  but as the article notes, it is not supported in all browsers yet :( 
I tested `position: fixed;` in chrome, it worked for the header.  I am not going to call it a solution though, because position: fixed; is always supposed to be to the window, not the parent.

I tried using the [ngSticky](https://github.com/d-oliveros/ngSticky) module with no success.

Going to branch my fork and see about adding the fixed property to my uiSelectHeader/uiSelectFooter directives.

joshrickert:
Would it make sense to transclude the directives a little higher up the DOM tree?

joshrickert:
Also, I was trying out your code and thought it looked a bit better if I included a bootstrap list separator as well:
```
transcludedChoices.prepend(angular.element('<li>').addClass('divider').prop('role', 'presentation'));
transcludedChoices.prepend(transcludedHeader);
```

plong0:
Hmm not sure if it would work with the directives higher in the DOM tree.  If I recall, it is the `<ul>` element that gets shown/hidden/scrolled.

joshrickert:
Yeah, there would have to be some new CSS and custom logic to show/hide those elements, or maybe just a new container around the `ul` element that can be toggled instead.

For what it's worth, I have your current code monkey-patched in and it's working well.

Is fixing the travis build just a matter of merging upstream changes?

plong0:
ya just merged upstream and travis build passes now.

plong0:
I've got a gulped branch at https://github.com/TechNaturally/ui-select/tree/feat-header-footer-bower

You can use it in your bower.json dependencies with:
`"angular-ui-select": "TechNaturally/ui-select#feat-header-footer-bower"`

wesleycho:
The history on this needs fixing - there should not be any merge commit.

plong0:
Fixed history

joshrickert:
It would probably also be good to add documentation and an example so that users know the feature exists once it is available.

plong0:
Just added a [wiki page](https://github.com/angular-ui/ui-select/wiki/ui-select-header-and-footer).  Working on an infinite-scroll example using `ui-select-footer` and [`ui-scrollpoint`](https://github.com/angular-ui/ui-scrollpoint)

user378230:
You'll need to rebase especially following changes to examples in #1550 (you can just move your `<script>` to the body and update `demo.js`

Once that's done I'll merge.

plong0:
@user378230 thanks...  rebased, and renamed/updated the demo to include a basic example.

I am not sure that I could use a `<script>` tag though, because I need to add `ui.scrollpoint` as a dependency of `demo` module and add the infinity object onto `DemoCtrl's` ~~`scope`~~ `vm`.  If you could please advise a method to do this from a `<script>`, I would be happy to.

user378230:
@plong0 Can we drop the infinite scrolling portion of the example for now? Instead how about you create it as a plunkr (You can use the examples at http://angular-ui.github.io/ui-select/#examples as a start point) I suggest you could leave a link to the plunkr in your `demo-...html file` and also on the wiki page.  

This way we can get it merged and released sooner. We can look at modifying the examples set up at a later date. 

Thanks a lot for your patience and hard work on this :+1:

plong0:
Hm, can't seem to get it working [on plunker](http://plnkr.co/edit/v7Z6ap4a6mLsPYlwR4fC)

Works in production, but on plunker it doesn't seem to add the `<li>`'s for header and footer.  Can't figure it out right now.

plong0:
Hm, looks like things changed quite a bit since I originally opened this PR.  Can't see why with the newer version the `<li>`'s I add to `transcludedChoices` aren't coming in.  [plunker](http://plnkr.co/edit/v7Z6ap4a6mLsPYlwR4fC)

user378230:
It looks like angular v1.2.18 -> v 1.4.x is fine but angular v1.5.0+ is broken. There was probably a breaking change in v1.5.0 that is causing this.

mityalebedev:
@plong0 I think [vm.infinity](https://github.com/angular-ui/ui-select/pull/1323/commits/58d49b7a12ba65ea29faa3631d7ad8498c61d23e#diff-958a7d58d3efd4144d0ab0c322bed464R215) should be a directive in case of multiple ui-select use. Then we can pass data like `$scope.countries` as  isolated scope, and be able to change settings (count, limit etc) inside template. WDYT?

plong0:
@mityalebedev yes, that is a great idea to set infinity as a directive (honestly, I whipped that demo up pretty quick, but definitely in production code an infinite-scroll directive is where it's at).

I hit a block when I started setting up the plunker and the `ui-select-header` and `ui-select-footer` directives failed to work...  @user378230 suggested it was a BC in angular 1.5.0, so I will need to dig into that a little bit to try to sort out.  Doesn't seem much sense to merge this PR before it's verified working with 1.5.0.

select.js in the [plunker](http://plnkr.co/edit/2TkHMx) has the latest code of this fork compiled and pasted in if anyone is able to dabble and help figure out why it's broken.  Looks like in the newer version of ui-select, ui-select-choice transclusion has gotten a tad more complicated.

plong0:
I just poked around at it a bit more, but can't see why it works in angular 1.4 but not 1.5

The trouble appears to be that any children appended to `transcludedChoices` don't stick around.  Other manipulations like `transcludedChoices.addClass('test-class');` seem to work though.

I don't have a whole lot of time to debug the difference between 1.4 and 1.5, so hoping someone in the community can help with that.

GDuchemin:
up ?

user378230:
@GDuchemin please post meaningful responses to issues, thanks! 

Also not that this PR doesn't work with angular 1.5.x as per previous replies, can't be merged until it does.

GDuchemin:
I just try reselect the target to prepend and append `transcludeHeader` and `transcludeFooter` with `element.querySelectorAll('.ui-select-choices')`, wrapped all in `$timeout` and it's seems working.

user378230:
@GDuchemin maybe post your code?

GDuchemin:
`uiSelectDirective.js` after line `230`:
`var transcludedHeader = transcluded.querySelectorAll('.ui-select-header');
          if(transcludedHeader && transcludedHeader.length){
            transcludedHeader.removeAttr('ui-select-header'); //To avoid loop in case directive as attr
            transcludedHeader.removeAttr('data-ui-select-header'); // Properly handle HTML5 data-attributes
            $timeout(function(){
              element.querySelectorAll('.ui-select-choices').prepend(transcludedHeader);
            });
          }

          var transcludedFooter = transcluded.querySelectorAll('.ui-select-footer');
          if(transcludedFooter && transcludedFooter.length){
            transcludedFooter.removeAttr('ui-select-footer'); //To avoid loop in case directive as attr
            transcludedFooter.removeAttr('data-ui-select-footer'); // Properly handle HTML5 data-attributes
            $timeout(function() {
              element.querySelectorAll('.ui-select-choices').append(transcludedFooter);
            });
          }`

I just defined again the selector before prepend and append code.

GDuchemin:
`var transcludedHeader = transcluded.querySelectorAll('.ui-select-header');
          if(transcludedHeader && transcludedHeader.length){
            transcludedHeader.removeAttr('ui-select-header'); //To avoid loop in case directive as attr
            transcludedHeader.removeAttr('data-ui-select-header'); // Properly handle HTML5 data-attributes
            $timeout(function(){
              element.querySelectorAll('.ui-select-choices').prepend(transcludedHeader);
            });
          }`

plong0:
Thanks so much for the input @GDuchemin! your solution for appending after $timeout works in angular 1.4 + 1.5.  You brought this back to life.

I've rebased my branch and done a bit more work to refactor with the `$timeout`.  I also added a `$compile` of the `ui-select-choices` after appending header and footer.

The reason for the re-compile is in case any directives in the header or footer require the `ui-select-choices` or any of its attribute directives.  For example: in the infinite-scroll demo, the `ui-scrollpoint` inside the `ui-select-footer` requires the `ui-scrollpoint-target` on the `ui-select-choices` in order to detect when the choices have been scrolled.

[Plunker demo updated](http://plnkr.co/edit/FzlXOH) with latest compiled select.js+select.css copied in.

**Everything seems to be working with angular 1.4 + 1.5!**

I think this branch might finally be ready to merge...

GDuchemin:
Great work @plong0 :)

dannydinges:
When will this be merged?

user378230:
@dannydinges if you *watch* the repo you should get a github notification when this thread is updated, you can then pop by to see if it's been merged. Hope that helps! :smiley:

kotmatpockuh:
@plong0, maybe you know, why ng-click in ui-select-header is fired twice?

margielm:
Hello,
Any idea when this PR will be merged and released?

plong0:
@user378230 what is the needs: PR test(s) label?

chansx:
@kotmatpockuh I also encountered the same problem. Did you solve the problem?

gpsinghsandhu:
@kotmatpockuh @chansx The problem seems to be [this](http://stackoverflow.com/a/32014248). Commenting out the $compile call (line 254 in src/uiSelectDirective.js) worked for me as there was no need for an extra compile in my use case.

plong0:
@kotmatpockuh @chansx @gpsinghsandhu

Based on the [answer @gpsinghsandhu linked to](http://stackoverflow.com/questions/18709297/angularjs-ng-click-fires-twice/32014248#32014248), maybe it needs to `$compile` each transcluded bit separately rather than the entire `selectChoices` -  [Something like this](http://pastebin.com/AeiQMAYk)

artgryn:
Probably merge and release? =)

shyamal890:
Any updates on this?

felixmosh:
@kotmatpockuh, i encountered the same problem, even worse, `ng-if` will generate each time you open & close the list.

The reason is that compliation: https://github.com/TechNaturally/ui-select/blob/b743fcce965ad6ed1ab4a200e961d119af774b42/src/uiSelectDirective.js#L254

felixmosh:
I`ve created a clean PR that implements header & footer without hacks.
You can create a theme for it, there is no double compilation therefore there is no double click trigger bugs.
https://github.com/angular-ui/ui-select/pull/1930

igorbar:
Could somebody say when this PR will be released?

Jefiozie:
It has been released. Closing this PR down.

georgeslegros:
Could you please validate the release number it is supposed to be included in? I checked the latest and 

1. select.js still says version 0.19.7
2. select.js does not contain any reference to "footer"
3. the directive uiSelectFooterDirective.js is not in the src folder anymore

Was the feature removed or something?

Thanks

felixmosh:
It was removed, there is another PR that waits for a long time to be merged.
https://github.com/angular-ui/ui-select/pull/1982

PR #1324, opened 2015-11-26
Improve the paste handler to work even without tagging function.

jojkaart:
This patch will improve copy&paste support. With this, it's no longer necessary to specify a scope function to be able to copy&paste into the ui-select field. copy&paste will now only require tagging.

Additionally, this improves tokenization handling. pasted data is now properly cut into pieces even when separator is ENTER, TAB or SPACE. Only splits the copy&paste with the first separator defined, though. The default is comma;

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -307,7 +307,7 @@ uis.directive('uiSelect',
 
               if ($select.dropdownPosition === 'up'){
                   //Go UP
-                  setDropdownPosUp(offset, offsetDropdown);
+//                  setDropdownPosUp(offset, offsetDropdown);
 
               }else{ //AUTO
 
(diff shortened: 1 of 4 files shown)


jojkaart:
commit a1cd1c5 comments out the line of code that causes travis to fail in something entirely unrelated to my changes. It should probably be fixed some other way.

jojkaart:
commit 5166d22 removes the filters I'd added to leave empty or whitespace-only tags out when copypasting. I've got no idea why it makes sense to be able to insert empty strings but since it was in the test suite, it's perhaps important. Now the code finally passes the tests.

src/uiSelectDirective.js
```diff
@@ -307,7 +307,7 @@ uis.directive('uiSelect',
 
               if ($select.dropdownPosition === 'up'){
                   //Go UP
-                  setDropdownPosUp(offset, offsetDropdown);
+//                  setDropdownPosUp(offset, offsetDropdown);
```
jojkaart: This line was causing tests to fail. I first tried to define offset and offsetDropdown right above this, but that then failed for the opposite reason. I decided I don't want to bang my head on that wall and just commented it out.

aaronroberson:
This has been merged and has been published in v0.14.5

PR #1329, opened 2015-11-27
Update select.tpl.html

orion514:
Edge need off to operate without auto complete. Everything works on other browsers

autocomplete="off".


wesleycho:
Closing, as there is another PR from earlier that does a more thorough job at making the necessary changes here.

PR #1337, opened 2015-12-02
Do not overflow matched item out of select block

just-boris:
If selected value is so long, the select looks not nice. See the plunk: http://plnkr.co/edit/QtrdYP?p=preview

This css fixes it.


just-boris:
FYI. Build has failed as well as other pull-requests because of some code, which already is in master.

wesleycho:
This looks pretty straightforward.

user378230:
This should probably be applied to the other templates too I think. Will merge for now.

user378230:
Had to revert this in d7d26ca as it broke with the `allow-clear` button (pushed it out of the select)

PR #1350, opened 2015-12-09
Use ng-attr for placeholder

justinwinslow:
The placeholder attr throws an error in IE10 and 11 (and maybe earlier versions?) and an easy solution is to use the ng-attr prefix. More info here: https://github.com/angular/angular.js/issues/5025

I visually validated that it works fine and ran gulp and all tests passed. I didn't include the updated dist files here because I didn't know what your process was but can happily do whatever you need.

### DIFF
--- a/src/bootstrap/select-multiple.tpl.html
+++ b/src/bootstrap/select-multiple.tpl.html
@@ -2,12 +2,12 @@
   <div>
     <div class="ui-select-match"></div>
     <input type="text"
-           autocomplete="false" 
-           autocorrect="off" 
-           autocapitalize="off" 
-           spellcheck="false" 
+           autocomplete="false"
+           autocorrect="off"
+           autocapitalize="off"
+           spellcheck="false"
            class="ui-select-search input-xs"
-           placeholder="{{$selectMultiple.getPlaceholder()}}"
+           ng-attr-placeholder="{{$selectMultiple.getPlaceholder()}}"
            ng-disabled="$select.disabled"
            ng-hide="$select.disabled"
            ng-click="$select.activate()"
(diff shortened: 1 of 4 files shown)


justinwinslow:
Added dist files to my repo so I can use it while I wait for this to be reviewed.

src/bootstrap/select-multiple.tpl.html
```diff
@@ -2,12 +2,12 @@
...
-           autocorrect="off" 
-           autocapitalize="off" 
-           spellcheck="false" 
+           autocomplete="false"
+           autocorrect="off"
+           autocapitalize="off"
+           spellcheck="false"
            class="ui-select-search input-xs"
-           placeholder="{{$selectMultiple.getPlaceholder()}}"
+           ng-attr-placeholder="{{$selectMultiple.getPlaceholder()}}"
```
wesleycho: I don't think the switch to ng-attr here is necessary

src/bootstrap/select.tpl.html
```diff
@@ -6,7 +6,7 @@
          aria-owns="ui-select-choices-{{ $select.generatedId }}"
          aria-activedescendant="ui-select-choices-row-{{ $select.generatedId }}-{{ $select.activeIndex }}"
          class="form-control ui-select-search"
-         placeholder="{{$select.placeholder}}"
+         ng-attr-placeholder="{{$select.placeholder}}"
```
wesleycho: Ditto

wesleycho:
Interesting - I stand corrected and deleted my comments.

This just needs the dist changes removed and rebasing.

Issue #981, opened 2015-06-02
Adding custom validation directive removes all selected values


martinoss:
I added a custom validation directive to ui-select that validates all selected items to make sure each item has a valid email property.

But after validation returns falsy, all selected items gets lost:
http://plnkr.co/edit/gW52dvUjMx4ydHEaHa4a?p=preview

awerlang:
Add `ng-model-options="{allowInvalid: true}"` next to ng-model.

Reference: https://docs.angularjs.org/api/ng/directive/ngModelOptions

meyerds:
I'm also in need of a fix for this issue - and allowing invalid won't work in my situation. I'm researching the right way to handle this, if I find the answer I'll at minimum post here, or try to submit a pull request.

meyerds:
I've submitted a pull request for this issue: https://github.com/angular-ui/ui-select/pull/1359

PR #1359, opened 2015-12-16
Bug Fix: Selected items should not be removed when they are found to be invalid

meyerds:
### ui-select configuration:
- tagging

### problem code flow:
1. User enters tag text and accepts the new tag(s) (with tab or the like)
2. A validator determines that the tag is invalid, and sets the model value of the ui-select instance to undefined. At this point the view value still shows the invalid value.
3. The items list is refreshed during the tagging process, and is triggered after validators run. The items list on the uiSelectCtrl is updated.
4.  The item watcher sees the change and causes the view value to be updated from the model value, in order to reflect the data in the fresh items.
5. Since the model value is undefined until the invalid condition is resolved, the formatters run and create an empty view value, erasing the tagged item(s) - the user does not have an opportunity to resolve the invalid condition, their data is just erased.

### solution:
Do not update the view value on items refresh, when the model value is undefined

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -200,7 +200,11 @@ uis.controller('uiSelectCtrl',
           //Remove already selected items (ex: while searching)
           //TODO Should add a test
           ctrl.refreshItems(items);
-          ctrl.ngModel.$modelValue = null; //Force scope model value and ngModel value to be out of sync to re-run formatters
+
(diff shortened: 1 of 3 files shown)


JakePartusch:
:+1:

meyerds:
This pull request should resolve this issue: https://github.com/angular-ui/ui-select/issues/981

wesleycho:
This just needs squashing, but LGTM

PR #1365, opened 2015-12-22
Fix angular.element.cache memory leak

Frozenball:
After doing memory leak hunting in my application I found that ui-select is leaking memory. I'm not sure why this is exactly needed as I would have expected $compile to handle it (I tried looking at uiSelectDirective.js with no results)

### DIFF
--- a/src/uiSelectChoicesDirective.js
+++ b/src/uiSelectChoicesDirective.js
@@ -61,6 +61,10 @@ uis.directive('uiSelectChoices',
 
         $compile(element, transcludeFn)(scope); //Passing current transcludeFn to be able to append elements correctly from uisTranscludeAppend
 
+        scope.$on('$destroy', function() {
+          choices.remove();
+        });
+
         scope.$watch('$select.search', function(newValue) {
           if(newValue && !$select.open && $select.multiple) $select.activate(false, true);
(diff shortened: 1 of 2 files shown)


wesleycho:
This is a good catch - LGTM

Frozenball:
Thanks! :clap:

raman-savitski:
Cause an issue if ui-select is being toggled on page by ng-if. If expression is initialized as "true" and immediately toggled to "false"  - choices element will be removed forever.
`<div ng-if="expression">   <ui-select>.....</ui-select> </div>`

In this case "compile" called only once, but "link" called each time ng-if is triggered. And this is the real reason of memory leak I suppose, because "link" is binded to "compile" function closure.

I propose to isolate "link" fn from "compile" scope:
`return (function(groupByExp, groupFilterExp){ return function link(scope, element, attrs, $select){...} })(groupByExp, groupFilterExp);`
replace `tElement` usage with `element`
and remove `scope.$on('$destroy'` at all, as it is not required anymore.

user378230:
@raman-savitski If you want to report a bug you should open a new issue here on github and follow the issue template. (You can obviously reference this PR in your issue).

You or someone else can then submit a PR to resolve it :+1:

PR #1370, opened 2015-12-29
Feat publisher 1.2

m3hul:


wesleycho:
It looks like the library has switched to gulp since then - feel free to reimplement this feature in gulp on top of current `master` and reopen a PR.

Issue #1298, opened 2015-11-09
Dropdown flickering cause opacity set to 0 when filtering


onemenny:
Please see: http://plnkr.co/edit/erN0pTgEVxBrfJF4p9U7?p=preview
Use the select2 theme and type 'a' and then 'd' you will see that on each letter you type the drop filckers.

It appears that this is cause by line 1091 
```javascript
// Hide the dropdown so there is no flicker until $timeout is done executing.
            dropdown[0].style.opacity = 0;

            // Delay positioning the dropdown until all choices have been added so its height is correct.
            $timeout(function(){
.
.
.
 // Display the dropdown once it has been positioned.
              dropdown[0].style.opacity = 1;
            });
```

opacity should not change.

chnldw:
Same here with angular 1.4.7 and ui-select 0.13.2

cfey:
this fixed it for me as well

shyamal890:
Linking similar issue https://github.com/angular-ui/ui-select/issues/1314

aaronroberson:
@shyamal890 Can you direct me to a PR that I can review and accept?

onemenny:
@aaronroberson look at my pull requests https://github.com/angular-ui/ui-select/pull/1371 I already enhanced it with more bug fixes

aaronroberson:
Any chance you can remove the updates to the dist files and optionally
squash the commits?

onemenny:
Sure, I'll do it later on today.

onemenny:
@aaronroberson ok, I hope I got it right. everything is squashed and dist is reset. 
There are many changes to this pull request and some new test. here are the highlights:

		a. Do not set opacity when typing (causes flickering)
		b. Appen-to-body=true cause loos of focus on tag add/delete, backspace…
		c. Using tags (dashboard tags): type the letter 'a' which should appear in one of the ddl items, then delete the letter (empty search), causes the <'a' new item> to remain in the ddl items
		d. when using closeOnSelect with tagging - hitting the enter key sets the new tag but keeps the search input value
		e. Find dupes: Tags duplication helper method was added (tags may be objects and need comparer function)
		f. Fix ensureHighlight to work only when ctrl.open

yogeshgadge:
I am going with the following hack for now.
```

            // Hide the dropdown so there is no flicker until $timeout is done executing.
            if (dropdown[0].clientHeight == 0) {
              dropdown[0].style.opacity = 0;
            }
```

[...]

PR #1371, opened 2015-12-29
Adding fixes mainly to tagging functionality

onemenny:
added isNewTagDuplication optional user method to check if new tag (when using tagging where each tag is an object) already exists in the list

tagging: making sure no empty elements gets in on paste operation

fix bug: when using closeOnSelect with tagging - hitting the enter key sets the new tag but keeps the search input value

fixing bug - when typing a letter that exists in one of the items in the ddl and then deleting it causes the (tag new item) to persists

fix lost focus (on backspace or adding new tag) when append-to-body is used

remove opacity changes - that causes flickering (https://github.com/angular-ui/ui-select/issues/1298) (https://github.com/angular-ui/ui-select/issues/1298)


wesleycho:
This needs work - there should be no changes to `dist`, and there should be no merge commit. The history needs work.

PR #1375, opened 2016-01-05
Fix RTL support when using tagging - calculate right input width

dani3l:
Hi. 
Just noticed that there's a "bug" when using the tagging option within an RTL document.
The width of the input wasn't calculated like it should. 
This pull request fixes it.


wesleycho:
Can you squash your commits and rebase it on current `master`?

In addition, it probably is worth having tests here for this functionality.

drewfreyling:
@dani3l what is the progress on this one?

Issue #980, opened 2015-06-02
Clear button shown on disabled select


michalmatovcik:
I have ui-select with ng-disabled =false.
Select itself behave as expected - so i cannot select another value.

But clear  button (allow-clear=true) is still shown and working!

Suggestion:
I have fixed it in my own app - by changing setAllowClear function to set $select.allowClear to false if $select.disabled is true.

Next step is to call setAllowClear function from inside of $select.disabled watcher.

Please evaluate this bug and fix it.
thanks

blowsie:
For me the clear button does not show when the input is disabled

michalmatovcik:
which version are you using? i have found it in latest release

2015-06-08 13:01 GMT+02:00 Sam Blowes <notifications@github.com>:

wesleycho:
Closing due to lack of reproduction in Plunker - feel free to open a new issue with a Plunker if it is still a problem.

PR #1388, opened 2016-01-15
hide clear button if disabled

Raytray:
This addresses this issue: https://github.com/angular-ui/ui-select/issues/980 
> Clear button shown on disabled select

This will hide the clear button if $select.disabled is true.


user378230:
@Raytray Can you update this to just be a `!$select.disabled`? It's a bit verbose at the moment, unless it was intentional?

Raytray:
It was intentional, if I remember correctly, it is because of other
falsy/truthy values.

-Raymond Tang

user378230:
Ok thanks :+1:

PR #1401, opened 2016-01-25
Why not check for nullability here?.

Rogerkael:
I am having a exception here because the regular expression is not matching anything.


wesleycho:
Closing as already addressed in `master`.

PR #1407, opened 2016-01-26
use findIndex instead of indexOf when looking up activeIndex

plong0:
When dealing with a list of object values, indexOf isn't always sufficient for accurately finding the activeIndex.  This patch uses findIndex instead of indexOf to look up the activeIndex.

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -66,7 +66,9 @@ uis.controller('uiSelectCtrl',
       ctrl.search = EMPTY_SEARCH;
       //reset activeIndex
       if (ctrl.selected && ctrl.items.length && !ctrl.multiple) {
-        ctrl.activeIndex = ctrl.items.indexOf(ctrl.selected);
+        ctrl.activeIndex = ctrl.items.findIndex(function(item){
+          return angular.equals(this, item);
+        }, ctrl.selected);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


user378230:
findIndex() isn't supported by IE

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex
> Feature | Chrome |	Firefox (Gecko) | Internet Explorer | Opera | Safari
> ----------|----------|----------|----------|----------|-------
> Basic support | 45.0 | 25.0 (25.0) | Not supported | Not supported |7.1

aaronroberson:
@plong0 Please use the MDN [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Polyfill) polyfill. If possible, please also squash your commits (i.e. `git rebase -i HEAD~4`).

plong0:
@aaronroberson RE: polyfill ... is it preferable to copy the polyfill code right into the uiSelectController.js or better to add a bower dependency on https://github.com/paulmillr/Array.prototype.findIndex ?

no problem on squashing once I get the polyfill in...

aaronroberson:
To prevent the introduction of any dependencies to the project I would recommend adding the code to uiSelectController.js at this time.

plong0:
polyfill installed and commits squashed

PR #1416, opened 2016-01-31
Feature suggestion

cedriclombardot:
Hi,

This is an implementation suggestion for a footer out of objects list, eg to propose create new element :

``````
 html
<ui-select ng-model="person.selected" theme="select2" ng-disabled="disabled" style="min-width: 300px;" title="Choose a person">
    <ui-select-match placeholder="Select a person in the list or search his name/age...">{{$select.selected.name}}</ui-select-match>
    <ui-select-choices repeat="person in people | propsFilter: {name: $select.search, age: $select.search}">
      <div ng-bind-html="person.name | highlight: $select.search"></div>
      <small>
        email: {{person.email}}
        age: <span ng-bind-html="''+person.age | highlight: $select.search"></span>
      </small>
    </ui-select-choices>


    <ui-select-footer>
      <a ng-click="footerClick($event)">Add new item</a>
    </ui-select-footer>
  </ui-select>
`````


nsang92:
+1

Narretz:
My 2 cents: I don't think we should throw even more features at this directive. Atm there are tons of bugs with existing features. Piling up on it will not make it better. There are also no tests for this change.

aaronroberson:
As @Narretz has said, this PR needs tests.

aaronroberson:
To follow up, I want to clarify that we are not keen on adding a plethora of directive attributes to the project. I would love to see a PR that refactors the the directive away from having a million attributes and instead accepting a config object. But regarding this footer directive, all it needs is unit tests to be accepted.

wesleycho:
The changes to the gulpfile and the `dist` directory are also no good.

user378230:
This will be superseded by #1323 when its working and merged so going to close this.

PR #1428, opened 2016-02-10
Don't alter Array.prototype.findIndex

graingert:


graingert:
@plong0 ^

graingert:
@aaronroberson it looks like this only fails because of existing test failures.

plong0:
I am confused about what your concern is as there is an if check around the polyfill.

Please notice that this is the polyfill copied from the MDN documentation on the [Array.findIndex page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Polyfill).  You can review the comments on PR #1407 which added the polyfill to see that it was recommended to include the polyfill vs writing a new function.

graingert:
@plong0 this polyfill actually broke our application and I had to downgrade to 0.14.1

graingert:
While there is a check, it would be much better to use a Shim over a Polyfill in library code

plong0:
Hm, that's unfortunate, but it does seem a bit strange that a polyfill would break your app.

As far as using a shim vs a polyfill in a library, from what I understand of them they are meant to serve different purposes.  Since Array.findIndex is becoming a standard feature of newer JS implementations, it seems more appropriate to use the polyfill...  If a polyfill really broke your application, wouldn't that mean your application is already broken in some browsers?

graingert:
@plong0 basically we didn't expect this property to exist in IE10, but then it suddenly did.  To be fair it broke legacy code in the app, but this could break other things.

Ultimately it comes down to the fact that "polite" libraries shim, and in fact in browserify/webpack land it's best to require a compatibility shim (that dispatches accordingly) even in application code.

aaronroberson:
Thomas, can you post the exception or stack trace for the error I'm
extremely curious to know how the polyfill could be causing an error

On Thu, Feb 11, 2016, 8:43 AM Thomas Grainger <notifications@github.com>
wrote:

> @plong0 <https://github.com/plong0> basically we didn't expect this
> property to exist in IE10, but then it suddenly did. To be fair it broke
> legacy code in the app, but this could break other things.
>
> Ultimately it comes down to the fact that "polite" libraries shim, and in
> fact in browserify/webpack land it's best to require a compatibility shim
> (that dispatches accordingly) even in application code.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/angular-ui/ui-select/pull/1428#issuecomment-182950489>
> .
>
-- 

*Aaron Roberson*
JavaScript Software Engineer

Mobile: *(559) 492-7467*
E-mail: *aaronaroberson@gmail.com <Aaron%20Roberson>*
Skype: *aaronroberson <file://localhost/skype/aaronroberson%3Fcall>*
LinkedIn: *http://linkedin.com/in/aaronrobes
<http://linkedin.com/in/aaronrobes>*
GitHub: *http://github.com/aaronroberson <http://github.com/aaronroberson>*
------------------------------

Interested in scheduling a meeting? View my calendar and book a time to
talk using *https://aaronroberson.youcanbook.me
<https://aaronroberson.youcanbook.me>**.*

I ♥ JavaScript

graingert:
@aaronroberson it's how adding this prototype effects for loops over arrays. I know you're not supposed to use them, but it's used everywhere in a legacy codebase of ours.

aaronroberson:
Are you checking for `hasOwnProperty` in your  `for` loops? If not,  as
painful as it may be, you need to update your code base.

On Thu, Feb 11, 2016, 8:47 AM Thomas Grainger <notifications@github.com>
wrote:

> @aaronroberson <https://github.com/aaronroberson> it's how adding this
> prototype effects for loops over arrays. I know you're not supposed to use
> them, but it's used everywhere in a legacy codebase of ours.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/angular-ui/ui-select/pull/1428#issuecomment-182952683>
> .
>
-- 

*Aaron Roberson*
JavaScript Software Engineer

Mobile: *(559) 492-7467*
E-mail: *aaronaroberson@gmail.com <Aaron%20Roberson>*
Skype: *aaronroberson <file://localhost/skype/aaronroberson%3Fcall>*
LinkedIn: *http://linkedin.com/in/aaronrobes
<http://linkedin.com/in/aaronrobes>*
GitHub: *http://github.com/aaronroberson <http://github.com/aaronroberson>*
------------------------------

Interested in scheduling a meeting? View my calendar and book a time to
talk using *https://aaronroberson.youcanbook.me
<https://aaronroberson.youcanbook.me>**.*

I ♥ JavaScript

graingert:
I agree we should update this, but some code depends on not checking hasOwnProperty and some does not. This is a legacy codebase where webpack isolation is helping us to re-factor.

This doesn't mean it's necessarily a good idea to assign to a global variable or prototype in library code.

danpalmer:
I think the "Be conservative in what you send, be liberal in what you accept" adage applies quite well here. Code using these arrays should obviously check that it's only getting the properties it needs, but also code modifying the array should probably leave it as-is rather than modifying it.

Another thing to consider is that consistency with what consumers expect is arguably more valuable than doing things the 'modern' way, especially given that 'modern' is incredibly subjective.

adarshbhat:
I was going to say, "who does a `for..in` over an array!"... until I found an instance in our code base too that failed over exactly the same thing. Luckily unit tests caught it during the upgrade, and we fixed it.

graingert:
@aaronroberson we also want to be able to use babel-polyfill and we get a different implimentation of Array.prototype.findIndex depending on which order we load these modules.

aaronroberson:
Why not use babel during build instead of at runtime in the browser? Seems
you guys are doing a few "non-standard" practices that are causing more
harm then good?

On Tue, Feb 16, 2016, 7:44 AM Thomas Grainger <notifications@github.com>
wrote:

> @aaronroberson <https://github.com/aaronroberson> we also want to be able
> to use babel-polyfill and we get a different implimentation of
> Array.prototype.findIndex depending on which order we load these modules.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/angular-ui/ui-select/pull/1428#issuecomment-184738366>
> .
>
-- 

Aaron Roberson

JavaScript Software Engineer

Mobile: (559) 492-7467

E-mail: aaronaroberson@gmail.com

Skype: aaronroberson

LinkedIn: http://linkedin.com/in/aaronrobes

GitHub: http://github.com/aaronroberson

------------------------------

Interested in scheduling a meeting? View my calendar and book a time to
talk using https://aaronroberson.youcanbook.me.

I ♥ JavaScript

graingert:
@aaronroberson babel runs at build babel-polyfill runs in the browser. It just loads things like core-js

graingert:
@aaronroberson I don't see the advantage of including a polyfill in this library. I don't see the disadvantage of allowing projects to pollyfill, or working around the lack of a polyfill with a plain function.

aaronroberson:
@graingert Agreed. Fixed in v0.14.4.

graingert:
@aaronroberson did you get hacked?

aaronroberson:
It's more embarrassing than that. I did a bulk invite using Gmail and apparently ever GitHub issue I've replied to from email is not getting spammed.

PR #1447, opened 2016-02-17
fix: Allow setting a ngClass on <ui-select> element

### LINKED ISSUE #277: ng-class on ui-select
Cannot put ng-class on element ui-select

Magador:
Since a ngClass is already present in the uiSelect theme templates, we could not add any ngClass on <ui-select> without getting an JS error
Currently, adding a ngClass on <ui-select> result in:
`ng-class="{class: expression} {open: $select.open}"`
With this fix, doing the same thing result in:
`ng-class="{class: expression, open: $select.open}"`

Closes #277

Maybe I should write a test case ?

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -17,6 +17,14 @@ uis.directive('uiSelect',
     controllerAs: '$select',
     compile: function(tElement, tAttrs) {
 
+      // Allow setting ngClass on uiSelect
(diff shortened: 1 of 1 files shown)


Narretz:
Even such a small change should have a test, imo.

aaronroberson:
Agreed. @Magador can you add the test in another PR?

Magador:
Sure !

PR #1448, opened 2016-02-19
test(uiSelectDirective): Add test when merging ngClass attributes on uiSelect

Magador:
References PR #1447
- Add missing test on uiSelect when merging ngClass attributes between DOM declaration and template
- Add one property `attrs.ngClass` in `createUiSelect(attrs)` to append a `ng-class` attribute to the `<ui-select>` element


aaronroberson:
Thanks @Magador!

PR #1464, opened 2016-02-26
Fix for filteredItems in async loading data

ggiovinazzo:
When loading data in async mode into the directive, if ngModel had already preloaded data, the filter was not applied correctly. This issue was due to behaviour of Array.indexOf prototype method. When comparing arrays of objects, the comparison is made by object reference and always fails.
Switched to angular.isEqual for in depth comparison of objects (this also aids to get rid of $ prefixed variables that angular adds to objects)


ggiovinazzo:
This issue has been resolved and merged with PR #1473

PR #1468, opened 2016-03-01
Fix: Made the ENTER key work for tagging in single (non-multiple) mode

jose-troche:
When tagging is enabled (and in non-multiple mode), the ENTER key was processed first under the _handleDropDownSelection, invalidating the selection for a new tag. 

You can see this failing in examples/demo-tagging.html, in the last two examples: "Tagging without multiple" and "Tagging without multiple, with simple strings". They both fail for new tags when user types the ENTER key to add a new tag. 

With this proposed fix that issue is resolved. The _handleDropDownSelection should only be called if no new tag is detected.


jose-troche:
I found this does not work in every case. Closing for now...

PR #1473, opened 2016-03-02
fix(uiSelectMultipleDirective): Fixes checks for existing tags during multiple tag creation and selection

ravishivt:
This is the same pull request as #647 without the dist/ files.

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -299,9 +299,19 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
               stashArr = stashArr.slice(1,stashArr.length);
             }
             newItem = $select.tagging.fct($select.search);
-            newItem.isTag = true;
-            // verify the the tag doesn't match the value of an existing item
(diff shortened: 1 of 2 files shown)


src/uiSelectMultipleDirective.js
```diff
@@ -299,9 +299,19 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
               stashArr = stashArr.slice(1,stashArr.length);
             }
             newItem = $select.tagging.fct($select.search);
-            newItem.isTag = true;
-            // verify the the tag doesn't match the value of an existing item
-            if ( stashArr.filter( function (origItem) { return angular.equals( origItem, $select.tagging.fct($select.search) ); } ).length > 0 ) {
+            // verify the new tag doesn't match the value of a possible selection choice or an already selected item.
+            if (
+              stashArr.some(function (origItem) {
+                 return angular.equals(origItem, $select.tagging.fct($select.search));
```
lukaw3d: Was there any hidden reason for choosing
```js
return angular.equals(origItem, $select.tagging.fct($select.search));
```
in this line, and not the same as below
```js
return angular.equals(origItem, newItem);
```

Issue #1462, opened 2016-02-25
Error when using ui-select-choices repeat without a filter


bogdanalexe90:
Hi,
I receive an `$parse` error when using `<ui-select-choices repeat="item in itemList track by item.id">`
I managed to track down the problem and it seems to be the regex from `uisRepeatParser`.

The problem seems to be:
* `source: $parse(!match[3] ? match[5] : match[6]),`
* `match[3]` is `undefined`
* `match[5]` is `itemList track by item.id` 
* `$parse` throws the following error `Token 'track' is an unexpected token at column 17 of the expression [itemList track by item.id] starting at [track by item.id]`.

user378230:
Fixed by #1477

Issue #1270, opened 2015-10-23
Latest build (0.13.2) not working with latest angular (1.4.7) and bootstrap (3.3.7)


Dimitar-Karazhov:
I have issues after upgrade to Angular 1.4.7, Bootstrap 3.3.5 and Select 0.13.2.
Select not working...

Error: [$parse:syntax] Syntax Error: Token 'track' is an unexpected token at column 11 of the expression [addresses track by $index] starting at [track by $index].

TypeError: Cannot read property 'source' of undefined

http://plnkr.co/edit/SQZXrRzOTZLG4YgXWOXr?p=info

dashawk:
Just remove the `track by $index` and it will solve your problem.

Dimitar-Karazhov:
The problem is not only the "track by $index", if remove other errors are generated.

codetrash:
@Dimitar-Karazhov  yaap I've got the same issue too....still dont know wht's wrong

tomasalmeida:
I am facing the same issue too.

faisalhasnain:
Putting in parens has solved, I don't know whats the mystery here:
`city in (cities) track by $index`
ui-select 0.13.2 with angular 1.4.8

user378230:
Resolved by #1477

Issue #1263, opened 2015-10-22
Track by broken


ChristianKohler:
See: https://github.com/angular-ui/ui-select/pull/1208#issuecomment-150211648

samuel-weber:
+1

chris-hanson:
While not really a fix - I've found if you add the filter before the track by it will work eg:

samuel-weber:
Hi, Thanks for the comment. We cannot do this in our solution. Do so we have to refactor the whole code...
According to the documentation it should be possible to use the standard syntax: https://github.com/angular-ui/ui-select/wiki/ui-select#examples-async

stringbeans:
+1

@chris-hanson is correct in that everything is fine if a filter is applied before the track by, but this should be fixed

MikaAK:
+1 I'm experiencing this aswell. Thanks for the workaround @chris-hanson

masterspambot:
:100: It's really important feature to fix!

sultan636:
+1 when filter is not present, but track by is, then app throw exception

meyerds:
+1

user378230:
Closing as resolved by #1477

PR #1477, opened 2016-03-06
fix(parser) allow track by without filter

### LINKED ISSUE #1233: "track by" in simple comprehension expression broke when updating from 0.11.2 to 0.13.1
The expressions
```html
<ui-select-choices
      repeat="question in questions track by question.question_id"
      group-by="'searchGroup'"
      refresh="refreshQuestions($select.search)"
      refresh-delay="250">
    <div question-preview-item="question"></div>
  </ui-select-choices>
```
used to work in 0.11.2, but after I upgraded to 0.13.1, I receive the following error:
```
Error: [$parse:syntax] Syntax Error: Token 'track' is an unexpected token at column 11 of the expression [questions track by question.question_id] starting at [track by question.question_id]
```

Changing the `repeat` value to `repeat="question in questions"` avoids the error message, but is probably taking a toll in reduced performance.

user378230:
## Allow track by without filter

Simple repeat expressions that used track by failed.
```html
<ui-select-choices repeat="question in questions track by question.question_id"...>
```
The previous regex attempted to extract the filters from the expression but greedily included the track by portion too. The filters, however, are only needed when using an object as a source as `ui-select` needs to manually apply them when converting the object to an array. 

Now the repeat parser will extract the source expression in one lump and let `$parse` deal with it if it can, only if an object source is used, will it attempt to extract the filters. 

This more closely aligns `ui-select` with `ngOption`'s parsing/regex (though it doesn't need to extract any filters). 

Fixes #1233 #1462 #1263 #1270

>_Apologies if the regex is a bit messy - it is tested though!_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+<a name="0.14.9"></a>
+## [0.14.9](https://github.com/angular-ui/ui-select/compare/v0.14.9...v0.14.9) (2016-03-06)
+
+
+
+
 <a name="0.14.8"></a>
 ## [0.14.8](https://github.com/angular-ui/ui-select/compare/v0.14.8...v0.14.8) (2016-02-18)
 
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "url": "git://github.com/angular-ui/ui-select.git"
   },
   "style": "dist/select.css",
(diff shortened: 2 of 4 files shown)


aaronroberson:
This has landed in ui-select@0.14.10. Thanks @user378230!

PR #1480, opened 2016-03-07
Fixed #850 Where Multiple Attribute Prevents Proper Required Validation

### LINKED ISSUE #850: required tag

kmccullough:
Please note that some of this code does not appear to be mine. I can only assume someone forgot to build.

### DIFF
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -128,6 +128,7 @@ describe('ui-select tests', function() {
       if (attrs.appendToBody !== undefined) { attrsHtml += ' append-to-body="' + attrs.appendToBody + '"'; }
       if (attrs.allowClear !== undefined) { matchAttrsHtml += ' allow-clear="' + attrs.allowClear + '"';}
       if (attrs.inputId !== undefined) { attrsHtml += ' input-id="' + attrs.inputId + '"'; }
+      if (attrs.ngClass !== undefined) { attrsHtml += ' ng-class="' + attrs.ngClass + '"'; }
     }
 
     return compileTemplate(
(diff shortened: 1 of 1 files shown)


user378230:
Whoops that was probably me, sorry :flushed: 

Also, wouldn't it make more sense to override `ngModel.$isEmpty` in `uiSelectMultipleDirective` rather than checking `attrs.multiple` or am I missing something?

kmccullough:
@user378230 It might make sense to do it there, and I can move it there if I need to, but I figured this was an OK place to put it, since it actually covers the $isEmpty for both multiple and non-multiple selects at runtime. This also allows the multiple attribute to be added/removed at runtime (in this context; I'm not sure if the rest of the control supports it).

aaronroberson:
@kmccullough Can you also remove the dist/ files and rebase to squash these commits?

kmccullough:
Also? In addition to what?

aaronroberson:
Sorry, I was referring to @user378230's comment and if the two have you have come to a resolution?

kmccullough:
Short of him responding, I suppose I can just move it since the default $isEmpty should cover that case. I'll try to have this done by end of day if @user378230 doesn't respond.

user378230:
>...since it actually covers the $isEmpty for both multiple and non-multiple selects at runtime
I think single selection should be covered by the default $isEmpty check in angular. (https://github.com/angular/angular.js/blob/f1aea54a9b4ceaf13676a4e7b74bf0de718cd6a4/src/ng/directive/ngModel.js#L320)
> This also allows the multiple attribute to be added/removed at runtime (in this context; I'm not sure if the rest of the control supports it).
I don't think it does

I mean its no big deal... its only one or two lines I think it just makes more sense being in the multiple directive. :smile:

kmccullough:
@user378230 @aaronroberson 
Made the requested changes and opened an actual branch https://github.com/angular-ui/ui-select/pull/1492

PR #1487, opened 2016-03-13
Configure skip focusser

excentris:
This PR provides a new feature to be able to configure skipping the focusser after selecting an item.
After much trying I have unfortunately not found a way to cover the changes with the necessary unit tests. Focusing and testing the involved elements for focus proved more complex than I thought...


aaronroberson:
@excentris Can you please update the wiki to document this new option?

excentris:
@aaronroberson Done! Thank you very much for merging this!

aaronroberson:
You're welcome. We try to stay away from adding too many options, supporting every possible feature and configuration because it can become kludgy and even more difficult to manage. However, enough people have needed this feature that I let it through ;)

PR #1492, opened 2016-03-16
fix(uiSelectMultipleDirective): add $isEmpty handler

### LINKED ISSUE #850: required tag

kmccullough:
Fixes #850


aaronroberson:
@kmccullough Thanks, this will land in the next release!

tmanoel:
Using single select and it's not working yet.. I'm Using version 0.19.0

kmccullough:
@tmanoel I'm not sure what you are expecting. This fix is for ng-required not working on multiple selection. It should already work for single.

tmanoel:
Thanks for the attention, kmccullough. But I'm using it into a directive. I've tried use required and ng-requeired, still not working for single select.

tmanoel:
Take a look, it doesnt work on single neither multiple..
http://plnkr.co/edit/5Zt0OfEFh18tlcKP1Yen?p=preview

kmccullough:
Can you give us a better idea of exactly what you are expecting in that plunkr @tmanoel ? It appears you are using version 0.11.2 of ui-select, which you have copied to an actual file included with the page, but it does not include my fix from this commit. Please try the latest version via CDN, update your sources with the latest, or apply my fix from *src/uiSelectMultipleDirective.js* to your own sources.

PR #1494, opened 2016-03-16
chore(contributing): update contributing doc and issue template

deeg:
This updates the contributing guidelines to include some text about how to open bugs or feature requests. The idea is we can close issues which get opened without following the guidelines.

Also adds an issue template.


deeg:
@user378230 want to send me your email address so I can add you to the chat room?

aaronroberson:
Perfect! Although I don't think we should enforce the idea of closing old issues prior to us implementing these requirements.

PR #1505, opened 2016-03-18
Fix(templates): Autocomplete=false does not work in Edge

### LINKED ISSUE #1346: Autocomplete="false" Does not work in Edge Browser
On the Edge browser in Windows 10, autocomplete="false" does not work on the inputs, causing bizarre behavior in which auto-complete suggestions show up on the dropdown.

Autocomplete="off" seems to work however, and according to [w3c autocomplete](http://www.w3schools.com/tags/att_input_autocomplete.asp) is supposed to be an on/off anyways, correct?

MattiJarvinen-BA:
Edge browser showed autocomplete with autocomplete=false Standard states
it should be autocomplete=off which works currently with Chrome too.

Closes #1346

### DIFF
--- a/src/bootstrap/select-multiple.tpl.html
+++ b/src/bootstrap/select-multiple.tpl.html
@@ -2,7 +2,7 @@
   <div>
     <div class="ui-select-match"></div>
     <input type="text"
-           autocomplete="false" 
+           autocomplete="off" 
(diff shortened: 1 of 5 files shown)


Ian5015:
Thank you for creating a PR to fix this, it's causing our users issues in Edge and IE.

PR #1517, opened 2016-03-23
fix(uiSelectCtrl): correcting input focus

### LINKED ISSUE #1253: on focus not working for first click
When ui-select is loaded in the page for the first time, it is disabled by default.
But after focusing on other input boxes and focusing on the ui-select it is working as expected.

My HTML:

``
<ui-select multiple ng-model="newMachine.machine_devices">
                        <ui-select-match>
                            {{$item.Esn}}
                        </ui-select-match>
                        <ui-select-choices repeat="device in devices | filter:$select.search">
                            {{device.Esn}}
                        </ui-select-choices>
                    </ui-select>
``

js:
``
$scope.devices = [{DeviceID:1,Esn:'SCE112514000145182'},{DeviceID:2,Esn:'SCE112514000145185'}]
``

patrickhousley:
Corrected input focus when ui-select contains no elements. Also corrected a memory leaked caused by event handlers never being removed.

Closes #1253

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -123,15 +123,29 @@ uis.controller('uiSelectCtrl',
       }
 
       var container = $element.querySelectorAll('.ui-select-choices-content');
-      if (ctrl.$animate && ctrl.$animate.on && ctrl.$animate.enabled(container[0])) {
-        ctrl.$animate.on('enter', container[0], function (elem, phase) {
-          if (phase === 'close') {
+      var searchInput = $element.querySelectorAll('.ui-select-search');
+      if (ctrl.$animate && ctrl.$animate.enabled(container[0])) {
+        var animateHandler = function(elem, phase) {
+          if (phase === 'start' && ctrl.items.length === 0) {
             // Only focus input after the animation has finished
             $timeout(function () {
+              ctrl.$animate.off('removeClass', searchInput[0], animateHandler);
+              ctrl.focusSearchInput(initSearchValue);
+            });
+          } else if (phase === 'close') {
+            // Only focus input after the animation has finished
+            $timeout(function () {
+              ctrl.$animate.off('enter', container[0], animateHandler);
               ctrl.focusSearchInput(initSearchValue);
             });
           }
-        });
+        };
+
+        if (ctrl.items.length > 0) {
+          ctrl.$animate.on('enter', container[0], animateHandler);
+        } else {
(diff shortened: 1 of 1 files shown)


patrickhousley:
This PR corrects two issues:

1. The last modification I made to this section of the code created a memory leak where by events were repeatedly added to the .ui-select-choices-content element but never removed. Those will now be removed once the event is fired.
2. With regards to issue #1253, in the event the ui-select has no elements, the enter event is never fired for the .ui-select-choices-content element. To work around this, the code will fall back to using the removeClass event on the .ui-select-search element. Testing has shown that we could rely solely on this event but I cannot test all possible browsers.

user378230:
@wesleycho any chance you could review this if/when you get a chance? I don't have any experience with angular animate.

src/uiSelectController.js
```diff
@@ -123,15 +123,29 @@ uis.controller('uiSelectCtrl',
...
-      if (ctrl.$animate && ctrl.$animate.on && ctrl.$animate.enabled(container[0])) {
-        ctrl.$animate.on('enter', container[0], function (elem, phase) {
-          if (phase === 'close') {
+      var searchInput = $element.querySelectorAll('.ui-select-search');
+      if (ctrl.$animate && ctrl.$animate.enabled(container[0])) {
+        var animateHandler = function(elem, phase) {
+          if (phase === 'start' && ctrl.items.length === 0) {
             // Only focus input after the animation has finished
             $timeout(function () {
+              ctrl.$animate.off('removeClass', searchInput[0], animateHandler);
```
wesleycho: I don't think this needs to be in the $timeout

src/uiSelectController.js
```diff
@@ -123,15 +123,29 @@ uis.controller('uiSelectCtrl',
...
+              ctrl.focusSearchInput(initSearchValue);
+            });
+          } else if (phase === 'close') {
+            // Only focus input after the animation has finished
+            $timeout(function () {
+              ctrl.$animate.off('enter', container[0], animateHandler);
```
wesleycho: Ditto

patrickhousley: Whats the feeling here about using only the removeClass event? I am a little hesitant to change something that seemingly has been working.

hitchcockwill:
Just chiming in that it would be awesome to see this fix in the next release!

wesleycho:
This LGTM @user378230 - feel free to merge, unless you feel like this needs a test.

PR #1518, opened 2016-03-23
fix(select-choices): Possible fix for #1355

### LINKED ISSUE #1355: [ui.select:choices] Expected multiple .ui-select-choices-row but got '0'.
screen shot 2015-12-11 at 12 00 07 pm

Error site found at code below

screen shot 2015-12-11 at 12 02 23 pm

Created a silly workaround for this issue

````       
        setTimeout(function(){
          var container = $element.querySelectorAll('.ui-select-choices');
          for(var i = 0;  i < 2;  i++) {
            var div = document.createElement('div');
            div.className = 'ui-select-choices-row';
            container[0].children[0].appendChild(div);
          }
        });
````

Thanks!

dondi:
...at the cost of potential unnecessary watches when dropdown is closed.
But the premise is that, until the resulting errors are fixed, the
preference is to avoid the errors for now.

Closes #1355


wesleycho:
This has negative performance implications - it probably would be more ideal to figure out the root cause of the errors.

dondi:
OK; I just send it because it was mentioned as an ad hoc change in the thread for #1355 and users there were generally positive about it. The potential performance impact was a known compromise. I can look harder for the root cause then update the request.

dondi:
Latest upstream source tree appears to remove the original #1355 issue; closing this one.

PR #1519, opened 2016-03-23
fix(uiSelect) support data-multiple attribute

### LINKED ISSUE #1451: Using HTML5 data-* attributes breaks the multi-select and theme
I'm using a HTML5 validator, so I add data- in front of all of my attributes that are not standard HTML attributes. When I do this with the "multiple" attribute for ui-select, this causes uiSelectMatch directive to load an incorrect template.
I believe the error comes from here https://github.com/angular-ui/ui-select/blob/master/src/uiSelectMatchDirective.js#L12-L13 .

user378230:
uiSelectMatch was a special case where the non-normalised `multiple` attribute is checked, therefore using `data-multiple` resulted in the single select match template being loaded instead of the multiple-select-match template.

test is more involved than I would have liked but I couldn't see any other way to test it. 

Closes #1451

### DIFF
--- a/src/uiSelectMatchDirective.js
+++ b/src/uiSelectMatchDirective.js
@@ -8,10 +8,12 @@ uis.directive('uiSelectMatch', ['uiSelectConfig', function(uiSelectConfig) {
       // Needed so the uiSelect can detect the transcluded content
       tElement.addClass('ui-select-match');
 
+      var parent = tElement.parent();
       // Gets theme attribute from parent (ui-select)
-      var theme = tElement.parent().attr('theme') || uiSelectConfig.theme;
-      var multi = tElement.parent().attr('multiple');
-      return theme + (multi ? '/match-multiple.tpl.html' : '/match.tpl.html');
+      var theme = getAttribute(parent, 'theme') || uiSelectConfig.theme;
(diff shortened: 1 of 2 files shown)


wesleycho:
I'm not sure what your claimed support is, but here's a snippet in UI Bootstrap for what we check for: https://github.com/angular-ui/bootstrap/blob/master/src/tabs/tabs.js#L224-L234

src/uiSelectMatchDirective.js
```diff
@@ -10,7 +10,10 @@ uis.directive('uiSelectMatch', ['uiSelectConfig', function(uiSelectConfig) {
 
       // Gets theme attribute from parent (ui-select)
       var theme = tElement.parent().attr('theme') || uiSelectConfig.theme;
-      var multi = tElement.parent().attr('multiple');
+
+      var multi = angular.isDefined(tElement.parent().attr('multiple')) ||
```
wesleycho: Probably worth caching `tElement.parent()` above and rewriting to remove duplicate calls to `.parent()`

user378230:
Yeah I was being somewhat lazy, updated if you get a chance to review. .

wesleycho:
LGTM

PR #1523, opened 2016-03-25
refactor(uiSelectChoices) remove setActiveItem refs

user378230:
setActiveItem was removed as part of PR #1211 but not all calls to the function were.


wesleycho:
LGTM

PR #1529, opened 2016-03-27
Fix tolerance for null or undefined values

homerjam:
Re-pull of #1035

### DIFF
--- a/src/uiSelectMultipleDirective.js
+++ b/src/uiSelectMultipleDirective.js
@@ -156,7 +156,7 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
         if(!angular.isArray(ngModel.$viewValue)){
           // Have tolerance for null or undefined values
           if(angular.isUndefined(ngModel.$viewValue) || ngModel.$viewValue === null){
-            $select.selected = [];
+            ngModel.$viewValue = [];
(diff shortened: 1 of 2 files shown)


user378230:
Was this causing an actual issue? (Thinking in terms of if we can add a test for this somehow..)

homerjam:
I did this a while ago but yes I believe it was causing the directive to break. I think a test would just need to give a null/undefined value for the model.

user378230:
@homerjam can you add one and then I can merge in? It's best for them to be part of the same commit, ideally.

homerjam:
Sure. I'll definitely give it a go

user378230:
@homerjam just to confirm, these test fail without your change right?

wesleycho:
Those tests currently pass without the changeset from the PR.

homerjam:
Ah yes I meant to to say. I think this fix may have been to make sure the
placeholder text appeared on a null/undefined value (or something like
that, have been meaning to check). I can't recall now unfortunately. Would
seem its unnecessary now though.

user378230:
Will close for now, it looks a bit silly admittedly but there's no need to introduce unnecessary change at the risk of breaking something else. Maybe that section will refactored at some point in the future, anyway. 

If anyone finds a need for this change please update with details and we can look at reopening this.

PR #812, opened 2015-04-01
Update package name.  ui-select is a different project

MichaelJCole:
PR for #779

ui-select is a different project.  Updated the bower.json to use the same package name used when installing.

PR #1532, opened 2016-03-28
chore(bower) update package name

user378230:
ui-select is a different package, angular-ui-select is the correct
package name.


user378230:
Supersedes #812

PR #1533, opened 2016-03-28
fix(uiSelectChoices) broken id interpolation

### LINKED ISSUE #1522: track by $index broken with angular 1.5
### Bug description:

When using `track by $index` for the `repeat` attribute, with angular < 1.5, ui-select used to generate an id of the form

```
<div id="ui-select-choices-row-0-2" …>
```

Now, with angular 1.5(.2) we get :

```
<div id="ui-select-choices-row-0-" …>
```

The last digit is missing. It should be replace by the value of `$index`, but `$index is undefined`. 

### I’ve written a test for the issue :

gorghoa/ui-select@d627eef45755d7ac0e4c81a231989f576d0512c8

The test is ok with angular 1.4.10, broken with 1.5

Angular: 1.5.2

Let me know if I can do more.
thanks!

user378230:
The id attribute on ui-select-choices-row is no longer interpolated correctly when using ui select with angular v1.5+.

This change switches to using ng-attr-id to ensure the id value is correctly interpolated before being assigned to the element.

Fixes #1522

### DIFF
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -399,6 +399,23 @@ describe('ui-select tests', function() {
     expect(getMatchLabel(el)).toEqual('Samantha');
   });
 
+  it('should correctly render initial state with track by $index', function () {
+    
+    var el = compileTemplate(
+      '<ui-select ng-model="selection.selected"> \
+        <ui-select-match placeholder="Pick one...">{{$select.selected.name}}</ui-select-match> \
+        <ui-select-choices repeat="person in people track by $index"> \
+          {{person.email}} \
+        </ui-select-choices> \
+      </ui-select>'
+    );
+
+    openDropdown(el);
+
+    var generatedId = el.scope().$select.generatedId;
+    expect($(el).find('[id="ui-select-choices-row-' + generatedId + '-0"]').length).toEqual(1);
+  });
+
   it('should utilize wrapper directive ng-model', function() {
     var el = compileTemplate('<wrapper-ui-select ng-model="selection.selected"/>');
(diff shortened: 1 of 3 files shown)


test/select.spec.js
```diff
@@ -399,6 +399,23 @@ describe('ui-select tests', function() {
...
+        <ui-select-choices repeat="person in people track by $index"> \
+          {{person.email}} \
+        </ui-select-choices> \
+      </ui-select>'
+    );
+
+    openDropdown(el);
+
+    var generatedId = el.scope().$select.generatedId;
+    expect($(el).find('[id$="row-' + generatedId + '-0"]').length).toEqual(1);
```
wesleycho: You can change this to `$(el).find('#row-' + generatedId + '-0')`

wesleycho:
Other than the minor adjustment, LGTM.

gorghoa:
Hi guys, sorry for the noise, I should not create pull request this early in the morning…

The full id selector should be `'#ui-select-choices-row-' + generatedId + '-0'`

:coffee:

gorghoa:
Dumb convention question: Should’nt the build files be included in the PR ?

test/select.spec.js
user378230: Would this capture all elements if they shared that id? (In case a bug was introduced)

I haven't jQuery'd in years, according to http://stackoverflow.com/a/6744674 it wouldn't? I'll have to check later.

wesleycho: Ah, I see what you're doing here - I think it would be better to just use the whole id, for simplicity.
```js
$(el).find('#ui-select-choices-row-' + generatedId + '-0')
```

PR #1536, opened 2016-03-29
feat(perf): debounce resize callback

wesleycho:
- The resize event fires often, so debouncing the function execution can
  improve performance dramatically


user378230:
LGTM :+1:

PR #1541, opened 2016-03-30
fix(build): fix sourcemap logic

wesleycho:
- Fix sourcemap generation
- Upgrade plugins


deeg:
This LGTM

gulpfile.js
```diff
@@ -74,16 +74,15 @@ gulp.task('scripts', ['clean'], function() {
...
 gulp.task('styles', ['clean'], function() {
 
-  return gulp.src('src/common.css')
+  return gulp.src(['src/common.css'], {base: 'src'})
+    .pipe($.sourcemaps.init())
     .pipe($.header(config.banner, {
       timestamp: (new Date()).toISOString(), pkg: config.pkg
     }))
     .pipe($.concat('select.css'))
-    .pipe(gulp.dest('dist'))
```
user378230: @wesleycho I've had to restore this line locally to get `select.css` saved to dist, are you sure it should have been removed?

wesleycho: It's not necessary to have sourcemaps unless it's minified, no?

user378230: Yeah I think what you're saying is correct, double negatives and all :eyes:

We are always minifying though, I think the maps are generated fine but the unminified css isn't saved to dist. 

With:
```js
  return gulp.src(['src/common.css'], {base: 'src'})
    .pipe($.sourcemaps.init())
    .pipe($.header(config.banner, {
      timestamp: (new Date()).toISOString(), pkg: config.pkg
    }))
    .pipe($.concat('select.css'))
    .pipe(gulp.dest('dist'))
    .pipe($.minifyCss())
    .pipe($.concat('select.min.css'))
    .pipe($.sourcemaps.write('../dist', {debug: true}))
    .pipe(gulp.dest('dist'));
```
I get
```
  - select.min.css
  - select.min.css.map
  - select.min.js
  - select.min.js.map
  - select.css
  - select.js
```

With:
```js
 return gulp.src(['src/common.css'], {base: 'src'})
    .pipe($.sourcemaps.init())
    .pipe($.header(config.banner, {
      timestamp: (new Date()).toISOString(), pkg: config.pkg
    }))
    .pipe($.concat('select.css'))
    //.pipe(gulp.dest('dist'))
    .pipe($.minifyCss())
    .pipe($.concat('select.min.css'))
    .pipe($.sourcemaps.write('../dist', {debug: true}))
    .pipe(gulp.dest('dist'));
```
I get:
```
  - select.min.css
  - select.min.css.map
  - select.min.js
  - select.min.js.map
  - select.js
```
Note the missing `select.css`, easy enough to fix just wanted to be sure I was doing it right.

PR #1547, opened 2016-03-31
chore: normalize line endings

### LINKED ISSUE #1546: Normalise line endings
Similar to https://github.com/angular-ui/bootstrap/issues/423 I think we need to normalise the line endings in the repo as they are beginning to become mixed up. 

I assume it's still done the same way as above (with [.gitattributes](https://github.com/angular-ui/bootstrap/blob/master/.gitattributes))

wesleycho:
- This adds the necessary files (.gitattributes and .editorconfig) to ensure line endings use lf

Note that this modifies some existing files automatically to conform (I did not manually make those changes).

This fixes #1546.


deeg:
This LGTM

PR #1555, opened 2016-04-04
refactor(uiSelectChoices) only compile template once

### LINKED ISSUE #1000: Third party directives added to ui-select-choices is created 2 times

user378230:
Previously the template was compiled, edited, and then recompiled. This was unecessary as the changes we unrelated to the scope and so could be made before the link function was run.

This change, moves the DOM transformations to the compile function so the directive does not have to be compiled manually.

Fixes #1000


wesleycho:
LGTM

PR #1557, opened 2016-04-06
fix(ui-select): avoid dropdown flickering on search

marco-jantke:
I hope commit message and subject are descriptive enough :) This is a fix for #1298 with no other side effects then the flickering.

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -295,6 +295,13 @@ uis.directive('uiSelect',
         scope.$watch('$select.open', function() {
 
           if ($select.dropdownPosition === 'auto' || $select.dropdownPosition === 'up'){
+            // Hide the dropdown to avoid flickering on initial opening
+            dropdown = angular.element(element).querySelectorAll('.ui-select-dropdown');
+
+            if (dropdown.length > 0) {
+              dropdown[0].style.opacity = 0;
+            }
+
             scope.calculateDropdownPos();
           }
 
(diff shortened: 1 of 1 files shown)


wesleycho:
This LGTM

src/uiSelectDirective.js
```diff
@@ -331,9 +338,6 @@ uis.directive('uiSelect',
               return;
             }
 
-            // Hide the dropdown so there is no flicker until $timeout is done executing.
-            dropdown[0].style.opacity = 0;
```
user378230: Removing this introduces a problem when dropdown is positioned up

http://plnkr.co/edit/rM3rdaekwqG8wN77Hgb6?p=preview

Enter text in select2 "up" and delete with `backspace` quickly (I typed `ada` to get most results).

With this line removed the select gets shifted/overlayed with the list as a flicker. (Comment/uncomment line ~1261)

We need to detect if dropdown is positioned up and apply this along with your additions on line 298.

marco-jantke:
Thanks a lot for feedback! I will have a look tomorrow.

marco-jantke:
I had a look again in this issue. Sorry for the time delay guys..

I see the point of @user378230, when you reduce your search and thus extend the list again, the dropdown flickers for a short moment below the placeholder and then positions up. Though I can not figure out what is actually causing this or how to detect it..

I set up a plunkr with a couple of log messages where I tried to figure this out. Maybe you can give some useful input where to continue the investigation.

Please have a look in this Plunkr on lines 1259 and following. http://plnkr.co/edit/7NGlK9SRi95jNYTqpX1D?p=preview

user378230:
@marco-jantke a shot in the dark but I wonder if the changes in #1132 help at all?

rpocklin:
This fixes the flicker on search, but there was a reason this was added in the first place 
```// Hide the dropdown so there is no flicker until $timeout is done executing.```

The goal should be to avoid the opacity reset if the dropdown is already visible, i've created a tweaked PR in #1594

PR #1562, opened 2016-04-11
fix tagging for all and feat tagging-equals function

Teegrounet:
fix tagging for all (I see that this was fixed in PR #972, but I can't find that commit either in the 0.16.1 tag, nor master.)
add tagging-equals option to verify if element exist in list before insert without some properties of object
add some tests


user378230:
#972 has _not_ been merged it still needs work I don't see a reason to include it in this PR though its not well related to the other change. 

Would it make sense to just have a generic custom equals which can be used everywhere? Or is there a reason you only added it for tagging?

It looks like you also have issues with indents too.

Teegrounet:
I need this function to compare the element create by tagging (which can be not contains all info) with item of the list.

For example : 
I have a list of countries (object with an id of my DB and a name), when user search an element, the tagging element doesn't have id 
-> we need to compare with country name and angular.equals can't be used.

user378230:
As previously stated there are issues with your indenting. 

Also you need to rebase not merge. 

Finally, if you're fixing bugs in the library can you submit this as a separate PR. This makes it easier to track and revert if necessary. 

I suggest also you have a read of [CONTRIBUTING.md](https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md)

If you have any questions please ask. Thanks! :+1:

Teegrounet:
I am sorry but I read the CONTRIBUTING.md but I don't understand all it write... It's very difficult to begin on GitHub...

I understand it's not a good thing to commit all in one (and without issues), but what do you want I do now ?

user378230:
Closing due to age/poor quality, sorry!

PR #1563, opened 2016-04-16
fix(uiSelectSingleDirective): strictly compare matching value

### LINKED ISSUE #1328: Issues when using values `''` and `false`
I have a bunch of items as so:

``` js
scope.items = [{
  label: '-- None Selected --',
  value: ''
}, {
  label: 'Yes',
  value: true
}, {
  label: 'No',
  value: false
}];

scope.model = '';
```

My HTML:

``` html
<ui-select ng-model="model">
  <ui-select-match>{{ $select.selected.label }}</ui-select-match>
  <ui-select-choices repeat="item.value as item in items track by item.value">{{ item.label }}</ui-select-choices>
</ui-select>
```

I'm expecting when the select box loads up to see the `--None Selected--` item, since my model is set to the empty string, `''`. However, due to [this line](https://github.com/angular-ui/ui-select/blob/4467b8238866f85003f8815f44a7ca8563222043/src/uiSelectSingleDirective.js#L28) using a double equals instead of a triple equals, the `No` item is actually displayed, since the formatter traverses from the end of the items array to the beginning, and when it finds the item with the `false` value, it compares `false == ''` which returns `true`.

Is there any historical reason why this is a double equals instead of a triple equals? Would you accept a PR which changes this to triple equals?

fiznool:
Fixes #1328

### DIFF
--- a/src/uiSelectSingleDirective.js
+++ b/src/uiSelectSingleDirective.js
@@ -25,7 +25,7 @@ uis.directive('uiSelectSingle', ['$timeout','$compile', function($timeout, $comp
           var checkFnSingle = function(d){
             locals[$select.parserResult.itemName] = d;
             result = $select.parserResult.modelMapper(scope, locals);
-            return result == inputValue;
+            return result === inputValue;
           };
           //If possible pass same object stored in $select.selected
           if ($select.selected && checkFnSingle($select.selected)) {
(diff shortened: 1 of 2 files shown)


user378230:
Perfect, thanks!

PR #1567, opened 2016-04-19
feat(removeSelected): Make removeSelected configurable

dondi:
Expose the `removeSelected` property, and adjust `isDisabled` behavior so that items that would have been removed from a `multiple` ui-select element because they are currently selected are disabled instead.

Sample behavior:
screen shot 2016-04-18 at 6 07 42 pm

In the existing version, `id` would not have appeared in the dropdown because it is currently selected. This pull request adds the option `removeSelected` so that, if set to `false`, selected items still appear in the dropdown but are disabled (like in the screenshot).


user378230:
@dondi The code itself looks ok but there aren't any unit tests. Can you add some and I'll review again?

Thanks! :+1:

dondi:
Ack, sorry! I had fixed some unit tests that broke due to this change but then got distracted that day and totally forgot about adding _new_ ones. Sorry, my bad, I will notify you when I’ve put something up.

thomasmiddeldorp:
@dondi Is this PR still something you're working on recently? The functionality this PR contains is something I'm looking for as well, so I could take over the PR if it doesn't fit in your schedule =)

dondi:
Thanks very much for the offer @thomasmiddeldorp 👍  As it turns out, I was chipping away at it and managed to have enough time to wrap things up this morning. Had a learning curve to climb there for a bit. I just pushed both unit tests and a demo—hope it's what you're looking for.

user378230:
Almost perfect @dondi! Can you just squash down into a single `feat(removeSelected)` commit? There is a lot of noise in the commit history some of which has been cancelled out. 

Once that's done I'll merge :+1: 

_Useful reference: [Commit Message Guidelines](https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md#commit)_

dondi:
Ah OK, I'll look into that…I presume the "merge from master" commit that is in between there is part of this noise. Yeah I can see how that would be problematic. Would it work better if I create a fresh branch off upstream then place my changed files there, do a single commit, and issue a new pull request?

dondi:
OK, I have a new pull request using a fresh branch that is updated from upstream. Hope that will be OK.

New pull request: https://github.com/angular-ui/ui-select/pull/1622

user378230:
You've opened the new one now but you can just force push to your existing branch after squashing, github would have automatically updated this PR with the new single commit. :-)

dondi:
Yes, I intended to do that at first, but after looking more closely at the overall commits, there were a couple that were no longer valid plus the merge from master commit was in between commits of mine that I wanted to keep, with potential dependent changes in between. So in the end it was more straightforward to start fresh.

PR #1571, opened 2016-04-23
perf: change test in ctrl.isActive

Teegrounet:
perf: change test in ctrl.isActive

before :
( itemIndex < 0 && ctrl.taggingLabel !== false ) ||( itemIndex < 0 && ctrl.taggingLabel === false)
after :
itemIndex < 0


user378230:
I saw this too. Thanks

PR #1572, opened 2016-04-23
Fix taggingLabel initialization for single mode

Teegrounet:
Fix taggingLabel initialization for single mode (when it works)

When taggingLabel is not define in ui-select tag and tagging is active with string (not object), the taggingLabel is undefine and we can see "<value> undefined" in the list.
With this code, taggingLabel is initialized with '(new)' as it could be when taggingLabel tag is present whithout value.


user378230:
Is it possible to add a test for this?

Teegrounet:
I think it's not possible as long as we don't have PR 972, because we can't seeing tagging element in single mode.

PR #1573, opened 2016-04-24
fix (tagging): try all taggingTokens for splitting on paste

### LINKED ISSUE #1566: Splitting on Paste only supports first token
### Bug description:
Splitting on pasted data only supports splitting on the first token.

[literally any working tagging ui-select instance - see select.js ln 826]

UI-Select: 0.16.1

This isn't a bug in that the code is doing something unexpected, so much as it is unexpected that the code would be written this way. An improvement would be to iterate over splitting tokens and split on the first token that exists in the string.

torstenrudolf:
loop through all `taggingTokens` and use first token that is contained in paste data for splitting (previously only first token was used)

Closes #1566

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -588,9 +588,18 @@ uis.controller('uiSelectCtrl',
     if (data && data.length > 0) {
       // If tagging try to split by tokens and add items
       if (ctrl.taggingTokens.isActivated) {
-        var separator = KEY.toSeparator(ctrl.taggingTokens.tokens[0]);
-        var items = data.split(separator || ctrl.taggingTokens.tokens[0]); // split by first token only
-        if (items && items.length > 0) {
+        var items = [];
+        for (var i = 0; i < ctrl.taggingTokens.tokens.length; i++) {  // split by first token that is contained in data
+          var separator = KEY.toSeparator(ctrl.taggingTokens.tokens[i]) || ctrl.taggingTokens.tokens[i];
+          if (data.indexOf(separator) > -1) {
+            items = data.split(separator);
+            break;  // only split by one token
+          }
+        }
+        if (items.length === 0) {
+          items = [data];
+        }
+        if (items.length > 0) {
         var oldsearch = ctrl.search;
           angular.forEach(items, function (item) {
             var newItem = ctrl.tagging.fct ? ctrl.tagging.fct(item) : item;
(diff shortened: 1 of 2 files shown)


test/select.spec.js
```diff
@@ -2504,6 +2504,14 @@ describe('ui-select tests', function() {
       expect($(el).scope().$select.selected.length).toBe(3);
     });
 
+    it('should split pastes on ENTER, even though ENTER is not first taggingToken, but not try another token', function() {
```
user378230: Can you turn this into two tests, one for each bit of functionality?
Example:

1. `it('should split pastes on token that isn't the first token',...)`
2. `it('should split pastes only on first tagging token found in paste string', ...)`

src/uiSelectController.js
```diff
@@ -588,8 +588,17 @@ uis.controller('uiSelectCtrl',
     if (data && data.length > 0) {
       // If tagging try to split by tokens and add items
       if (ctrl.taggingTokens.isActivated) {
-        var separator = KEY.toSeparator(ctrl.taggingTokens.tokens[0]);
-        var items = data.split(separator || ctrl.taggingTokens.tokens[0]); // split by first token only
+        var items = [];
+        for (var i = 0; i < ctrl.taggingTokens.tokens.length; i++) {  // split by first token that is contained in data
+          var separator = KEY.toSeparator(ctrl.taggingTokens.tokens[i]) || ctrl.taggingTokens.tokens[i];
+          if (data.indexOf(separator) !== -1) {
```
user378230: Very minor but `data.indexOf(separator) > -1` is a bit more concise.

src/uiSelectController.js
```diff
@@ -588,8 +588,17 @@ uis.controller('uiSelectCtrl',
...
+            items = data.split(separator);
+            break;  // only split by one token
+          }
+        }
+        if (items.length === 0) {
+          items = [data];
+        }
         if (items && items.length > 0) {
```
user378230: Can the `items` truthy check can be removed now? (Since you initalised items above L591: `var items = [];`)

torstenrudolf:
@user378230 good suggestion. I split the test into two separate tests

torstenrudolf:
@user378230 good suggestions. I believe I've fixed everything.

one question; now I did a force-push  to not litter the commit history in the `ui-select` project. But because of that we lost your line comments on the former commit. What would be the appropriate way to do this cleaning work?

user378230:
Thanks for your speedy update! :smiley: 

I think that workflow is expected. This LGTM now so I'll merge. 

Thanks again :+1:

PR #1585, opened 2016-04-28
feat(ui-select-choices): expression based minimum-input-length

tystol:
Allows scope expressions to be used to calculate a minimum-input-length filter for the ui-select-choices directive.

Useful for showing all options when the option list is small, but applying a minimum when the list is large eg:

`minimum-input-length="people.length > 50 ? 3 : 0"`

### DIFF
--- a/src/uiSelectChoicesDirective.js
+++ b/src/uiSelectChoicesDirective.js
@@ -68,7 +68,8 @@ uis.directive('uiSelectChoices',
         scope.$watch('$select.search', function(newValue) {
           if(newValue && !$select.open && $select.multiple) $select.activate(false, true);
           $select.activeIndex = $select.tagging.isActivated ? -1 : 0;
-          if (!attrs.minimumInputLength || $select.search.length >= attrs.minimumInputLength) {
+		  var minimumInputLength = attrs.minimumInputLength ? scope.$eval(attrs.minimumInputLength) : null;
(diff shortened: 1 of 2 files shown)


user378230:
Your indenting is slightly off. 

But I think maybe for consistency we follow the same pattern as the other observable attributes?..
```js
attrs.$observe('refreshDelay', function() {
  // $eval() is needed otherwise we get a string instead of a number
  var refreshDelay = scope.$eval(attrs.refreshDelay);
  $select.refreshDelay = refreshDelay !== undefined ? refreshDelay : uiSelectConfig.refreshDelay;
});
```

tystol:
How's this?

user378230:
That looks good now. 

Just to check, something like `minimum-input-length="{{people.length > 50 ? 3 : 0}}"` isn't supported already is it?

tystol:
****facepalm****

I did try that first off directly within my app and it didn't work. Not sure what I did wrong, but it is working fine in this plunker:
http://plnkr.co/edit/K6EwjGZsetp4FsQEY9Hv?p=preview

So you can probably close this PR, unless there is any other reason to prefer the bracketless expressions?

jamespsterling:
Still not working for me using the latest release, the $scope.$select.minimumInputLength variable is never updated after the initial assignment on load.

PR #1590, opened 2016-05-04
Cannot read property 'filter' of null

melnikaite:
fix(uiSelectCtrl): change condition that checks if data is empty before applying filter

Error `Cannot read property 'filter' of null` occurs after initializing ui-select


user378230:
Hi, thanks for your PR. Is there a unit test you can add for this?

PR #1594, opened 2016-05-09
Remove flicker on search but preserve original hide when dropdown not visible

rpocklin:
Remove flicker on search but preserve original hide when dropdown not visible

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -323,6 +323,8 @@ uis.directive('uiSelect',
 
         };
 
+        var opened = false;
+
         scope.calculateDropdownPos = function(){
 
           if ($select.open) {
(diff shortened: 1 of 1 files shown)


aaronrenner:
Works for me 👍

src/uiSelectDirective.js
```diff
@@ -331,8 +333,11 @@ uis.directive('uiSelect',
               return;
             }
 
-            // Hide the dropdown so there is no flicker until $timeout is done executing.
-            dropdown[0].style.opacity = 0;
+           // Hide the dropdown so there is no flicker until $timeout is done executing.
+           if ($select.search === '' && !opened) {
+              dropdown[0].style.opacity = 0;
+              opened = true;
```
user378230: @rpocklin Does `opened` never need to be reset to `false`?

rpocklin: Not that i've found, since it's already been rendered once - AFAIK the opacity change was for the initial render only.

Yurko-Fedoriv:
@rpocklin thanks for the fix, it's very helpful, I ported it to my fork, since I need to integrate it, and I'm having one note:
you are defining ```opened``` in the top most global scope, which is not very cool. Here is my version, it probably has slightly better code smell:

https://github.com/nutritionix/ui-select/blob/f20c4d45feb2f21701bd579b19c9a90c672994e6/src/uiSelectDirective.js#L373-L376

rpocklin:
@Yurko-Fedoriv not sure you know what you're doing here.  You're adding a variable under a function declaration.  My version is not using **global scope**, its within the directive in the `compile` function.  It can only be initially opened once anyway, so it's fine in this case.

Yurko-Fedoriv:
ah, sorry, nvm, I somehow missed outer declaration.

Issue #1588, opened 2016-05-02
Dropdown remains visible when open list is empty if using bootstrap theme


nfeldman-at-onshape:
### Bug description:
As a consequence of https://github.com/angular-ui/ui-select/commit/e65995f1eb2e190e71c9226a2f754671a7277c86#diff-4c6dab330fc9c181481aa33e5feb15b9 an open select list is visible even if there are no options to choose from. 

Activate the first select box in the following plnkr
http://plnkr.co/edit/N1AlvedLTmSr2BDsNvSi?p=preview
to see.

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular:
1.3.15 and 1.5.0

UI-Select:
0.16.1

Bootstrap/Select2/Selectize CSS (if applicable):
bootstrap 3.1.0 and 3.1.1

PR #1598, opened 2016-05-11
fix(select2): hide dropdown if there are no items to show

torstenrudolf:
this is equivalent to #1588 for bootstrap

the empty dropdown behaviour was especially annoying when using ui-select for tagging without choices.

### DIFF
--- a/src/select2/select-multiple.tpl.html
+++ b/src/select2/select-multiple.tpl.html
@@ -26,7 +26,7 @@
     </li>
   </ul>
   <div class="ui-select-dropdown select2-drop select2-with-searchbox select2-drop-active"
-       ng-class="{'select2-display-none': !$select.open}">
+       ng-class="{'select2-display-none': !$select.open || $select.items.length === 0}">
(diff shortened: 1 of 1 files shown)


user378230:
Is there actually a dropdown displayed? I switched plunkr in #1588 to select2 theme and I didn't see anything that looked wrong. 

Can you add a screenshot & plunkr demo of the issue? (And possibly one with your fix?)

torstenrudolf:
@user378230 
hmm, it might actually be in a different situation. Namely when you use the ui-select for tagging with `tagging-label="false"` and no choices.

Please have a look at the following (plunker)[http://plnkr.co/edit/42MxM2XbLiANgr0qPO1K?p=preview]
with the select2-theme you get a dropdown when you click into the input but with the bootstrap theme you don't.

This might be a side affect of #1588

user378230:
Sorry I was looking at select2 single mode when the problem was in select2 multiple mode. 

This looks ok then, thanks!

PR #1622, opened 2016-05-25
feat(removeSelected): Implement removeSelected property for multiple selects

dondi:
Expose and implement `remove-selected="false"` for multiple selects. This will disable a choice
in the dropdown of a multiple-select element instead of removing it.

Sample behavior:
screen shot 2016-04-18 at 6 07 42 pm

In the existing version, `id` would not have appeared in the dropdown because it is currently selected. This pull request adds the option `removeSelected` so that, if set to `false`, selected items still appear in the dropdown but are disabled (like in the screenshot).

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -18,7 +18,7 @@ uis.controller('uiSelectCtrl',
   ctrl.refreshDelay = uiSelectConfig.refreshDelay;
   ctrl.paste = uiSelectConfig.paste;
 
-  ctrl.removeSelected = false; //If selected item(s) should be removed from dropdown list
+  ctrl.removeSelected = uiSelectConfig.removeSelected; //If selected item(s) should be removed from dropdown list
   ctrl.closeOnSelect = true; //Initialized inside uiSelect directive link function
   ctrl.skipFocusser = false; //Set to true to avoid returning focus to ctrl when item is selected
   ctrl.search = EMPTY_SEARCH;
@@ -240,9 +240,9 @@ uis.controller('uiSelectCtrl',
       }else{
         if ( data !== undefined ) {
           var filteredItems = data.filter(function(i) {
-            return selectedItems.every(function(selectedItem) {
+            return angular.isArray(selectedItems) ? selectedItems.every(function(selectedItem) {
               return !angular.equals(i, selectedItem);
-            });
+            }) : !angular.equals(i, selectedItems);
           });
           ctrl.setItemsFn(filteredItems);
         }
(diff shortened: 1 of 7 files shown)


src/uiSelectController.js
```diff
@@ -325,9 +330,9 @@ uis.controller('uiSelectCtrl',
     var isDisabled = false;
     var item;
 
-    if (itemIndex >= 0 && !angular.isUndefined(ctrl.disableChoiceExpression)) {
+    if (itemIndex >= 0 && (!angular.isUndefined(ctrl.disableChoiceExpression) || ctrl.multiple)) {
```
user378230: @dondi could you add an extra set of brackets here to make the condition precedence clear? 

Eg. 
```
if (itemIndex >= 0 && ((!angular.isUndefined(ctrl.disableChoiceExpression) || ctrl.multiple)))
```
Or..?
```
if ((itemIndex >= 0 && (!angular.isUndefined(ctrl.disableChoiceExpression)) || ctrl.multiple))
```

dondi: Actually the parentheses already disambiguate the precedence. The current code is shown below, with the matching middle parentheses spaced out for emphasis:

    if (itemIndex >= 0 &&   (  !angular.isUndefined(ctrl.disableChoiceExpression) || ctrl.multiple  )   ) {

Those parentheses ensure that the `||` in the middle takes precedence. The suggestions made above would result in double-parentheses on the same expression (the disjunction in the first example; the entire condition in the second example).

user378230: Sorry I totally misread it, my bad! :flushed:

dondi: No worries, all good—thanks for reviewing!

user378230:
@dondi did you want to update the [wiki](https://github.com/angular-ui/ui-select/wiki/ui-select) with some docs? (add a note saying available v0.17.2+)

Thanks again :+1:

dondi:
Great, thanks! I've added a section on this to the wiki. I made it an example section instead of adding it to the attribute table because it really only applies for `multiple`. (I noticed that `multiple` itself is not in that attribute table at the beginning, so I took my cue from there.) Let me know if the writeup needs any changes. Thank you again!

thomasmiddeldorp:
@user378230 Is there any indication on when this feature will be released?

marcbachmann:
Is it possible that this pr introduced a new bug in the normal dropdown? Wrong options get removed in my dropdown. After a reload they appear again. Also in older versions you didn't remove the option on select.
 
dropdown

dondi:
Yes, a bug was indeed caught and the fix has since been merged, presumably to emerge in the next release. Please see #1672 and #1681 to see if these behaviors match what you are seeing. You can try a build from `master` to see if the latest code addresses what you’re seeing.

PR #1625, opened 2016-05-27
933 Fix text overflow in placeholder and match text

### LINKED ISSUE #933: Width of the select when an item with large name is selected (question)
This is my ui-select:

```
<ui-select ng-model="selectedReceiver.selected" ng-disabled="fetchingReceivers || receivers.length == 0">
    <ui-select-match placeholder="">{{$select.selected.code}} - {{$select.selected.name}}</ui-select-match>
    <ui-select-choices repeat="receiver in receivers | filterBy: ['name', 'code']: $select.search"> 
        <div ng-bind-html="receiver.name | highlight: $select.search"></div>
        <small>Code: <span ng-bind-html="receiver.code | highlight: $select.search"></span></small>
        <hr style="margin-top: 0px; margin-bottom: 0px; border-top: 1px solid #C0C0C0;"/>
    </ui-select-choices>
</ui-select>
```

When an item is selected, say:

```
     {
           code: "some large code",
           name: "some very large name, foo bar qux"
     }
```

the `ui-select-match` shows "some large code - some very large name, foo bar qux" (as expected), but this is so large that resizes the control (the `<button>` in which the directive is translated).

visual example

Question: Is there a setting for preventing this resize?

dsalvagni:
Just added a `display: inline-block` to the @rzschech [solution](https://github.com/angular-ui/ui-select/issues/933#issuecomment-116773277).

Fix #933

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -261,3 +261,12 @@ body > .ui-select-bootstrap.open {
 .ui-select-container[theme="bootstrap"].direction-up .ui-select-dropdown {
     box-shadow: 0 -4px 8px rgba(0, 0, 0, 0.25);
 }
+
+/* fix ellipsis on placeholder and match text */
+.ui-select-placeholder, .ui-select-match-text {
+  width: 100%;
+  overflow: hidden;
+  display: inline-block;
+  text-overflow: ellipsis;
+  padding-right: 40px;
+}


dsalvagni:
@Toilal Can you send us a print screen? I couldn't simulate that here.

PR #1627, opened 2016-05-28
remove unused missing background image

wuliupo:
background-image: url(header.png);


DavidCain:
It's worth mentioning that this PR will restore the ability to collect static files when using ui-select with Django Pipeline. Since it expects `header.png` to be exist in the `assets` directory, `collectstatic` will raise the following error:

    ValueError: The file 'angular-ui-select/docs/assets/header.png' could not be found with 
    <pipeline.storage.CachedStorage object at [address]>

wuliupo:
why did we remove it?

DavidCain:
@wuliupo, there never was such a file in this repository:

```bash
$ git log --oneline --all -- **/header.png | wc -l
0
```

It would seem that it's just a sample URL. However, this still does cause problems for some utilities.

PR #1630, opened 2016-05-31
fix: make compatible with Angular 1.5 and non-cached templates

### LINKED ISSUE #1609: ui-select-no-choice isn't displayed
I have already downloaded the version 0.17.1, the directive "uiSelectNoChoice" is in file "dist/select.js", but when the search returns no results, it does not appear.

Doesn't work with ui-select add no-choice, [plunker link](http://plnkr.co/edit/T2HsEvQW1fBQJz3hg2dh?p=preview).

My code:
```
<ui-select ng-model="elementExtensions.ator" theme="bootstrap" fix-focus-ui-select>
    <ui-select-match allow-clear="true" placeholder="{{::'TYPE_FOR_SEARCH' | translate}}">
      {{$select.selected.name}}
    </ui-select-match>
    <ui-select-choices refresh="loadAsyncActors($select.search)" refresh-delay="300"
                       repeat="actor in actorPropertyViewControl.availableActors">
      <div ng-bind-html="actor.name | highlight: $select.search"></div>
    </ui-select-choices>
    <ui-select-no-choice>
      Dang! We couldn't find any choices...
    </ui-select-no-choice>
</ui-select>
```

HTML:
selecao_014

bogdanalexe90:
Same change introduced in 0e85670ed377148ac3a7d5ef6ff5ed7c46446858

Closes #1609


user378230:
Derp, seems so obvious now. Thank you. :+1:

PR #1638, opened 2016-06-04
Added setSearchToAnswer 

whitef0x0:
Added setSearchToAnswer, which, if true, sets the search input to the last chosen answer. It requires 

set-search-to-answer = true AND searchBy="valueThatYouAreIndexingBy"

Example:
```
<div class="col-xs-12 field-input">
		<ui-select ng-model="field.fieldValue"
                theme="selectize"
			    search-enabled="true"
			    search-by="option_value"
			    set-search-to-answer="true"
				ng-required="field.required"
				ng-disabled="field.disabled"
			    on-tab-and-shift-key="prevField()"
			    on-tab-key="nextField()"
				ng-change="$root.nextField()">
			<ui-select-match placeholder="Type or select an option">

			</ui-select-match>
			<ui-select-choices repeat="option in field.fieldOptions | filter: $select.search"
							   ng-class="{'active': option.option_value === field.fieldValue }">
				<span ng-bind-html="option.option_value | highlight: $select.search"></span>
			</ui-select-choices>
        </ui-select>    
    </div>
```


whitef0x0:
Couldn't find any @ng-doc comments in the current repo so I wasn't sure if I should add this, even though the ***Contributor Guidelines** states this as a requirement. Can you clarify whether this is a required step?

user378230:
Can you provide a typical usage scenario for this? What problem does it help solve?

I'm concerned that this might be heading towards a little feature bloat and might not be all that useful.

whitef0x0:
reset-search-input just reset the input to blank. Set search to answer sets the input to the last correct answer

whitef0x0:
See question #4 of this form for example usage https://stage.tellform.com/#!/forms/573f8329d18419854559c29c

user378230:
> reset-search-input just reset the input to blank.

Setting the value to `false` actually prevents this, that is why this attribute exists. (By default it is set to `true`).

whitef0x0:
Yes but I wanted it set to the currently set value, not the previous input which is different

user378230:
I'm still doubtful how useful this is, maybe you could try creating this as an addon directive checkout [Accessing `$select`](https://github.com/angular-ui/ui-select/wiki/Accessing-%24select) in the wiki for info.

whitef0x0:
Hmm ok I'll look into that

user378230:
@whitef0x0 did you succeed? It would be useful to share, I can add to wiki if so.

PR #1641, opened 2016-06-06
Change Property Prefix from "_" to "$" so that angular built-in functions ignore them

LucasThielmann:
Angular's built in functions like angular.copy() and angular.toJson() automatically ignore object properties that are prefixed with $.

This PR changes the names of "_uiSelectChoiceDisabled" and "_uiSelectChoiceLocked" to "$uiSelectChoiceDisabled" and "$uiSelectChoiceLocked" respectively to make use of this feature when using ui-select as a model for rest resources.


user378230:
Just a note that this could be a breaking change.

peterreisz:
I'd prefer another solution, because you also can use a string list in a repeat: #1661

PR #1646, opened 2016-06-07
feat(search box): add initSearchValue

### LINKED ISSUE #1645: Feature request: add initial search value
Our users found it confusing when the text in the search field gets cleared when the dropdown is opened, if there is already a selection. To fix this, I added an attribute called `init-search-value` that, if set, will be automatically entered in the search box on opening. Our use case for this is to set the init-search-value to the currently selected option, so that when the user opens the dropdown, their selection doesn't seem to disappear.

An example of the issue is actually on the already existing demo plunkr [here](http://plnkr.co/edit/a3KlK8dKH3wwiiksDSn2?p=preview).
To reproduce the issue, select an option with the last dropdown. Then close the dropdown by clicking on some of the whitespace around it. Then, if you click on the dropdown again, the text for what you had selected before disappears. 

This happens with the most recent version of ui-select, #0.17.1.

jshearer:
Set the search box's text to initSearchValue instead of empty string when the dropdown gets toggled.

NOTE: Attempted to write a valid test, but couldn't trigger `activate()` using either `openDropdown()` or `$(el).click()`.

More info in the comment. 

Closes #1645

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -18,6 +18,7 @@ uis.controller('uiSelectCtrl',
   ctrl.refreshDelay = uiSelectConfig.refreshDelay;
   ctrl.paste = uiSelectConfig.paste;
 
+  ctrl.initSearchValue = EMPTY_SEARCH;
   ctrl.removeSelected = uiSelectConfig.removeSelected; //If selected item(s) should be removed from dropdown list
   ctrl.closeOnSelect = true; //Initialized inside uiSelect directive link function
   ctrl.skipFocusser = false; //Set to true to avoid returning focus to ctrl when item is selected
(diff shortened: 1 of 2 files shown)


user378230:
I already left a comment in your issue I'm not sure this is useful feature to have in core library. 

I think it could probably be covered by having a separate directive, you can see [Accessing `$select`](https://github.com/angular-ui/ui-select/wiki/Accessing-%24select) in the wiki for examples. If you're successful be sure to post for others to see/use. :smiley:

PaulMougel:
Hi everyone. I'm facing the same issue as @jshearer, and tried to implement this behavior as a separate directive… without luck. I can't seem to find how to tell `$select` what to store when the dropdown is toggled.

user378230:
@PaulMougel Did you try overwriting `$select.search`?

PaulMougel:
Here is what I tried, which kinda works but doesn't feel idiomatic:
```javascript
app
.directive('myUiSelect', function() {
  return {
    require: 'uiSelect',
    link: function(scope, element, attrs, $select) {
      var _activate = $select.activate;
      $select.activate = function (initSearchValue, avoidReset) {
        if (initSearchValue === undefined) {
          initSearchValue = $select.ngModel.$viewValue;
        }
        _activate(initSearchValue, avoidReset);
      }
    }
  };
})
```

Issue #367, opened 2014-11-03
Multiple mode: Search text is not cleared on tab out


afcastano:
When the search value does not match any option and the user tabs out the select, the text remains visible. This might lead to confusions since the user could think the value is valid.
It doesn't happen if the user clicks out. In this case, the search text is deleted.

EpiphanyMachine:
+1 Uncommenting this solves the issue for me, thanks @iroa.  If there is an invalid or empty input I want it to reset to the matched (model) option.

```
// if(~[KEY.ESC,KEY.TAB].indexOf(key)){
// //TODO: SEGURO?
// ctrl.close();
// }
```

icfantv:
I can also confirm that the attribute does not work, but uncommenting that code does work.

jthibeaux:
+1

yuyaohshimo:
+1

miyukiw:
+1

logstown:
+1

icfantv:
Hey guys, please refrain from only adding `+1` to issues.  It adds nothing to the discussion and clutters up the thread.  If you would like a feature implemented or bug fixed and it's not getting the attention you feel it deserves we do support user-supplied PRs.

MichalJakubeczy:
+1

fabiowitt:
@iroa solution does work, but it loses the functionality of selecting an item when you press tab. The proper solution would be to clear it, only if no results were found.

mihai-vasilache:
I have the 0.17.1 version (released on 2016-05-16) and it is still not fixed. Do you have any intention to fix it?

user378230:
@mihai-vasilache feel free to submit a PR that fixes the issue (including unit tests)  will be happy to review and merge. :smiley:

ahocquet:
I came upon this issue today and I've made a PR fix it : #1659 
Be free to test it out and give me your feedback !

Red-3:
So @user378230 asks for a PR because too many people were +1-ing it (which I think means I agree, it's an issue for me, please fix), @ahocquet provides said PR and still this remains an issue.
This has been open since Nov 2014. Update please?

mikouaj:
+1
(not working in (0.19.6)

annalauraaldlc:
+1

rajesh6291:
+!

PR #1659, opened 2016-06-13
Add allow-free-text attribute

ahocquet:
Fixes the issue #367 

The attribute name allow-free-text almost says it all : it allows the user to enter a non-existing item and keep the value in the model when he presses the TAB key or blur the element by clicking outside it.


user378230:
Can you provide more information about what this actually does?

How does this differ from `tagging` mode?

Not enough information right now, sorry!

ahocquet:
`tagging` mode allows multiple values, whereas this attribute allows filling free text on a classic single value ui-select instance.

user378230:
We should take advantage of existing tagging functionality instead of duplicating it then. As people may which to run tagging functions etc. even in single select mode.

user378230:
Closing due to lack of update.

PR #1669, opened 2016-06-17
fix(tagging): Support paste with tagging enabled and tagging-label="false"

### LINKED ISSUE #1668: (tagging) paste not working if tagging-label="false" and tagging-tokens are set
### Bug description:
when using the tagging feature paste is not working if `tagging-label="false"` and `tagging-tokens` are set. see [this plunker](http://plnkr.co/edit/kauNS7YjckRfi616k4PT?p=preview)

I believe this is because the paste event calls `ctrl.select(item)` (see [uiSelectController.js#L614](https://github.com/angular-ui/ui-select/blob/v0.18.0/src/uiSelectController.js#L614)) for each item after splitting the pasted string. But in [uiSelectController.js#L356](https://github.com/angular-ui/ui-select/blob/v0.18.0/src/uiSelectController.js#L356) in the case of `ctrl.taggingLabel === false` `ctrl.select` overrides the passed in `item` with the current `ctrl.search` which was set empty in the `paste` event handler earlier. Therefore the passed in `item` is lost and note used.

angularjs: 1.5.0
ui-select: 0.18.0

torstenrudolf:
Fixes #1668

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -350,10 +350,12 @@ uis.controller('uiSelectCtrl',
 
       if (!item || !item._uiSelectChoiceDisabled) {
         if(ctrl.tagging.isActivated) {
-          // if taggingLabel is disabled, we pull from ctrl.search val
+          // if taggingLabel is disabled and item is undefined we pull from ctrl.search
           if ( ctrl.taggingLabel === false ) {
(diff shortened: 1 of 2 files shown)


user378230:
Looks like a perfect PR, thanks! :smile:

PR #1670, opened 2016-06-17
Correct the tagging demo page

patou:
The tagging demo page not works since the use of the controller as in documentation.


user378230:
Thanks!

PR #1676, opened 2016-06-22
select2 theme working well when the direction for the dropdown is set…

### LINKED ISSUE #1674: Select2 styles don't render dropdown is positioned 'up' and theme is globally set
### Select2 styles don't render dropdown is positioned 'up' and theme is globally set

[ Plunkr ](http://plnkr.co/edit/RgSNxa4bGTwdjABOoyDH?p=preview)

Click on the Always Up Select

hearsid:
Resolves #1674 , select2 theme working well when the direction for the dropdown is set to up, earlier it was working only when theme was set by the directive and not working for globally set theme

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -46,12 +46,14 @@ body > .select2-container.open {
 }
 
 /* Handle up direction Select2 */
-.ui-select-container[theme="select2"].direction-up .ui-select-match {
+.ui-select-container[theme="select2"].direction-up .ui-select-match,
+.ui-select-container.select2.direction-up .ui-select-match {
     border-radius: 4px; /* FIXME hardcoded value :-/ */
     border-top-left-radius: 0;
     border-top-right-radius: 0;
 }
-.ui-select-container[theme="select2"].direction-up .ui-select-dropdown {
+.ui-select-container[theme="select2"].direction-up .ui-select-dropdown,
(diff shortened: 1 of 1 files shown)


user378230:
The definitions are the same as the ones already there for:
```
.ui-select-container[theme="select2"].direction-up ...
```
instead of duplicating the CSS can't you add another set of selectors for:
```
.ui-select-container.select2.direction-up ...
```
?

Also these changes should be made to `/src/common.css` not `/dist/select.css` please undo changes to `/dist` files and squash to single commit. 

Thanks

hearsid:
Hello @user378230, 
1) Have added the css selectors to the existing CSS.
2) Have added the code to src/common.css
3) I also ran gulp build after making the change so that the changes get reflected in the dist folder files, please inform me if I have to undo that.

user378230:
Please undo changes to all `/dist/` files

user378230:
Can you format correctly? Standard practice is to place the new selector on a new line..

```css
.ui-select-container[theme="select2"].direction-up .ui-select-dropdown .select2-search,
.ui-select-container.select2.direction-up .ui-select-dropdown .select2-search {
```

hearsid:
@user378230, thank you for the clear communication and quick reply, have made the change.

user378230:
Thanks!

Issue #1576, opened 2016-04-25
Clear button jump to new line


maxbundchen:
When the selected text is bigger than the select element the clear button jump to a new line. In the original Select2 the text get truncated with ellipsis before the clear button. Is possible duplicate this feature? It's looks like a css adjust.

user378230:
Can you please update your post to follow the [**issue template**](https://github.com/angular-ui/ui-select/blob/master/.github/ISSUE_TEMPLATE.md) and provide the version numbers and a demo plunkr?

This will make it easier for anyone who wants to submit a PR fix. 

If you know a fix, please feel free to open a PR yourself, thanks!

maxbundchen:
Here is the [plunkr](https://plnkr.co/edit/H87bI8nPfagKyjguT9qP?p=preview).

As you can see the long names should be truncated (ellipsis is a good option as used in the original select and select2 jquery components). Also the clear button is in the wrong place. I'm trying solve this with css but at the moment I couldn't find any solution.

bootstrap: 3.3.6
ui-select: 0.16.1
angular: 1.5.3

user378230:
There was a previous attempt #1337 but it pushed the clear button to a new line, maybe it will help you start.

PR #1680, opened 2016-06-22
fix(bootstrap layout): Restrict selected length to ui-select-container

burka:
In bootstrap theme, if you have a ui-select with a given width and a long selected option it can overflow its box.

Reproduced with most recent versions at https://plnkr.co/edit/eHUcO3gdTY86owCOXL18?p=preview .
angularjs 1.5.7, bootstrap 3.3.6, ui-select 0.18.0

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -261,3 +261,19 @@ body > .ui-select-bootstrap.open {
 .ui-select-container[theme="bootstrap"].direction-up .ui-select-dropdown {
     box-shadow: 0 -4px 8px rgba(0, 0, 0, 0.25);
 }
+
+.ui-select-bootstrap .ui-select-match-text {
+    width: 100%;
+    padding-right: 1em;
+}
+.ui-select-bootstrap .ui-select-match-text span {
+    display: inline-block;
+    width: 100%;
+    overflow: hidden;
+}
+.ui-select-bootstrap .ui-select-toggle > a.btn {
+  position: absolute;
+  height: 10px;
+  right: 10px;
+  margin-top: -2px;
+}


user378230:
Can you confirm this works with the clear button? Had to revert a number of bootstrap CSS issues because it broke styling with clear button (eg. #1337)

burka:
It doesn't make it any worse from my point of view. The clear button jumps from above the select to below the select ... :-1:

user378230:
The clear button should be displayed next to the ui-select (like `input-group-addon` in bootstrap) at least it does for me...

burka:
As far as I can see it it is wrong as soon as you restrict the width of the ui-select and have a long text, see https://www.browserstack.com/screenshots/28d5ffe5e0e3742265f046f27d3575f85acc4795 for lots of screenshots of it.

burka:
:-1: Before the fix it worked on short texts, after it doesn't.

user378230:
It needs to work with short and long texts... :confused:

burka:
I added a absolute positioning to the remove link taken from the caret besides it. That works for both short and long texts. Is a.btn below .ui-select-toggle sufficient? 
```
.ui-select-bootstrap .ui-select-toggle > a.btn {
  position: absolute;
  height: 10px;
  right: 10px;
  margin-top: -2px;
}
```
See the updated plnker: https://plnkr.co/edit/eHUcO3gdTY86owCOXL18?p=preview

burka:
This fixes #1576 by keeping the remove link at its position. You could add some magic to shorten the text further to not hide below the button, but it improves the situation a lot.

burka:
Any news on this? It's just a small change and I addressed all raised concerns.

aaronroberson:
@burka accepting as an improvement over the existing functionality, thanks!

Issue #1681, opened 2016-06-22
Typeahead selected item missing in drop down list.


hwangTPN:
The issues forum is __NOT__ for support requests. It is for bugs and feature requests only.
Please read https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md and search
existing issues (both open and closed) prior to opening any new issue and ensure you follow the instructions therein.

### Version of Angular : 1.5.5, UI-Select: 0.18.0 (0.16.1 works)

Angular:

UI-Select:

Bootstrap/Select2/Selectize CSS (if applicable):

allanpoppe:
I'm facing this issue too.

static-max:
Same for me, worked with 0.14.8 and stopped working after upgrading to 0.17.1

ehsaanwelcome:
same for me

hwangTPN:
@allanpoppe not yet. I just downgraded to 0.16.1 as my feature is requested in rush.

user378230:
Possible duplicate #1672 ?

Please include a plunkr as requested in the issue template.

harry-kalligeros:
A quick workaround is to make your markup like this:
`<ui-select remove-selected="false" .... >`

user378230:
Assuming resolved as of v0.18.1

Issue #1672, opened 2016-06-20
`removeSelected` removes any pre-selected item from dropdown 


fiznool:
### Bug description:

When linking the `ui-select` component to a model which points to an item in the items array, the preselected item is not displayed in the dropdown list.

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/bMlvGYc0Jfq3PX8NQcpv?p=preview

Notice that although 'Amalie' is preselected, she is not displayed in the dropdown list.

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: 1.5.0

UI-Select: 0.18.0

Bootstrap/Select2/Selectize CSS (if applicable): 3.1.1

fiznool:
It's worth noting that this situation works fine with 0.17.1.

A temporary fix is to add a `remove-selected="false"` attribute to the `<ui-select>` tag.

p10:
I think it's not a bug but feature ;-) https://github.com/angular-ui/ui-select/commit/3ad084f

You can switch it back to false globaly in configuration section:

```
app.config(function(uiSelectConfig) {
  uiSelectConfig.removeSelected = false;
});
```

fiznool:
@p10 I tried this but it didn't work... I think the config block is run after the uiSelectController is created, and so [this code](https://github.com/angular-ui/ui-select/blob/7ab5f175421e8d263e4e6c5eb38a215b4c515c42/dist/select.js#L286) which caches the `removeSelected` flag is still `true`.

user378230:
@dondi any chance you could take a look at this?

dondi:
OK I'll investigate.

dondi:
Quick update: I have been able to replicate this behavior and have written a unit test for it. Based on what I can replicate, the previous version (0.17.1) indeed does not remove the item from the dropdown, but the highlighted item when the dropdown is displayed for the first time does not match the pre-selected object either (initial highlighted item is always the first one).

Is that what others saw in the previous release? If this is the behavior that everyone is used to, then even if it is not entirely correct, it will be useful to know that just reverting to this will suffice. However if I’m not seeing the correct prior behavior, or if the preference is to get this working _completely_ right, then that changes the code trajectory a bit.

Please advise—thanks!

fiznool:
From my point of view, I've upgraded from 0.14.1 where the preselected item
was correctly displayed and highlighted, my vote would be to revert to this
expected behaviour. Thanks!

[...]

PR #1690, opened 2016-06-27
fix(removeSelected): fix incorrect removal of preselected item

dondi:
Make sure that `removeSelected` only affects `multiple`s, include unit test, and add comment in demo examples.

Addresses #1672 and possibly #1681


user378230:
Cheers @dondi!

PR #1691, opened 2016-06-27
fix(Selectize): Input box is not hidden when selected data is 0 (as n…

### LINKED ISSUE #1304: [Selectize] Input box is not hidden when selected data is 0 (as number)
When selectize theme is used, if the selected data is `0` (as number) input box is not hidden and the bar is not well shown.
I noticed that the criterion to hide the input box is
```javascript
ng-hide="!$select.searchEnabled || ($select.selected && !$select.open)"
```
It should be: "hide it if search is disabled or hide it if something is selected and the box is closed", but it doesn't work if `$select.selected` is `0` (as `0` is false in javascript).
So i think that is should be something like
```javascript
ng-hide="!$select.searchEnabled || (($select.selected || $select.selected===0) && !$select.open)"
```
I tried only with selectize theme. Didn't check if it is handled with other themes.

Jefiozie:
…umber)

Closes #1304


user378230:
Wouldn't it be better to check `$select.selected !== undefined && $select.selected !== null`?

Jefiozie:
@user378230, i think you are right. Maybe it is even better to use the isEmpty function?

user378230:
Yes, good spot!

Jefiozie:
@user378230  okay will change it later today.

Jefiozie:
@user378230, changed the code as discussed in the comments above.

src/selectize/select.tpl.html
```diff
@@ -8,7 +8,7 @@
            ng-click="$select.toggle($event)"
            placeholder="{{$select.placeholder}}"
            ng-model="$select.search"
-           ng-hide="!$select.searchEnabled || ($select.selected && !$select.open)"
+           ng-hide="!$select.searchEnabled || (($select.selected || !$select.isEmpty()) && !$select.open)"
```
user378230: Can't we can remove `$select.selected` now...?

Jefiozie: Yes we can :smile:

Issue #1695, opened 2016-07-01
allow-clear="true" does not work when tagging and reset-search-input="false" are specified


hughanderson4:
The issues forum is __NOT__ for support requests. It is for bugs and feature requests only.
Please read https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md and search
existing issues (both open and closed) prior to opening any new issue and ensure you follow the instructions therein.

### Bug description:
`allow-clear="true"` does not work when `tagging` and `reset-search-input="false"` are specified

### Link to minimally-working plunker that reproduces the issue:
http://plnkr.co/edit/5Y0TL6CNo09LaQfk8nnR?p=preview
the 1st dropdown is to demonstrate working scenario.  2nd dropdown shows the bug

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: 1.4.9

UI-Select: 0.15.0

Bootstrap/Select2/Selectize CSS (if applicable):

hughanderson4:
I have solution for this.  Pull request is forthcoming...

edlahoz:
Is this issue still in the works?

user378230:
@edlahoz  the linked PR still needs a unit test and rebasing due to conflicts. Feel free to pick it up if you like a submit a new one.

mikemicmike:
Hey, i tried the PR but it still has glitches where the selected values cant be cleared. I found another way to do it which seems to fix the problems with 3 simples lines of code. If anyone is interested let me know I will submit a pull request.

heyunxia:
@mikemicmike I have the same situation .can you share your solution ?

mikemicmike:
Ok so basically in the ui-select source, add the following and everything should work. Let me know if anything seems to not work i'll look into it. 

https://gist.github.com/mikemicmike/ed847517f4f6862d27b6789aedfcc574

Up there is the UI-Select JS file, just look for the ****MODIFIED******* tag to find the 3 lines that were updated

Gl! I dont know exactly how pull requests and whatnot work im sort of new to Github but let me know if i can do anything to help. I was gonna have a workmate help me out but he's gone for a while. So if anyone wants this in a PR just tell me how or use my code to do it.

PR #1696, opened 2016-07-01
allow processing to continue when clearing with tagging

hughanderson4:
Fixes [#1695](https://github.com/angular-ui/ui-select/issues/1695)

### DIFF
--- a/dist/select.js
+++ b/dist/select.js
@@ -632,17 +632,17 @@ uis.controller('uiSelectCtrl',
             if ( ctrl.activeIndex === 0 ) {
               // ctrl.tagging pushes items to ctrl.items, so we only have empty val
               // for `item` if it is a detected duplicate
-              if ( item === undefined ) return;
-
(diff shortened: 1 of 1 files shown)


user378230:
I know its a minor change but is there a test that be added for this?

hughanderson4:
Bug #1695 includes link to Plunker that displays the issue:
http://plnkr.co/edit/5Y0TL6CNo09LaQfk8nnR?p=preview

I forked that Plunker and changed it only to use my patched select.js
Here is same Plunker with updated select.js from my patch branch:
http://plnkr.co/edit/PHpiN9sbse8akQ9tP7aM?p=preview

As you can see the forked Plunker works with my patched code.
Hope that is enough.  Let me know if you want me to write a unit test.

user378230:
@hughanderson4 Yeah, I meant in terms of a unit test. Checked against your plunkr but to avoid regressions it would be good to have a unit test. :smiley:

PR #1704, opened 2016-07-07
fix(uiSelectNoChoice): support Select2 theme

### LINKED ISSUE #1608: no-choice.tpl.html is missing from select2 and selectize
In the latest release v0.17.1, the no-choice.tpl.html is missing from the $templateCache for the themes "Select2" & "selectize". It seems to be available only for bootstrap

Error:
Failed to load template: select2/no-choice.tpl.html

stefannikolic:
Closes #1608


user378230:
Have you tested and confirmed this styling is OK?

stefannikolic:
All test went fine by me and styling is ok.

stefannikolic:
@user378230 Do you maybe know when we can expect new version with selectize and select2 support for uiSelectNoChoice, so it can be installed for example via bower and have everything in angular-ui-select/dist/select.js?

Issue #1214, opened 2015-09-29
Latest release breaks select


karptonite:
I don't have a plunker for it yet, but the latest release has broken select for me. Here is a quick video example:
http://quick.as/r2zvin2lb

karptonite:
I suspect that the bug is related to this commit:
https://github.com/angular-ui/ui-select/commit/f6b6d819dbcd019b2b752c60cd3a947188744106
activeIndex is no longer being set on mouseover, but appears to be being used.

dimirc:
I found the bug, I'll release a patch today with other fix too

dimirc:
I cannot reproduce it, I tried the [multi select demo](http://plnkr.co/edit/juqoNOt1z1Gb349XabQ2?p=preview) but when pressing a choice it works correctly, can you setup a plunker or confirm how to reproduce it?

karptonite:
I'll see if I can get it on a plunkr tomorrow, if I have time.

karptonite:
For now, here is the code for the select that doesn't work:
https://gist.github.com/karptonite/3f72ec9e0f8c0d912d57

dimirc:
Can you confirm if `force-options-down` is a private directive you did or where this comes from?

karptonite:
Force options down is something I did before the position attribute was added (it was equivalent to position="down"). I'm pretty certain it is unrelated to this bug, but I don't have the code in front of me.

realityking:
This seeminly also broke the styling for the bootstrap theme, previously the hover effect was a blue background, now it's gray.

karptonite:
@dimirc Actually, as @realityking says, you can even see in you multiselect demo that the bootstrap styles are broken in the current release. I suspect if you track down that bug, the bug I see in functionality (but have yet reproduced in a plunker) may be fixed.

dimirc:
@realityking @karptonite maybe the confusion here is that with https://github.com/angular-ui/ui-select/pull/1211 we removed the ability to set activeIndex on hover and actually only change the background (with css) on hover to give some visual feedback. To change the activeIndex was causing a useless performance hit on hover since the activeIndex was change (and a full $digest). I see more sense on having the current activeIndex change only by the UP/DOWN keys and only a hover effect when mouse over.

karptonite:
[...]

PR #1713, opened 2016-07-09
fix(uiSelectMultiple): ensure first item can selected

user378230:
Issue caused when using simple string item source and tagging label
set to false, causing confusion in selection code between a new tag
and an existing item.

This refactors the selection logic to simplify the code paths and
hopefully make it less prone to errors.

Possible fix for #1214

### DIFF
--- a/src/uiSelectSingleDirective.js
+++ b/src/uiSelectSingleDirective.js
@@ -64,6 +64,14 @@ uis.directive('uiSelectSingle', ['$timeout','$compile', function($timeout, $comp
         focusser.prop('disabled', true); //Will reactivate it on .close()
       });
 
+      $select.clear = function($event) {
+        $select.selected = undefined;
+        $event.stopPropagation();
+        $timeout(function() {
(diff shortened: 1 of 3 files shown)


src/uiSelectSingleDirective.js
```diff
@@ -64,6 +64,14 @@ uis.directive('uiSelectSingle', ['$timeout','$compile', function($timeout, $comp
         focusser.prop('disabled', true); //Will reactivate it on .close()
       });
 
+      $select.clear = function($event) {
+        $select.selected = undefined;
+        $event.stopPropagation();
+        $timeout(function() {
```
aaronroberson: Can you save this $timeout to a variable so that onDestroy it can be cancelled?

aaronroberson:
This is much more inline with the direction that the project needs to go!

user378230:
Parking this for the time being as I was planning on redoing the tagging stuff a bit more sanely.

PR #1715, opened 2016-07-09
feat(groupBy): Define group by expression in repeat

user378230:
Changes the method for defining group by to simplify the codebase,
improve performance and better support changes going forward.

BREAKING CHANGE:

Removes support for group-by and group-filter attributes. Instead you
should now use an `ng-repeat` *like* syntax: `... group by x | ...`

For example:
```
<ui-select-choices group-by="'country'" group-filter="['USA']"
  repeat="person in ctrl.people">
```
Now becomes:
```
<ui-select-choices
  repeat="person in ctrl.people group by person.country | filter:{name: 'USA'}">
```
Note that in order support filter groups based on an array with more
than one value (eg.`group-filter="['A', 'B'...]"`) you will need to
use the new `uisGroupFilter` filter.

Example:
```
<ui-select-choices
  repeat="person in ctrl.people group by person.country | uisGroupFilter:['USA','Canada']">
```

### DIFF
--- a/docs/assets/demo.js
+++ b/docs/assets/demo.js
@@ -39,7 +39,13 @@ app.filter('propsFilter', function() {
     return out;
   };
 });
+app.filter('reverseOrderFilterFn', function() {
+  return function(items) {
+    if(!angular.isArray(items)) return items;
 
+    return items.slice().reverse();
+  };
+});
 app.controller('DemoCtrl', function ($scope, $http, $timeout, $interval) {
   var vm = this;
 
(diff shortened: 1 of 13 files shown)


aaronroberson:
@user378230 This is exactly the direction I would like the project to go in. I'd like to discourage contributors from trying to add an endless amount of new attributes/options to support every conceivable use-case. Instead, a good use of filters and config objects like you've done here is in the spirit of this project, makes the code base more easily maintainable and configuration and use of the component much, much more sane!

PR #1153, opened 2015-08-19
Added open and close event hooks for directive

RavenHursT:
Addresses a few of the feature requests in https://github.com/angular-ui/ui-select/issues/432

Specifically, directive APIs for onOpen and onClose callbacks.

PR #1723, opened 2016-07-12
feat(events): expose uis-open-close callback

### LINKED ISSUE #432: uiSelect should expose OPEN and other events
uiSelect should expose **open** and other events that select2 has: **opening**, **change**, **close**, **highlight**, **selecting**, **removing**, **removed**, **loaded**, **focus**, and **blur**. I understand several are already available but it would be nice to have them all accessible through a single format, e.g. broadcast/on or via attr/observe.

Den-dp:
In order to push this feature (#1153) I'm suggesting this PR which has a slightly updated and rebased code with a couple of unit tests.
If all this makes sense but some changes are required please let me know.

Closes #432 #1153


Den-dp:
@user378230 hey, Is this feature in general fits in current roadmap or for now entire codebase is in "maintenance only" mode?

user378230:
Could you extract this as a separate directive? (See: [Accessing $select](https://github.com/angular-ui/ui-select/wiki/Accessing-%24select))

Den-dp:
@user378230 Yep, this is a very good idea. But not sure which path to choose:

---
Extract this functionality to *a separate directive* called *uiSelect* (directive stacking) with *two optional attributes on-open and on-close.*

pros: :+1:
* provides nice developer experience because ui-select directive is already in place and putting on-open and/or on-close is on developer
* only one $watch

cons: :-1:
* the abuse of directive stacking

---
Or create *two separate directives: on-open and on-close*

pros: :+1: 
* same as in previous from the usage perspective
* kind of more readable code

cons: :-1:
* introduces two $watchers (one per directive)

user378230:
> Or create two separate directives: on-open and on-close
> 
> pros: 
> 
> same as in previous from the usage perspective
> kind of more readable code
> cons: 
> 
> introduces two $watchers (one per directive)

One directive (Eg. (uis)OpenCloseCallback or (uis)OpenClose?), one callback and pass true/false to callback for `isOpen`

Thirsa:
It works for me! would love to see this merged :+1:

Den-dp:
@Thirsa Thank you for your interest in this PR. I will get it done by this weekend.
Also it is worth noting that API will be slightly changed according to @user378230 comment.

Den-dp:
Ok, here is how it looks like after refactoring.

@user378230 when you got some time please take a look and share your thoughts on this.
So if this "api" looks bad we can work it out.

Thanks

Thirsa:
This works smoothly for me as well :+1:

user378230:
@Den-dp can you add notes [Wiki](https://github.com/angular-ui/ui-select/wiki) prior to release? Thanks!

Den-dp:
@user378230 done. https://github.com/angular-ui/ui-select/wiki/uis-open-close Looks pretty simple, but should be a good starting point.

ceeram:
Might be handy to mention the version in which this was added?

Op 1 aug. 2016 15:56 schreef "Denis Bendrikov" <notifications@github.com>:

Den-dp:
@ceeram yes, it's easy. Not 100% sure what version will be the next though :smile:

Den-dp:
@PaulL1 Hi. Despite on the [changelog](https://github.com/angular-ui/ui-select/blob/v0.19.1/CHANGELOG.md#0181-2016-08-07) this feature has landed in 0.19.2.

I have added a note to the wiki page.

ceeram:
it is in 0.19.0 already : https://github.com/angular-ui/ui-select/commits/v0.19.0

it is incorrectly listed in the changelogs

PhilDore11:
@Den-dp When is 0.19.2 supposed to drop?

Den-dp:
@ceeram 

> it is in 0.19.0 already

Unfortunately this is not quite true for npm. You can try to search for `openClose` in here https://npmcdn.com/ui-select@0.19.0/dist/select.js

The fun part is this directive is available in `src` https://npmcdn.com/ui-select@0.19.0/src/ in the same npm package.

@PhilDore11 0.19.2 is aldready available in npm.

pdore-netfore:
@Den-dp Is that also true for bower?

Den-dp:
@pdore-netfore sorry, but it is not.

It looks like you should wait for the 0.19.2 tag https://github.com/angular-ui/ui-select/tags or maybe file an issue.

PR #1724, opened 2016-07-13
Fix button in dropdown

alaingilbert:
Fix the case when you want a button in a dropdown.

**Expected behavior:** The page should not reload when the button is clicked
**Current behavior:** Page reload when button clicked

Bug introduced here: https://github.com/angular-ui/ui-select/pull/1295

Broken (current master): http://plnkr.co/edit/gkm4ymXQbgemCpUteATg?p=preview
Fixed: http://plnkr.co/edit/AJT0rL2OGQw2vV8mRdkN?p=preview


user378230:
Does it need to be an anchor tag? Could we just use a styled `span`?

alaingilbert:
I tried with a `span` and it actually change the styling a lot.
And he cell is not completely covered by the `cursor: pointer;`.
Here is an example with the `span`: http://plnkr.co/edit/vfYKkdWzxmgb3bPIZGxU?p=preview

user378230:
There are some CSS selectors looking for anchor tag in src/common.css (I can't link to line on mobile) but look for `.ui-select-choices-row>a` can you try with reconfiguring them to spans? Thanks for your efforts so far 😃

alaingilbert:
I have updated the previous plnkr with the updated code html/css:
http://plnkr.co/edit/vfYKkdWzxmgb3bPIZGxU?p=preview

user378230:
Is ng-click attribute still required now? `span` doesn't have a default action we need to prevent (I don't think?)

alaingilbert:
Good point.
I have updated the plnkr (http://plnkr.co/edit/vfYKkdWzxmgb3bPIZGxU?p=preview)

alaingilbert:
Any update ?

Issue #1666, opened 2016-06-16
Placeholder text is clipped to 10px sometimes


longzheng:
### Bug description:
I've noticed the placeholder text in some occurrences if the ui-select directive is toggled hidden/show quickly after another.

Screenshot of clipped placeholder text
![image](https://cloud.githubusercontent.com/assets/484912/16117560/1c49c89a-3416-11e6-864d-069f3f3bec26.png)

I noticed in the inspector that the width of this text is exactly 10px

Looking at the source code, I see a line that sizes the input to exactly 10px.
https://github.com/angular-ui/ui-select/blob/master/src/uiSelectController.js#L478

However the lines around this do update the size of the input to the correct container size.

I believe there is a possible race condition here if the actual container size is set correctly and then the width is overridden to 10px after.

Removing this line fixes the issue for me.

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: 1.5.6

UI-Select: 0.18.0

Bootstrap/Select2/Selectize CSS (if applicable):

luizvidoto:
I have the same problem here, just removing this line fixes the issue.

Tim-Erwin:
Same problem. For me the width is always 10px. Removed the line.

tiffanywei:
Can this be fixed soon? Using ui-select via rails-assets pipeline so I can't simply remove this line for my project.

danieldaugherty:
+1

Removing the line also fixes the issue for me.

Submitted PR [#1726](https://github.com/angular-ui/ui-select/pull/1726)

longzheng:
My original fix actually causes an issue where the ui-select box doesn't get correctly sized after selecting one option.

![image](https://cloud.githubusercontent.com/assets/484912/19295516/b9f816ca-9080-11e6-86c0-ad42438ea15f.png)

I've come up with an alternate fix that is a bit safer. 

In the ``refreshComponent`` method, it already calls the ``sizeSearchInput`` method which is suppose to correct this, however it might be done before a digest cycle (before elements are removed).

So wrapping it in a ``$timeout`` seems to fix it

``
ctrl.refreshComponent = function(){
        //Remove already selected items
        //e.g. When user clicks on a selection, the selected array changes and
          //the dropdown should remove that item
          $timeout(function() {
              if ($select.refreshItems) {
                  $select.refreshItems();
              }
              if ($select.sizeSearchInput) {
                  $select.sizeSearchInput();
              }
          });
      };
``

[...]

PR #1726, opened 2016-07-13
Remove code that sizes search input to 10px

danieldaugherty:
This relates to issue [#1666](https://github.com/angular-ui/ui-select/issues/1666)


voluntadpear:
this should be merged as soon as possible

longzheng:
This has weird spacing changes

danieldaugherty:
@longzheng my text editor is set to remove all trailing whitespace from files on save. I think it's weirder to have trailing whitespace in your files at all, but it's all the same when minified.

aaronroberson:
@danieldaugherty Please remove code style and `dist` file changes. Also, I believe the 10px default is a hack to ensure that the the placeholder `hasLayout`.

Can you provide a plunkr so that this can be tested in multiple browsers to ensure that it doesn't cause a regression?

PR #1439, opened 2016-02-16
fix(uiSelectController): Select by click on non-multiple tagging

soyuka:
When in tagging mode, not multiple (`taggingLabel === false`), selecting
an item by click was instead calling `taggingFunc()`. This fix checks
for this manual selection, whether `ctrl.search` is filled or not and
acts accordingly.

closes #1357

PR #1727, opened 2016-07-13
fix(uiSelectController): Select by click on non-multiple tagging (bis)

### LINKED ISSUE #1496: Can't select elements with click when when using `tagging`
I am having the same issue as https://github.com/angular-ui/ui-select/issues/1357 using 0.16.0 (from bower).

I see that this was fixed in https://github.com/angular-ui/ui-select/pull/1439, but I can't find that commit either in the 0.16.0 tag, nor master. Am I missing something?

kfedorov:
When in tagging mode, not multiple (taggingLabel === false), selecting an item by click was instead calling taggingFunc(). This fix checks
for this manual selection, whether ctrl.search is filled or not and
acts accordingly.
Same changes made in #1439 that were mysteriously lost.

Closes #1496

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -340,7 +340,7 @@ uis.controller('uiSelectCtrl',
   function _isItemDisabled(item) {
     return disabledItems.indexOf(item) > -1;
   }
-  
+
   ctrl.isDisabled = function(itemScope) {
 
     if (!ctrl.open) return;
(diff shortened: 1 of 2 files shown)


kfedorov:
@aaronroberson Could you look at this PR, please? It's my first, so feedback would be appreciated.

src/uiSelectController.js
```diff
@@ -375,7 +375,11 @@ uis.controller('uiSelectCtrl',
       if ( ! ctrl.items && ! ctrl.search && ! ctrl.tagging.isActivated) return;
 
       if (!item || !_isItemDisabled(item)) {
-        if(ctrl.tagging.isActivated) {
+        // if click is made on existing item, prevent from tagging, ctrl.search does not matter
+        var manualSelection = false;
+        if($event && $event.type === 'click' && item)
```
user378230: We have a check down below already...
```js
if ($event && $event.type === 'click') {
   ctrl.clickTriggeredSelect = true;
 }
```
Can you combine this/assignment into one and then just check for item?

kellym:
This PR is the cause of this bug: https://github.com/angular-ui/ui-select/issues/1806.

PR #1001, opened 2015-06-10
Create new tag on blur of searchInput

mattmcardle:
As in [this issue](https://github.com/angular-ui/ui-select/issues/544), I had a requirement in my project for a tag to be created when focus had left the search input. To try enable the creation of tags on blur, add the attribute tag-on-blur="true" to the ui-select directive.

PR #1732, opened 2016-07-19
feat(tagging): enable tag-creation on blur

### LINKED ISSUE #544: Allow creation of a new tag on blur
When using ui-select to allow users to create new tags (for example, a list of email addresses), blurring the input field should automatically add that tag, to cover the case where the user enters a new tag but forgets to hit "enter" before submitting the form.

torstenrudolf:
Implement `tagging-on-blur` option on `ui-select` directive. 
This closes #544 and is a cleanup of PR #1001 that never got merged because of formatting issues.

### DIFF
--- a/docs/assets/demo.js
+++ b/docs/assets/demo.js
@@ -173,6 +173,7 @@ app.controller('DemoCtrl', function ($scope, $http, $timeout, $interval) {
   vm.multipleDemo = {};
   vm.multipleDemo.colors = ['Blue','Red'];
   vm.multipleDemo.colors2 = ['Blue','Red'];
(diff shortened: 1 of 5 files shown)


user378230:
Can you extract this to an external directive? (Like: [`src/uisOpenCloseDirective.js`](https://github.com/angular-ui/ui-select/blob/master/src/uisOpenCloseDirective.js) - also see [Accessing $select](https://github.com/angular-ui/ui-select/wiki/Accessing-%24select) for info)

hegoku:
This change hasn't been merged to v0.19.4, right?

PR #1733, opened 2016-07-20
fix(uiSelectSingle): Tagging without multiple with new tags doesn't works

### LINKED ISSUE #1409: Selecting an item not in the list
how can i select (or add) an item that is not in the list? I.e. when a user starts typing and filters to the point where there isnt a match, an option to create a new entry appears.

<ui-select ng-model="order.email" theme="bootstrap">
          <ui-select-match placeholder="Search by email...">{{order.email}}</ui-select-match>
          <ui-select-choices repeat="item in customers | filter: $select.search">
            <div ng-bind-html="item.email | highlight: $select.search"></div>
          </ui-select-choices>
        </ui-select>

skykarthick:
Tagging in single select is not working which has been testing in latest version of angular. Here, I've just copied working version of code from uiSelectMultiple directive and removed some code which causes the problem.

All the test has been passed. Examples are working fine

Closes #1409, #890


wangzhenyu1260:
The [tagging](http://angular-ui.github.io/ui-select/demo-tagging.html) example, Tagging in Single select mode with new tags doesn't work. You can fix it?

skykarthick:
Yes, this has been fixed in my [commit ](https://github.com/angular-ui/ui-select/pull/1733/commits/79b5e004ec359915ed7afdf475b92821893161d6)and I have sent pull request. 

Moderator have to check it, if everything is ok, he/she can merge it to master.

samlaitworker:
I have tried, But It still does not work when Tagging in Single select mode, with simple strings

user378230:
Closing for now since it doesn't work.

PR #1737, opened 2016-07-26
change tabindex and add custom validation

philippdrebes:
Changed the tabindex behavior, so that it is possible to set the tabindex of the input field. Before it was not possible to tab into the input field and type.

Added a property, that allows to set a custom validation method, which is executed on blur. Works when selecting a value by mouse (click), enter or when tabbing out of the input field. Can be used as a workaround for ng-blur.

### DIFF
--- a/src/bootstrap/match.tpl.html
+++ b/src/bootstrap/match.tpl.html
@@ -1,6 +1,5 @@
 <div class="ui-select-match" ng-hide="$select.open && $select.searchEnabled" ng-disabled="$select.disabled" ng-class="{'btn-default-focus':$select.focus}">
-  <span tabindex="-1"
-      class="btn btn-default form-control ui-select-toggle"
+  <span class="btn btn-default form-control ui-select-toggle"
       aria-label="{{ $select.baseTitle }} activate"
       ng-disabled="$select.disabled" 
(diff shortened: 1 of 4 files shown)


user378230:
Is it necessary to set tabindex on searchInput? It is set on focusInput already. 

Can you provide plunkr to demonstrate where this causes a problem?

You can create a custom directive for on-blur functionality. See (Accessing $select)[https://github.com/angular-ui/ui-select/wiki/Accessing-%24select]

philippdrebes:
> Is it necessary to set tabindex on searchInput? It is set on focusInput already.

I have the requirement, that the user should be able to tab inside a table column. In order to achieve that I set tabindex manually.

This plunkr shows the behavior without my changes: [http://plnkr.co/z8e2sp](http://plnkr.co/z8e2sp)

If you just tab through the fields it works as it should, but as soon as you type in the focused ui-select and tab out, the tabindex is no longer in the correct order.

tabindex behavior

Now the same plunkr with the modified ui-select code: [http://plnkr.co/1UVtIH](http://plnkr.co/1UVtIH)
tabindex behavior



> You can create a custom directive for on-blur functionality. See (Accessing $select)

Ok, I will try that.

PR #1742, opened 2016-07-28
Scroll to selected items - bugfix when using with animate

karas111:


src/uiSelectController.js
```diff
@@ -144,6 +144,11 @@ uis.controller('uiSelectCtrl',
 
         if (ctrl.items.length > 0) {
           ctrl.$animate.on('enter', container[0], animateHandler);
+          $timeout(function () {
```
user378230: Shouldn't this be done in `animateHandler`?

user378230:
I feel like there is some duplication here as we already have:
```js
if (ctrl.$animate && ctrl.$animate.on && ctrl.$animate.enabled(container[0])) {
       ...
      } else {
        $timeout(function () {
          ctrl.focusSearchInput(initSearchValue);
          if(!ctrl.tagging.isActivated && ctrl.items.length > 1) {
            _ensureHighlightVisible();
          }
        });
      }
```

And `animateHandler` already calls `ctrl.focusSearchInput(..)`. The timeout call should be extracted to a helper function which is called appropriately, rather than having it separated out.

sonphnt:
This one fixed the problem with scroll to current selected item. But the scrolling position is not very perfect to GUI. I think it should scroll selected item in the middle between top and bottom of the menu-dropdown.

PR #1743, opened 2016-07-29
fix(removeSelected) Unable to reselect item when removeselected set true

### LINKED ISSUE #1739: Unable to (re)select an item with remove-selected set to false
### Bug description:
```remove-selected=false``` seems to have a bug when you select an item and unselect the same item. You couldn't select again. Item was not disabled but click has not effect.

Steps to reproduce:

1. Go to http://plnkr.co/edit/o0JWRLkoMbMBkFi0crqh (same as documentation)
2. Go to the _Disabling instead of removing selected items_ section
3. Select _Amalie_ and remove _Amalie_ from the list.
4. Try to select _Amalie_ again. Not working.

http://plnkr.co/edit/o0JWRLkoMbMBkFi0crqh

Angular: 1.5.8

UI-Select: 0.18.1

Bootstrap/Select2/Selectize CSS (if applicable): Bootstrap

### PR DESCRIPTION
**Pull request by Jefiozie**
This will fix #1739

### CODE REVIEW

File: src/uiSelectMultipleDirective.js [hunk 1]
```diff
@@ -46,6 +46,7 @@ uis.directive('uiSelectMultiple', ['uiSelectMinErr','$timeout', function(uiSelec
         locals[$select.parserResult.itemName] = removedChoice;
 
         $select.selected.splice(index, 1);
+        $select.isDisabled(removedChoice);
```
user378230: Is this needed? The result isn't actually used anywhere.

src/uiSelectController.js
```diff
@@ -333,7 +333,7 @@ uis.controller('uiSelectCtrl',
     }
 
     if (!isDisabled && disabledItemIndex > -1) {
-      disabledItems.splice(disabledItemIndex, 0);
+      disabledItems.splice(disabledItemIndex, 1);
```
user378230: Oops! Good spot! :blush:

src/uiSelectMultipleDirective.js
Jefiozie: You are correct.. missed that one :wink:

PR #1725, opened 2016-07-13
fix(resetSearchInput) change search box doesn't clear after selecting

Jefiozie:
fix(resetSearchInput) change search box doesn't clear after selecting

Previously when the closeOnSelect was set to false the resetSearchInput was not executed after selecting a item even when resetSearchInput was set to true. 

This fix will always execute the resetSearchInput when set to true even with the closeOnSelect set to false.

Fixes #1484

Issue #1384, opened 2016-01-12
Multi Selection with close-on-select="false"


pom-avaleo:
Entered value remains after it's selected for multiple ui-select with close-on-select="false".
http://plnkr.co/edit/BhaOI2Yb6wZDAPXqTkpF?p=preview

schan93:
@pom-avaleo I am having the exact same issue except I didn't declare close-on-select="false".. any ideas why this might be happening?

pom-avaleo:
@schan93 can you paste your code to Plunker?

wiibaa:
@pom-avaleo I think this is the expected behavior by keeping your input it keeps the the listing filtered and let you quickly select a set of value. For example in your plunker, you can type the letter **u** and quickly select all the color with this letter. Anyway this value is cleared when you click out of the field.

Can you explain further what you want to achieve, but at the moment I agree with this "primary" use case.

user378230:
Setting `reset-search-input="true` works as v0.19.4 and resolves this. Closing.

Issue #1083, opened 2015-07-11
Search Filter does not clear when item is selected


cgeo7:
Even though reset-search-input="true" it does not work in the latest version

MiguelMike:
+1

MiguelMike:
I'd really love to have this working. For now you can use the following as a workaround:

In template add the following (change `ctrl.` to match your situation) to your `ui-select` directive:

```
on-select="ctrl.resetInput($select)"
```

Next, in your controller add:

```
ctrl.resetInput = function($search) {
	$search.search = '';
}
```

Refer to docs for more information on `on-select` attribute.

Sheeryada:
+1 for the issue. 
@MiguelMike thanks for the workaround.

pdoreau:
@MiguelMike it works with a flickering effect as selection is added just before the text is removed
+1 for an internal and clean fix !

wesleycho:
Can you post a Plunker of the problem?

user378230:
Closing due to lack of plunkr and possibly resolved by #1746 anyway.

PR #1015, opened 2015-06-16
reset on selection if needed (fix #777)

sagivf:
A solution for https://github.com/angular-ui/ui-select/issues/777

PR #1746, opened 2016-08-01
Fix(resetSearchInput) change search box doesn't clear after selecting

### LINKED ISSUE #1484: Search box doesn't clear after selecting an option
In a tagging and multiple select input when you select an option the search input doesn't clear even when the reset-search-input is true.

The [documentation ](https://github.com/angular-ui/ui-select/wiki/ui-select) says: 
> reset-search-input : Clears the search box after selecting an option

More info here:
[Plunker](https://embed.plnkr.co/a8CLOUFTnPeYxBAF8JMB/)

Jefiozie:
This PR is instead of #1725.

Previously when the closeOnSelect was set to false the resetSearchInput was not executed after selecting a item even when resetSearchInput was set to true.

This fix will always execute the resetSearchInput when set to true even with the closeOnSelect set to false.

I believe this will fix #1484, #1083 , #1384, #777 and is making sure that #1015 is not needed.
@user378230 Did not test #1083 but as I believe will be fixed.


user378230:
@Jefiozie I noticed that if you click an item it doesn't refocus the search area so you can't type in unless you manually click. 

Repro. Steps:
1. Type text (eg. "Y")
2. Click Yellow
3. Try typing... no work :disappointed: 

http://plnkr.co/edit/Cc9XuVRwWHWpVYpbIKCs?p=preview

*It works if you select with keyboard navigation though*

Jefiozie:
Hi @user378230, I see it is not focusing the control . Should it always refocus on the control after selecting a option?  Or should this only be done if multiple is enabled?


p.s. the plunker you shared used v 0.13.1 btw

user378230:
I didn't check single select use case actually but I think it should only refocus in multiple case. (I don't think people use `close-on-select=false` with single select mode anyway...)

And plunkr is against 0.19.4 I used can to update existing example from someone. 😛

Jefiozie:
Okay I will fix it, when the `multiple `and` 'close-on-select=true'`  are used, the focus will be back at the control when the user has selected a item.

user378230:
Thanks! If you could make plunkr with fixed version we can double check single select scenario too just to be sure 😃

Jefiozie:
I think this will do? [Plunker focus](http://plnkr.co/edit/M3nm0irV5UmZwJyepHOk?p=preview)

PR #1029, opened 2015-06-19
Displaying spinner when refreshing

slawekkolodziej:
This PR adds support for displaying spinner. It appears when refresh method returns a promise and hides once the promise is resolved. Only bootstrap templates are included.

PR #248, opened 2014-10-01
Show select2 spinner when updating dynamic content

michaeldjeffrey:
Select2 has a spinner that replaces the search icon in the input when the options are refreshing.

Helps with expectations when content is still loading and no matches are shown.

Issue #919, opened 2015-05-12
Show "searching..." + loading spinner when using `refresh`


nvquanghuy:
I know that I can use `refresh` to asynchornously fetch data from API server. Is there a way for me to give some visual indication to the user that we're loading the data? Similar to how select2 has a spinner, and a "Searching..." text: http://i.imgur.com/3BsemlD.png

http://select2.github.io/select2/ (please see the Loading Remote Data example)

Or if this is not currently supported, I'm happy to open the code and contribute a patch, I just need to be pointed at a right direction.

Thanks!

moneytree-doug:
This is a nice feature that I would like too. My use case is that my list is just too long and it takes a little time to load.

bmherold:
+1

think2011:
+1

jarool:
+1

ggondim:
++1

sanderaido:
+1

venkatesh13:
+1

rusboy:
+1

sebaas:
+1

Rykie:
+2

thiagoross:
+1

ErikDubbelboer:
I think this is something most people like to have implemented differently. To accommodate for everyones wishes would complicate the code quite a bit (as you can already see with other features). It's not easy for people to currently implement this themselves so I don't think it's good idea to put this into the library.

I think it's better if there would be some good documentation on the various ways people can implement this themselves.

We use `ui-disable-choice="item.id < 0"` and in our 'refresh(val)' function we set `$scope.list = [{id: -1, name:'Searching for ' + val}];`.

We also have an `ng-class="{'loading': loading}"` on the `ui-select` and during the ajax call we set `$scope.loading = true`. Then we use some css to add a loading indicator:
```css
.loading:after {
z-index: 10;
    border: solid rgba(9, 62, 78, .2);
    border-top-color: #189dc8;
    -webkit-border-radius: 50%;
    -moz-border-radius: 50%;
    border-radius: 50%;
    -webkit-animation: spin 1s infinite linear;
    -moz-animation: spin 1s infinite linear;
    animation: spin 1s infinite linear;
    content: "";
    position: absolute;
    left: 50%;
    top: 50%;
    width: 40px;
    height: 40px;
    margin: -20px 0 0 -20px;
    border-width: 5px;
}
```

jihunleekr:
+1

julianosaless:
[...]

PR #1779, opened 2016-08-22
feat(spinner) Added spinner on refresh and select2

Jefiozie:
This PR is based on #248 and #1029, needed it for my own and saw that these PR's where still open. 
If approved 248 can be closed and maybe 1029 also but there are some other commits there.


user378230:
I thought the main purpose of the icon would be for async refresh calls so it should check if refreshAttr returns a promise and waits if so, see #1029. 

Note: there's no tests either 🙁. 

It would be nice to support selectize too if we can, as otherwise this would just mean a new issue gets raised claiming yet another bug in ui-select.

Jefiozie:
You are probably right. Will refactoring it to this and add some tests.

user378230:
For reference #919 will also be closed.

Jefiozie:
@aaronroberson can you have a look at this? :smile:

Jefiozie:
@user378230 did you had have some time to review this?

DaniVarga:
It seems there is a error by the class quotation marks https://github.com/Jefiozie/ui-select/blob/50b6bf9331782f940eb1119a785572513a20152f/src/bootstrap/select.tpl.html#L3

Jefiozie:
@DaniVarga , thank you missed that one.

aaronroberson:
@Jefiozie Can you resolve the merge conflicts and I'll get this merged in right away?

Jefiozie:
@aaronroberson , Everything is fine.. but for some reason I added the other commits. Any thoughts on how to fix it? Don't know what happend exactly.

aaronroberson:
@Jefiozie The conflicts appear to be resolved now and I've gone ahead and merged.

Jefiozie:
@aaronroberson  can you close #1029, #919, #248?

src/uiSelectController.js
```diff
@@ -292,16 +295,20 @@ uis.controller('uiSelectCtrl',
...
       if (_refreshDelayPromise) {
         $timeout.cancel(_refreshDelayPromise);
       }
       _refreshDelayPromise = $timeout(function() {
-        $scope.$eval(refreshAttr);
-      }, ctrl.refreshDelay);
+        var refreshPromise =  $scope.$eval(refreshAttr);
+        if (refreshPromise && angular.isFunction(refreshPromise.then) && !ctrl.refreshing) {
+          ctrl.refreshing = true;
+          refreshPromise.then(function() {
```
bogdanalexe90: Shouldn't you take into account the rejection case?
I think the best fit it will be something like 
```js
refreshPromise.finally(function() {
```

shyamal890:
I think instead of adding spinner it would be great if only a `div` is made visible when `refreshAttr = true`. Developers than can add whatever they want inside that `div`. Will make the library much more extensible.

Jefiozie:
@bogdanalexe90 , you are probably right will have a look at it Thanks for the feedback!.

@shyamal890 , interesting and people can then change the div in html with the style they like?

shyamal890:
@Jefiozie Exactly! people can then change content inside the `div` and its `style` in whatever way they want.

Jefiozie:
@shyamal890 I can change it as describe above but shouldn't there be a out of the box working solution? If I change it to the above behavior everybody needs to make adjustments instead of just using it as is? And if people want to make adjustments isn't it better that they use a custom theme?

shyamal890:
@Jefiozie Hmm, I guess it should be something like this:

Expect for `refreshAttr` there should be a `customRefreshAttr`. When `customRefreshAttr = true` then and then only the code below will take effect otherwise a default loading div is shown
              <div ui-select-refresh>
                        //Where user can put in whatever they want
              </div>

Jefiozie:
I still believe the implementation as it is now is a voor solution. People van use custom theme for custom behaviour/style

user378230:
> I still believe the implementation as it is now is a good solution. People van use custom theme for custom behaviour/style

Agreed.

PR #1783, opened 2016-08-25
fix(): "Add no-choice tag in select2-multiple template"

ldipasquale:
Missing `no-choice` tag in `select2/select-multiple.tpl.html`


user378230:
Hmm what happens if user isn't using `no-choice` but `$select.items.length == 0` I feel like the original condition is there intentionally so it doesn't display empty drop down.

Issue #1753, opened 2016-08-06
When ui-select-no-choice is not provided, the default element is always displayed


mdarveau:
### Bug description:

@khanhto Since https://github.com/angular-ui/ui-select/commit/fcd1eb48e5b4dbe6d0305abcefb6c038901863ab, my UI has alignment problems. I tracked the bug down to some CSS on the ```ui-select-no-choice``` div. Turns out that if the ```ui-select-no-choice``` is not present, the default one (from ```src/bootstrap/select.tpl.html```) is placed in the DOM but *without* the ng-show attribute thus always present. 

Check the example in http://plnkr.co/edit/?p=preview and look at the DOM:
![image](https://cloud.githubusercontent.com/assets/390071/17459652/a160dade-5c0f-11e6-84c3-59084097f1de.png)

The ```ui-select-no-choice``` is present in the DOM and the ng-show is not applied. I guess that something is wrong in ```src/uiSelectDirective.js```.

Angular: 1.5.6

UI-Select: 0.18.1

Bootstrap: 3.3.6

mdarveau:
Workaround: explicitly add a div with ```ui-select-no-choice``` and it will be hidden properly.

khanhto:
That pull request was posted long ago, I really don't remember. From what you described, maybe we could apply ng-show to `select.tpl.html` to fix them problem. Can you confirm and post another pull request?

user378230:
Your plunkr link is not valid...

As it is a styling issue, maybe a screenshot of the broken ui-select also.

user378230:
Assuming user resolved their issue, closing.

mdarveau:
Create PR https://github.com/angular-ui/ui-select/pull/1784

PR #1784, opened 2016-08-26
Handle absence of ui-select-no-choice

mdarveau:
Removed the template's ui-select-no-choice from the default template when not specified by user.

The extra element that was not removed was causing issues as it was always present on the dom. If I do add a ui-select-no-choice element, it is hidded by a ng-show added by the ui-select-no-choice directive. See https://github.com/angular-ui/ui-select/issues/1753#issuecomment-238061191


user378230:
You've yet to provide a plunkr or screenshot of the problem this actually causes though? Ie. What problem is caused by it remaining in the DOM, please provide an example.

user378230:
Closing due to lack of response.

PR #1785, opened 2016-08-26
Fix: issue #877 Typing fast in firefox skips the second character 

piebe:
Changing the focus on elements does not always go fast enough in firefox to reliably catch all keystrokes. This fix changes the order of events, and keeps catching keystrokes after the first until the focus has shifted.

(I messed up the original pull request #896) Also see #877

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -111,8 +111,6 @@ uis.controller('uiSelectCtrl',
     if (!ctrl.disabled  && !ctrl.open) {
       if(!avoidReset) _resetSearchInput();
 
-      $scope.$broadcast('uis:activate');
-
       ctrl.open = true;
 
       ctrl.activeIndex = ctrl.activeIndex >= ctrl.items.length ? 0 : ctrl.activeIndex;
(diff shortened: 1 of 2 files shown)


src/uiSelectController.js
```diff
@@ -163,8 +161,12 @@ uis.controller('uiSelectCtrl',
   };
 
   ctrl.focusSearchInput = function (initSearchValue) {
-    ctrl.search = initSearchValue || ctrl.search;
+    if (initSearchValue) {
+      ctrl.search = initSearchValue.val() || ctrl.search;
```
user378230: I don't agree that `initSearchValue` should be a DOM element, angular controllers should avoid dealing with DOM. Can you look at leaving this as text string?

piebe: Uou are right. I adjusted the code.

Plus3531:
I would like to use the fix in this pull request because it is rather annoying that the second character is skipped while typing fast.
I'm waiting for the fix to be merged with the master. Could you please tell me when this pull request will be merged?

lekzd:
@Plus3531 
I made temporary solution -> https://gist.github.com/lekzd/42c6d7e36e9c9fc21c42c335ef9b7ef9
this directive records all user input before all digest actions ends and forsible writes it to $select.search

aaronroberson:
@piebe Can you add tests and possibly a Plunkr example for quicker evaluation?

piebe:
I love to help but...

I am not sure what you want to see in the plunkr. As described in #877 the problem can be reproduced in the standard demo plunkr using Firefox.

Also I do not know how to create a test to simulate fast typing by a user in Firefox.

Issue #1624, opened 2016-05-26
Focusing different element does not work properly when using append-to-body


pedroclayman:
### Bug description:

1. Focus the select element and interact with it in any way you like (select some options, filter stuff etc.)

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular:

UI-Select:

Bootstrap/Select2/Selectize CSS (if applicable):

PR #1791, opened 2016-09-02
fixed problems with focusing and selection of elements when using app…

pedroclayman:
…endToBody

This fixes the issue #1624.


aaronroberson:
@pedroclayman Can you remove the `dist/` files and add unit tests for the updated `resetDropdown` function?

PR #1793, opened 2016-09-05
fix problems with several example files in docs

bobisch:
fix(demos): Fix problems with several example files in docs

Make demos actually work in many cases.  Mostly add missing
controller references and incorrect model names.

   - demo-append-to-body.html - expand width of grid
   - demo-basic.html - expand width of grid
   - demo-group-filter.html - add missing controller reference
   - demo-multiple-selection.html - add many (8) missing controller references
   - demo-object-as-source.html - ~~fix search filters~~, removed incorrect search highlights
   - demo-select2-with-bootstrap.html - fix incorrect model names and group-by fields.
   - demo-selectize-with-bootstraph.html - fix group-by field (use 'country' since 'group' does not exist).
   - demo-tagging.html - fix incorrect model name (selectedPeople*2*).  Display `tagging-tokens=...` example.

Some possibly gratuitous changes:

   - demo-tagging.html - add "(NOT WORKING)" to the two "Tagging in single object mode"
     examples.
   - .gitignore - addition of "*~" to ignore vim backup files.
   - [vim] removed some trailing spaces.
   - [vim] removed BOM from beginnings of some files.
   - [vim] added a line ending to last lines if necessary.
   - gulpfile.js - added docs:watch task.


bobisch:
Looks like there was some overlap with some changes @Jefiozie submitted recently.  Apparently I rebased and got some of his fixes that matched identically but there are several additional ones in here.  Also, we differed on the filter fixes in demo-object-as-source.html and I added some filter highlighting.

I use vim and it removed some BOMs and trailing spaces.  Let me know if that is a big problem and what the convention is if it needs to be changed.

It would be nice if we could get these working - it's bad when the demos have been broken for apparently a long time.  There are still two that I do not think work - the "Tagging in single object mode" ones if anyone has any ideas on those we could fix it here or do another PR?

docs/examples/demo-object-as-source.html
```diff
@@ -24,7 +24,7 @@
   <p>Selected: <code>{{ctrl.person.selectedSingle}}</code></p>
   <ui-select ng-model="ctrl.person.selectedSingle" theme="select2" ng-disabled="ctrl.disabled" style="min-width: 300px;" title="Choose a person">
     <ui-select-match placeholder="Select a person in the list or search his name/age...">{{$select.selected.value.name}}</ui-select-match>
-    <ui-select-choices repeat="person.value.name as (key, person) in ctrl.peopleObj | filter: { value : { name: $select.search }}">
+    <ui-select-choices repeat="person.value.name as (key, person) in ctrl.peopleObj | filter: $select.search">
```
user378230: `filter: { value : { name: $select.search }}` was used here as I believe `filter: $select.search` does not work correctly. (It might angular v1.5+ only).

docs/examples/demo-selectize-with-bootstrap.html
```diff
@@ -45,7 +45,7 @@
 
           <ui-select ng-model="ctrl.person.selected" theme="selectize" title="Choose a person">
             <ui-select-match placeholder="Select or search a person in the list...">{{$select.selected.name}}</ui-select-match>
-            <ui-select-choices group-by="'group'" repeat="item in ctrl.people | filter: $select.search">
+            <ui-select-choices group-by="'age'" repeat="item in ctrl.people | filter: $select.search">
```
user378230: Not sure why this is changed. `group` is a property on the person object IIRC.

docs/examples/demo-select2-with-bootstrap.html
```diff
@@ -40,7 +40,7 @@
 
           <ui-select ng-model="ctrl.person.selected" theme="select2" class="form-control" title="Choose a person">
             <ui-select-match placeholder="Select or search a person in the list...">{{$select.selected.name}}</ui-select-match>
-            <ui-select-choices group-by="'group'" repeat="item in ctrl.people | filter: $select.search">
+            <ui-select-choices group-by="'age'" repeat="item in ctrl.people | filter: $select.search">
```
user378230: Not sure why this is changed. `group` is a property on the person object IIRC.

user378230:
@bobisch thanks for fixing some of them. I think some of the changes need to be reverted though (the `filter` and `group-by` changes).

docs/examples/demo-select2-with-bootstrap.html
bobisch: @user378230 - No, there is no `group` property so I changed it to `age` to fix that.  There is a `country` property on these I believe.  Would it be better to use that?  `people` seem to have `name`, `email`, `age`, and `country`.

docs/examples/demo-object-as-source.html
bobisch: @user378230 - Yes, actually this WAS `filter: {'value': $select.search}`, which was not working.  @Jefiozie corrected it to the value you cite, which is apparently a better fix than the simple `$select.search` that I choose.  I will revert this one and look over the highlighting.

docs/examples/demo-select2-with-bootstrap.html
user378230: @bobisch My bad, I not sure where I got that from. I think `country` might be better in this case though. Thanks!

bobisch: @user378230 - sounds good.  It is changed to country now - 2454d59

bobisch:
Was traveling for a while but wanted to finish this off if possible now.  I merged and rebased with upstream master and I think the PR is up to date.  Let me know what the best way to clean this up for merging.  Thanks.

aaronroberson:
@bobisch Can you separate out the feature added (src/selectize/choices.tpl.html, src/selectize/match-multiple.tpl.html, src/selectize/select-multiple.tpl.html) into a different PR?

user378230:
> Can you separate out the feature added (src/selectize/choices.tpl.html, src/selectize/match-multiple.tpl.html, src/selectize/select-multiple.tpl.html) into a different PR?

Can't see any of these now, so I'll merge this. Thanks @bobisch! 

Now I just need to fix travis to build the docs again :stuck_out_tongue_closed_eyes:

PR #1822, opened 2016-10-10
Bug fix - Search input isn't blocked 

nabilnaffar:
Scenario:
- multiple select
- search-enabled is set to false
- all items are selected
- user type something in search input

expected results:
input keys should be blocked, input search should stay clean

actual results:
keyboard keys are shown in input search


aaronroberson:
@nabilnaffar Can you please remove the `dist/` files from the PR?

nabilnaffar:
@aaronroberson sure!

aaronroberson:
Thanks @nabilnaffar!

chadfurman:
I feel like this was a bad PR @aaronroberson -- When I type something into the search box and it returns no results from my API call, I cannot hit the backspace button.  In fact, I can no longer hit _any_ button while I have the input focused.

Jefiozie:
Added PR for reverting this maybe @nabilnaffar  can have a look at the comments and make a new PR.

nabilnaffar:
@chadfurman can you specify the scenario that doesn't work + browser/OS? I'll try to reproduce it

ohjames:
Is this the PR that broke release 0.19.5? I think this version should be pulled from bower etc. it is unusable for the reasons @chadfurman lists.

lundens:
With the current dist, I get this behaviour:
http://plnkr.co/edit/yuyks6ddYFWw1gUhnmEP?p=preview

aaronroberson:
Merged and published as 0.19.6. Unfortunately, NPM won't let me unpublish 0.19.5.

rthrfrd:
For us 0.19.5 completely broke any multiple selects that have no options preset and rely on `refresh` to fetch the options after the user begins typing, as the user is unable to type anything at all.

user378230:
The tag was missing from the `0.19.6` release for this revert so I've just pushed one to ensure bower is updated.

PR #1823, opened 2016-10-11
Skip disabled items while navigating with keyboard

Seprum:
When user presses up/down keys within dropdown, disabled items become active too. It has no visual effect but still is a bit confusing. I have added check for disabled items insted of plain incrementing/decrementing `activeIndex` value.


src/uiSelectController.js
```diff
@@ -569,11 +569,25 @@ uis.controller('uiSelectCtrl',
...
+          }
+        }
         break;
       case KEY.UP:
         if (!ctrl.open && ctrl.multiple) ctrl.activate(false, true); //In case its the search input in 'multiple' mode
-        else if (ctrl.activeIndex > 0 || (ctrl.search.length === 0 && ctrl.tagging.isActivated && ctrl.activeIndex > -1)) { ctrl.activeIndex--; }
+        else if (ctrl.activeIndex > 0 || (ctrl.search.length === 0 && ctrl.tagging.isActivated && ctrl.activeIndex > -1)) {
+          var newIndex = ctrl.activeIndex - 1;
+          while (_isItemDisabled(ctrl.items[newIndex]) && newIndex >= 0)
+            newIndex--;
```
user378230: Similar to the other comment, to save on duplication of checks

user378230:
@Seprum this looks very useful! But it does need unit tests before it can be merged. :disappointed_relieved: 

Any chance you could add some? :smiley:

user378230:
@Jefiozie you might like to pick this one up with tests etc. if you like it. I think it seems like a reasonable change.

Jefiozie:
@user378230 , you want me to implement this fix?

user378230:
@Jefiozie Only if you want to! It looks fairly easy and makes sense, so thought you might like it :stuck_out_tongue_closed_eyes:

Jefiozie:
@user378230 , see my PR 👍

Jefiozie:
Closing pr as handeld by other PR

PR #1824, opened 2016-10-11
fix(_updateItemLocked) properly unlock locked items when lock conditions change

jasonthomasgray:
Previously when the lock expression value changes from true to false the locked value was not removed from the locked list.


user378230:
Should probably be a test that caught this but we'll just quietly slide this in while no-ones looking...

Thanks :smile:

Issue #1820, opened 2016-10-09
Multiple Tagging with "Array of object" is allowing insertion of duplicate entries.


Rahul-Sagore:
### Bug description:
I created tagging feature, where user can select from existing stored tags or create new if not existed. So if I have one tag "Algorithm" in dropdown, and user type "algo" then it allows creating new one. But if user type "algorithm" or "Algorithm", then both existing tag and "create new" displays. So it should allow insertion of new tag only if it is not already existing tag.
<img width="440" alt="screen shot 2016-09-25 at 12 59 18 pm" src="https://cloud.githubusercontent.com/assets/3587851/19218419/223ca6e0-8e17-11e6-8e1c-ed0e0a7c9b63.png">

This is the array, which populates the ui-select:
`
[ {uid: "0x943835545", name: "Algorithm"}]
`

Probably it should allow to filter on one property of object, for example name of the tag and ignore others. And if tag name exist then "Create new" label should disappear or disabled.

### Link to minimally-working plunker that reproduces the issue:
This is happening on official [Tagging example](http://angular-ui.github.io/ui-select/demo-tagging.html) as well. This is someone else's [plunker link](http://embed.plnkr.co/szYtoGgyFZVXAx12Y13c/). Check array of object example.

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: AngularJS v1.3.13

UI-Select: 0.19.3

davidetwc:
Hi, I have th same issue...

Jefiozie:
Are you using a custom tagging function? Or the default?

Rahul-Sagore:
Custom tagging means user can add new tags on the fly, right? This is the code I wrote, where addNewTag function return user typed string. 

```
<ui-select multiple tagging="questionVm.addNewTag" theme="select2" ng-model="addQueVm.newQuestion.tags" class="mdl-select">
    <ui-select-match  placeholder="Enter tags">
        <span ng-bind="$item.name"></span>
    </ui-select-match>
    <ui-select-choices repeat="tag in (questionVm.allTags | filter: $select.search)">
        <span ng-bind="tag.name"></span>
    </ui-select-choices>
</ui-select>
```

I have handled this problem on frontend by calling custom function on "on-select" event, which handles unique tag insertion. But I wanted to know if there is something in-built in ui-select to make it unique based on single object property.

Jefiozie:
[...]

PR #1830, opened 2016-10-20
Issue 1820

zajacmp3:
In response to issue raised: https://github.com/angular-ui/ui-select/issues/1820

I have added a new html tag called object-unique-key.
This tag allows us to define based on what we define if an object is or is not unique.

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -46,6 +46,7 @@ uis.controller('uiSelectCtrl',
   ctrl.clickTriggeredSelect = false;
   ctrl.$filter = $filter;
   ctrl.$element = $element;
+  ctrl.objectUniqueKey = undefined;
 
   // Use $injector to check for $animate and store a reference to it
   ctrl.$animate = (function () {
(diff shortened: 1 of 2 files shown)


user378230:
Shouldn't we actually just be using the value specified in `track by`?

It appears we are duplicating functionality with this.

zajacmp3:
I'll investigate that and get back to you with an answer by the end of the weekend.

zajacmp3:
@user378230 

I failed to reproduce my solution with ```track by``` so personally I would opt to say that we are not duplicating functionality. While doing code analysis when using multiple select and tagging there is no other logic that compare objects. I modified the only one in a generic way.

During my investigation I found out that:
```track by``` is used in directive used to show choices/match items **and to support ng-repeat directive**
```<ui-select-choices repeat="address in ctrl.addresses track by $index"```
http://angular-ui.github.io/ui-select/demo-basic.html
``` 
* Original discussion about parsing "repeat" attribute instead of fully relying on ng-repeat:
* https://github.com/angular-ui/ui-select/commit/5dd63ad#commitcomment-5504697
```
uisRepeatParserService.js:7-8
```
  self.getGroupNgRepeatExpression = function() {
    return '$group in $select.groups track by $group.name';
  };
```

So ```track by``` does not seems to do the trick at the minute.

But maybe you wanted to ask could we re use ```track by``` argument in ui select directive as it seems that we might not want to add yet another argument to <ui-select> directive?

Issue #1831, opened 2016-10-20
Does not work with custom start/endSymbol defined over $interpolateProvider


HarpoMarx69:
The issues forum is __NOT__ for support requests. It is for bugs and feature requests only.
Please read https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md and search
existing issues (both open and closed) prior to opening any new issue and ensure you follow the instructions therein.

### Bug description:
Because of combinating with Symfony/Twig I use 
$interpolateProvider.startSymbol('[[').endSymbol(']]');

Shouldn't then line 68 in uiSelectSingleDirective.js better be...

`var focusser = angular.element("<input ng-disabled='$select.disabled' class='ui-select-focusser ui-select-offscreen' type='text' id='" + $interpolateProvider.startSymbol() + " $select.focusserId " + $interpolateProvider.endSymbol() + "' aria-label='" + $interpolateProvider.startSymbol() + " $select.focusserTitle " + $interpolateProvider.endSymbol() + "' aria-haspopup='true' role='button' />");`

...to work correctly?

And therefore line 1:

`uis.directive('uiSelectSingle', ['$timeout','$compile','$interpolateProvider', function($timeout, $compile, $interpolateProvider) {`

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: 1.5.8

UI-Select: 0.19.5

renatomarteli:
# **HarpoMarx69**  do you make an exemplo using a match multiple searching data itens in back end?
I hava any problems with this version ui-select , using Angular 1.5.8

user378230:
@HarpoMarx69 this seems reasonable, it sounds like you already know how to fix. Feel free to open a PR with the change and a test. Would be happy to merge. :smile:

PR #1834, opened 2016-10-26
Fixed issue #1831

HarpoMarx69:
Added featuring start-/endSymbol defined in $interpolateProvider

### DIFF
--- a/src/uiSelectSingleDirective.js
+++ b/src/uiSelectSingleDirective.js
@@ -1,4 +1,4 @@
-uis.directive('uiSelectSingle', ['$timeout','$compile', function($timeout, $compile) {
+uis.directive('uiSelectSingle', ['$timeout','$compile','$interpolateProvider', function($timeout, $compile, $interpolateProvider) {
   return {
     restrict: 'EA',
     require: ['^uiSelect', '^ngModel'],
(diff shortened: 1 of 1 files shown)


user378230:
Thanks @HarpoMarx69 but we could do with a unit test to make sure we don't get this popping up as a regression in future. 

Wonder if we could tidy up a little like so:
```html
var focusser = angular.element("<input ng-disabled='$select.disabled' class='ui-select-focusser ui-select-offscreen' type='text'  aria-haspopup='true' role='button' />"); 
focusser.attr("id", $interpolate.startSymbol() + " $select.focusserId " + $interpolate.endSymbol());
focusser.attr("aria-label", $interpolate.startSymbol() + " $select.focusserTitle " + $interpolate.endSymbol());

$compile(...)
```
*(**Disclaimer**: I don't know if that will actually work?! :smile:)*

Jefiozie:
Hi @HarpoMarx69 , any luck on the test part so we can merge this?

HarpoMarx69:
Sorry, unit tests on this are overchallenging me at the moment...

dant89:
Please can you merge this to master? Very frustrating bug...

Issue #1794, opened 2016-09-06
onSelect callback is executed when limit is reached in multiple mode


orestisioakeimidis:
### Bug description:

The `onSelect` callback is executed when `limit` is reached for multiple select. That shouldn't be the case because nothing is actually selected.
### Link to minimally-working plunker that reproduces the issue:

[Example](https://plnkr.co/edit/cA1HjRkYEzSmXX4P5Dcx?p=preview)
### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: 1.5.8

UI-Select: 0.19.4

Jefiozie:
What do you believe is the best way to fix this. Just a check on the limit but then the items in the list are still clickable? Or disable all the items?

orestisioakeimidis:
I think that all items should also be disabled. You have reached the limit and you are not allowed to select any more, so it makes sense to disabled them.

orestisioakeimidis:
What is the status of this case?

Jefiozie:
Sorry didn't have a lot of  time. I agree on the disabled state when you are on the defined limit.  Maybe you can make a PR? Happy to review?

PR #1836, opened 2016-10-27
fix(uiSelectMultiple): Don't call onSelectCallback if limit already reached

Jefiozie:
This PR will fix the onSelect problem that was described by #1794 .


user378230:
Thanks @Jefiozie  

Could we instead move the call to `ctrl.onSelectCallback(...)` be moved to the `$on('uis:select')` handlers in `uiSelectSingleDirective` & `uiSelectMultipleDirective`? 

It seems a little convoluted to have the intermediary `ctrl.limitReached` boolean floating about.

Jefiozie:
@user378230, Agree will make some changes! Thanks for the feedback.

user378230:
@Jefiozie Would you mind also checking for case when `limit` is not specified? Not easy to tell from code what would happen. 

Thanks! :smiley:

Jefiozie:
@user378230 , sure no problem.

Jefiozie:
@user378230 Have removed the selectCallBack to the directives and added a test for selectcallback without limit.

user378230:
@Jefiozie I think you messed up your commits here. The diff is showing changes other than yours.

You can force push to your existing limitOnCallback branch once you've fixed, rather than creating a new PR.

src/uiSelectSingleDirective.js
```diff
@@ -51,6 +51,15 @@ uis.directive('uiSelectSingle', ['$timeout','$compile', function($timeout, $comp
 
       scope.$on('uis:select', function (event, item) {
         $select.selected = item;
+        var locals = {};        
+        locals[$select.parserResult.itemName] = item;
+
+        $timeout(function(){
+          $select.onSelectCallback(scope, {
```
user378230: Is `scope` the correct scope here? Is there a `$select.scope` it should be instead?

test/select.spec.js
```diff
@@ -1831,7 +1833,7 @@ describe('ui-select tests', function() {
             </ui-select>'
         );
     }
-
+
```
user378230: Check for trailing whitespace in your commits too :+1:

src/uiSelectSingleDirective.js
Jefiozie: I did some checking with the unit test if this is the case but then it will fail. Why do you think it should be $select.scope?

user378230: My only thought was that it wasn't the same scope as the one in uiSelectController but I guess we don't use a new scope anyway for the uiSelectSingle directive so it should be ok...

test/select.spec.js
```diff
@@ -2764,6 +2766,49 @@ describe('ui-select tests', function() {
...
+        expect(scope.$item).toBeFalsy();
+        expect(scope.$model).toBeFalsy();
+
+        clickItem(el, 'Samantha');
+        $timeout.flush();
+        clickItem(el, 'Natasha');
+        $timeout.flush();
+        expect(el.scope().$select.selected.length).toEqual(2);
+    });
```
user378230: Need to check the value for `scope.$model.name`?

test/select.spec.js
```diff
@@ -2764,6 +2766,49 @@ describe('ui-select tests', function() {
...
+         clickItem(el, 'Natasha');
+         expect(el.scope().$select.selected.length).toEqual(1);
+    });
+
+    it('should only have 1 item selected and onSelect function should only be handled once.',function(){
+        scope.onSelectFn = function ($item, $model) {
+          scope.$item = $item;
+          scope.$model = $model;
+        };
+        var el = createUiSelectMultiple({limit:1,onSelect:'onSelectFn($item, $model)'});
```
user378230: minor tidy up of your spacing here.. :smiley:

user378230:
Just some tiny ones, then it LGTM :+1:

Jefiozie:
Fixing it now 😄

PR #1843, opened 2016-11-01
Load CDN assets over HTTPS

softwareresearchwork:
When accessing https://angular-ui.github.io/ui-select/demo-tagging.html, the browser blocks insecure (http://) content.

### DIFF
--- a/docs/partials/_header.html
+++ b/docs/partials/_header.html
@@ -5,12 +5,12 @@
     <title>AngularJS ui-select</title>
 
     <!--
-      IE8 support, see AngularJS Internet Explorer Compatibility http://docs.angularjs.org/guide/ie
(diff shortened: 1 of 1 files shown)


user378230:
Thanks but can you refer to [CONTRIBUTING.md](https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md#commit-message-format) for guidelines regarding commit messages. 

Could we remove the protocol entirely? It looks like that's what they do over on [angular-ui/bootstrap](https://github.com/angular-ui/bootstrap/blob/master/misc/demo/index.html#L11)

softwareresearchwork:
Amended the commit message.

Regarding the protocol in the URL, use of protocol-relative URLs is now considered an anti-pattern. (Refer https://www.paulirish.com/2010/the-protocol-relative-url/)

PR #625, opened 2015-02-02
refresh on active

abudel:
From #271, 
By setting to true the attribute "refresh-on-active" in the "ui-select-choiche", refresh method is called when the component is activated the first time.

Issue #619, opened 2015-01-29
Need ability to postpone refresh function call until select is open


Envek:
### Description

I have a lot (tens or thousands) selects on page that need to fetch choices for each select independently.

Each select allows to choose a user with a some rank that is free in some period of time. As only server knows that, I need to fetch data from it. But there is a **lot** of selects and data for all of them starting to load immediately after page load making browser to freeze. But at that time all of selects are hidden with `ng-show` directive of some parent node and most probably most of them wouldn't be used now.

Here is the code:

``` html
<div ng-show="crew.editing">
  <div ng-repeat="member in crew.members">
    <ui-select ng-model="member.user" theme="bootstrap">
      <ui-select-match>{{$select.selected.shortName()}}</ui-select-match>
      <ui-select-choices 
        repeat="user in users[member.slot.rank_id] | filter: {name: $select.search, lastname: $select.search, patronymic: $select.search}"
        refresh="loadUsers(crew.time_interval, member.slot.rank)"
      >
        <div ng-bind-html="user | highlight: $select.search"></div>
        <small>{{user.ranks | map:'name' | join:', '}}</div>
      </ui-select-choices>
    </ui-select>
  </div>
</div>
```
### Required behaviour

Need ability to postpone function call in `refresh` attribute to time when select will be opened/clicked/focused/hovered/anything.
### Workaround

Hide blocks containing selects with `ng-if` directive. It allows to load users in somewhat lesser batches, but problem still exists.

sporkd:
You could probably define your own fetch method on the current scope and have it return empty options until you need the data. See https://github.com/angular-ui/ui-select/wiki/ui-select#examples-async

Trick will be detecting the event you want to trigger loading the options...

```html
<ui-select ng-model="model.selected" ng-click="loadOptions()">
  <ui-select-choices repeat="option in options"
           refresh="refresh($select.search)">
</ui-select>
```

```coffeescript
# Inside your directive
link: (scope, elem, attrs) ->
  scope.enabled = false
  scope.refresh = (query='') ->
    if scope.enabled
      OptionsService.fetch(query).then (data) ->
        scope.options = data
    else
      scope.options = []

  scope.loadOptions = ->
    scope.enabled = true
    scope.refresh()
```

Still, I agree it would be nice to delay the first call to fetch since it happens immediatley when the directive is rendered in a template. Another workaround is to delay the rendering  the actual template it lives in using resolving routes...

```coffeescript
.when '/selection-page',
[...]

Issue #271, opened 2014-10-08
Wait for the first open to first load remote data


diogodomanski:
Currently, if <ui-select-choices> has "refresh" attribute, its expression is performed as soon as the directive is instanciated.

It would be useful if there was a parameter to inform <ui-select-choices> to perform the "refresh" expression only when the <ui-select> gets focus (and dropdown is opened) or the user changes input value (and taking into account the "refresh-delay" attribute).

dimirc:
Related to PR https://github.com/angular-ui/ui-select/pull/557

Jefiozie:
As the PR is merged please verify if it fix your problem. For now I will close the issue if needed you can reopen.

DmitryGonchar:
@Jefiozie this issue is to have an option to "**first** load remote data on first open of dropdown", not on page init. According to the official demo behavior and the code in the PR, this use case is still not supported. So probably makes sense to leave this issue open

Jefiozie:
@DmitryGonchar  i think this is already possible. When using the `refresh="fetchFromServer($select.search, $select.open)"` like this you already can verify if the ui-select is open or not. If it is you can load the data you want. Correct me if i'm wrong see also the dicussion here: [1845](https://github.com/angular-ui/ui-select/pull/1845#issuecomment-258615839)

DmitryGonchar:
Thanks. I think this was not mentioned in the docs

Issue #1174, opened 2015-09-01
Lazy options loading!


erobwen:
Hi, this is more of an open question and a feature request, rather than an "issue report". 

What I would like is a way to lazily load all of the select options the first time when the user opens up the menu. Initially there is only one single preselected item, or a text in the select-box, and when the user clicks the dropdown open, a spinner appears as options are loaded. 

I have searched the web for solutions for this, but found non satisfactory. I have thought of various ways to do this, but I understand it might even be tricky to listen to the open event from the ui-select. (there are some far fetched solutions for this I found). And If I am to populate the "unloaed" option list with a single non-selectable spinner option, I dont know how to swap it for the real option list once the loading is complete. My attempts at  combining ng-if or ng-show with ui-select-choices did not turn out well. 

So I am pretty much stuck here. 

Does anyone have a workaround for this. But perhaps even better, if ui-select could support this out of the box, it would be even better.

vnod-ideas2it:
+1

nmeln:
+1

Untit1ed:
+1

zamabraga:
+1

natwallbank:
+1.  Has anyone found any useful workarounds in the meantime?  I've just had a requirement for this where we have lots of filters (ui-select elements) on a single page but very few will be used per request.  It's wasteful to load the values for these filters until they're actually used.

ErikDubbelboer:
It's very easy to do this yourself using the `refresh` function. It's important to set `refresh-delay` to 0 so the function is called immediately when the users clicks the menu.

```html
<ui-select-choices refresh="refresh" refresh-delay="0" ...>
```
```javascript
...

  var loaded = false;
  $scope.refresh = function (search) {
    if (!loaded) {
      // Do lazy loading here...
      loaded = true;
    }
  };

...
```

natwallbank:
Isn't refresh called as the control renders, with refresh-delay simply affecting how quickly? My testing seemed to confirm that..

ErikDubbelboer:
My mistake, you're right. I have removed my answer not to confuse people skimming thought the issue.

mkrijtenberg:
+1

sandro1111:
+1

mjeson:
+1

Poltuu:
That would be much appreciated indeed !

HDaghash:
+1

radyz:
[...]

PR #1845, opened 2016-11-03
fix(#271) if option is set the refresh function is getting the request.

Jefiozie:
I think this is something that is requested multiple times: #271, #625,#619.

Used the initial changes from 625 with some minor adjustments to let everything work and only watch variables if needed.

If this is a viable solution i think all the other can be closed down.


user378230:
Could we instead pass `ctrl.open` as an argument to the refresh function?

Jefiozie:
Instead of the refreshIsActive ? Will run some tests with it.

user378230:
Ah we actually have two "issues" here:
1. refresh is being called when ui-select is initialized (but not opened)
2. refresh is not called when select is activated

1. Can be solved by passing ctrl.open to the refresh function (inside the refresh function users can say if select is not open then don't call server)
2. Can be solved by calling in `$select.open` `$watch` like you did. 

For **2** we don't need a new attribute, just call refresh all every time, as we call refresh whenever `$select.search` changes anyway.

Jefiozie:
But if we look if `ctrl.open` is true we could fix also point 1 right? If you search but it is not active (open will be set to true in the function active).  This will prevent the refresh function to go off on initializing and only execute the function when it is active. Right?

Jefiozie:
Did some test to verify if this solution is working.....and it will break a lot of tests.  Need to dig a little bit deeper because in the test the function `openDropDown` will set the `ctrl.open` manually to true. I thought it should be `$select.activate()` but still problems. Ideas are welcome 😉

user378230:
@Jefiozie I think you misunderstand, we do **not** need to do anything with `ctrl.open` because user can already do `refresh="fetchFromServer($select.search, $select.open)"` right? 

> User can check manually in `fetchFromServer(...)` if select is open. 

 *I cannot test this at the moment - but I think it should be correct*

---
The only thing **extra** we need to do it, is call `ctrl.refresh` in `activate` (so that refresh is called whenever select is opened).

user378230:
Tagging #1174 for reference.

Jefiozie:
@user378230 , understand you now and agree on both points. Will remove the extra parameters/ variables and add the `ctrl.refresh` in  `activate` function

Jefiozie:
@user378230 ,
So making these changes have a lot of impact. This peace of code is executed on initializing [uiSelectChoicesDirective.js#L68](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectChoicesDirective.js#L68). Adding a check on this if the `$select.open` equals true (as this is set by the activate function) it will break all tests but has the behavior as it should ( I believe).

**Next problem**: 
In a use case where the ui-select doesn't use the refresh functionality it will not set the items until it has a search value.  Trying to find out why this is happening, if you don't use the refresh it should just set the items. But it looks like the have a relation somewhere. 

Keep you posted

src/uiSelectChoicesDirective.js
```diff
@@ -68,7 +67,7 @@ uis.directive('uiSelectChoices',
         scope.$watch('$select.search', function(newValue) {
           if(newValue && !$select.open && $select.multiple) $select.activate(false, true);
           $select.activeIndex = $select.tagging.isActivated ? -1 : 0;
-          if (!attrs.minimumInputLength || $select.search.length >= attrs.minimumInputLength) {
+          if ((!attrs.minimumInputLength || $select.search.length >= attrs.minimumInputLength) && (!$select.refreshOnActive || ($select.refreshOnActive && $select.open))) {
```
user378230: We can undo changes here also. It will be up to user to check in their own function.

src/uiSelectChoicesDirective.js
```diff
@@ -80,10 +79,14 @@ uis.directive('uiSelectChoices',
           var refreshDelay = scope.$eval(attrs.refreshDelay);
           $select.refreshDelay = refreshDelay !== undefined ? refreshDelay : uiSelectConfig.refreshDelay;
         });
-
+        
         scope.$watch('$select.open', function(open) {
           if (open) {
             tElement.attr('role', 'listbox');
+            if(!angular.isUndefined($select.refreshOnActive)){
```
user378230: I wonder instead if we do this in a `uis:activate` listener?

Jefiozie:
@user378230  I removed this already only I didn't commit it. Can commit the latest changes if you want?

user378230:
Commit what you have I can always pull down and test myself 😁

Jefiozie:
Here it is 😈

src/uiSelectController.js
```diff
@@ -257,8 +254,6 @@ uis.controller('uiSelectCtrl',
       if (ctrl.dropdownPosition === 'auto' || ctrl.dropdownPosition === 'up'){
         $scope.calculateDropdownPos();
       }
-
-      $scope.$broadcast('uis:refresh');
```
user378230: Why was the removed?

src/uiSelectChoicesDirective.js
```diff
@@ -68,7 +66,7 @@ uis.directive('uiSelectChoices',
         scope.$watch('$select.search', function(newValue) {
           if(newValue && !$select.open && $select.multiple) $select.activate(false, true);
           $select.activeIndex = $select.tagging.isActivated ? -1 : 0;
-          if (!attrs.minimumInputLength || $select.search.length >= attrs.minimumInputLength) {
+          if ((!attrs.minimumInputLength || $select.search.length >= attrs.minimumInputLength) && $select.open) {
```
Jefiozie: I changed this as  this will prevent the initial request of the refresh function on initializing.
Yes reverting will fix the tests.

src/uiSelectController.js
Jefiozie: Because nothing is listing on it.

user378230: Can't users listen on it? 🤔 I don't use events much in angular so not completely sure!

src/uiSelectChoicesDirective.js
Jefiozie: Based on your comment yes we can revert. But still on initializing the refresh will be exectued. And this is not the behavior as it should be i think?

src/uiSelectController.js
Jefiozie: Yes you could both  I cannot think of any use case where you need this? And if we want to support this it should be in the `refresh `function instead of `activate`.

src/uiSelectChoicesDirective.js
user378230: As I said before I want the user to check if it's open not us, users should already be able to do this. (I'm writing on mobile so my replies are short sorry!).

I also don't want to add breaking change as some users might like to load on initialise.

src/uiSelectController.js
user378230: No harm is keeping it there now, someone might use it already and we should avoid breaking their use case 😬

src/uiSelectChoicesDirective.js
Jefiozie: Okay, if that is the case then we can revert almost everything. Only thing we need to add is when `activating/focus` the control we need to execute the `refresh `function.

The users can already check if it is open by adding (as you already mentioned)  a second param `$select.open`

Jefiozie:
@user378230, so reverted the broadcast event. But also found a other use case where we could fix the original problem.By checking if the `newValue !== ''` and check if` newValue !== oldValue` we will prevent the initial call on load. As the watch is fired every time a user  types in a char it will execute the refresh function.

Second we will fire a refresh on activating the ui-select where the value of `$select.search` is already empty (as it should because nothing is present at that time).

Do you believe this is a good way to solve the problem?

p.s. all test are passing. 😄

user378230:
> But also found a other use case where we could fix the original problem.By checking if the newValue !== '' and check ifnewValue !== oldValue we will prevent the initial call on load. As the watch is fired every time a user types in a char it will execute the refresh function.

Isn't this the use case I said we wanted to keep? (Calling the refresh function when the directive is first loaded?) Or am I misunderstanding what you said...

After this PR we should have 3 different routes to calling refresh:
1. When directive is first loaded (`$select.search = ""`, `$select.open = false`) - *existing*
2. When ui-select is opened (`$select.search = ""`, `$select.open = true`) - **new**
3. When $select.search changes (`$select.search = "xyz"`, `$select.open = true`) - *existing*

You are making me work hard for this PR haha! 😆

Jefiozie:
Then this is it (i believe) reverted all changed and added the new use case. 

> When ui-select is opened ($select.search = "", $select.open = true) - new

With some tests of course.

user378230:
@Jefiozie Think you just need to review your commit once more (can you revert empty line changes too and an extra whitespace)

And tests are like gold dust in library, so the more the better! :+1:

Jefiozie:
@user378230, so after this intensive PR 😈  i think we have learned and made it a little beter. Right 😉 , again thanks for the feedback/ help and good discussion.

user378230:
@Jefiozie check the files tab on here please... 🙂

Jefiozie:
Think it looks better now.

user378230:
Queued for next released. Just want to undo the unnecessary empty line changes before I merge.

Jefiozie:
@user378230 , Any ideas when this will be merged?

aaronroberson:
@Jefiozie Can you resolve the conflicts so that I can merge? Thanks for all your work!

Jefiozie:
Sure will have a look.

Does @user378230  not maintain this repository any more?

Jefiozie:
@aaronroberson , see my previous comment.

Does @user378230 not maintain this repository any more or is he gone (as far as you know)?
Many issue and PR's are still open and I think we need to do something about it 😄

Jefiozie:
@aaronroberson , made the changes.  I can be merged also can you reply on previous comments?

Jefiozie:
@aaronroberson sorry to bother you but I'm seeing the issue list grow and want to help getting this a little bit smaller. Can you merge this? Thanks.

Jefiozie:
@aaronroberson @user378230 sorry to bother you guys. But this libary needs some attention.
Can I maybe help you guys in a way?

CC @wesleycho , maybe you can contact people above?  (thanks and sorry to bother you just wanne keep te libary "up to date")

aaronroberson:
Can you send me an email to discuss having you help maintain the project? My email is aaronaroberson at gmail.

houwei544:
when ui-select is multiple , function refresh can't get $select .

PR #1848, opened 2016-11-13
fix(uiSelectController) ActiveIndex should skip disabled choises

Jefiozie:
When pressing key up or/and key down the activeIndex should be set to a item that isn't a disabled choice.
refrence: #1823


user378230:
I think we can shorten the logic down to:
```js
...
       case KEY.DOWN:
        if (!ctrl.open && ctrl.multiple) ctrl.activate(false, true); //In case its the search input in 'multiple' mode
        else if (ctrl.activeIndex < ctrl.items.length - 1) {
          var idx = ++ctrl.activeIndex
          while(_isItemDisabled(ctrl.items[idx]) && idx < ctrl.items.length) {
            ctrl.activeIndex = ++idx;
          }
        }
        break;
      case KEY.UP:
        var minActiveIndex = (ctrl.search.length === 0 && ctrl.tagging.isActivated) ? -1 : 0;
        if (!ctrl.open && ctrl.multiple) ctrl.activate(false, true); //In case its the search input in 'multiple' mode
        else if (ctrl.activeIndex > minActiveIndex) {
          var idx = --ctrl.activeIndex;
          while(_isItemDisabled(ctrl.items[idx]) && idx > minActiveIndex) {
            ctrl.activeIndex = --idx;
          }
        }
        break;
...
```

user378230:
We also need a test for items being disabled due to being selected, if you wouldn't mind incorporating a test. Can you just check nothing breaks with tagging enabled and tagging label set to something and also set to `false`?

Thanks!  😀

src/uiSelectController.js
```diff
@@ -562,14 +562,26 @@ uis.controller('uiSelectCtrl',
 
   function _handleDropDownSelection(key) {
     var processed = true;
+    var idx;
     switch (key) {
       case KEY.DOWN:
         if (!ctrl.open && ctrl.multiple) ctrl.activate(false, true); //In case its the search input in 'multiple' mode
-        else if (ctrl.activeIndex < ctrl.items.length - 1) { ctrl.activeIndex++; }
+        else if (ctrl.activeIndex < ctrl.items.length - 1) {
+          idx = ++ctrl.activeIndex;
```
user378230: Can we just declare the `var` here? I know it leaks out anyway but it makes it clearer where its used then.

test/select.spec.js
```diff
@@ -2831,6 +2833,36 @@ describe('ui-select tests', function() {
         expect(scope.$model.name).toEqual('Natasha');
         expect(el.scope().$select.selected.length).toEqual(2);
     });
+    it('should ignore disable items key up with tagging on false', function() {
```
user378230: `should ignored disabled items in the down direction with tagging off`

test/select.spec.js
```diff
@@ -2831,6 +2833,36 @@ describe('ui-select tests', function() {
...
+      var el = createUiSelectMultiple({uiDisableChoice:"person.age == 12", tagging:false});
+      openDropdown(el);
+      var searchInput = el.find('.ui-select-search');
+      expect(el.scope().$select.activeIndex).toBe(-1);
+      triggerKeydown(searchInput, Key.Down);
+      triggerKeydown(searchInput, Key.Enter);
+      expect(el.scope().$select.activeIndex).toBe(2);
+    });
+
+    it('should ignore disable items key up with tagging on true', function() {
```
user378230: `should ignored disabled items going up with tagging on`

test/select.spec.js
```diff
@@ -3167,4 +3200,58 @@ describe('ui-select tests', function() {
     });
   });
 
+   describe('Test key down key up and activeIndex should skip disabled choice',function(){
+    it('should ignore disable items key down', function() {
```
user378230: `should ignore disabled items, going down`

test/select.spec.js
```diff
@@ -3167,4 +3200,58 @@ describe('ui-select tests', function() {
...
+      var el = createUiSelect({uiDisableChoice:"person.age == 12"});
+      openDropdown(el);
+      var searchInput = el.find('.ui-select-search');
+      expect(el.scope().$select.activeIndex).toBe(0);
+      triggerKeydown(searchInput, Key.Down);
+      triggerKeydown(searchInput, Key.Enter);
+      expect(el.scope().$select.activeIndex).toBe(2);
+    });
+
+    it('should ignore disable items key up', function() {
```
user378230: `...going up`

test/select.spec.js
```diff
@@ -3167,4 +3200,58 @@ describe('ui-select tests', function() {
...
+      openDropdown(el);
+      var searchInput = el.find('.ui-select-search');
+      expect(el.scope().$select.activeIndex).toBe(0);
+      triggerKeydown(searchInput, Key.Down);
+      expect(el.scope().$select.activeIndex).toBe(2);
+      triggerKeydown(searchInput, Key.Up);
+      expect(el.scope().$select.activeIndex).toBe(0);
+    });
+
+    it('should ignore disable items key up with tagging on true', function() {
```
user378230: Can you rename similar to others? :)

test/select.spec.js
```diff
@@ -3167,4 +3200,58 @@ describe('ui-select tests', function() {
...
+      openDropdown(el);
+      var searchInput = el.find('.ui-select-search');
+      expect(el.scope().$select.activeIndex).toBe(-1);
+      triggerKeydown(searchInput, Key.Down);
+      expect(el.scope().$select.activeIndex).toBe(2);
+      triggerKeydown(searchInput, Key.Up);
+      expect(el.scope().$select.activeIndex).toBe(-1);
+    });
+
+    it('should ignore disable items key up with tagging on false', function() {
```
user378230: same here too 👍

user378230:
@Jefiozie thanks for the amazing work, as always! 

When you have a chance can you add a test for multiple select with `remove-selected="false"` and ensure it skips those items as well? (The `remove-selected` means that items still remain in the select list after a user has selected them, they are just disabled. I think items are more likely to be disabled this way than using the `disable` expression.

Jefiozie:
As requested the extra tests and changes to the idx check.

Jefiozie:
@user378230 , Any ideas when this will be merged?

Jefiozie:
@user378230  & @aaronroberson , any idea when this will be merged?

Jefiozie:
@user378230 & @aaronroberson , any idea when this will be merged?

aaronroberson:
Done! Thanks @Jefiozie

PR #1850, opened 2016-11-17
fix(uiSelectChoices): prevent template element from being modified

zollinger:
Previously, the .ui-select-choices-row element got deleted whenever the linked scope got destroyed which prevents rendering of the options in some cases.

### DIFF
--- a/src/uiSelectChoicesDirective.js
+++ b/src/uiSelectChoicesDirective.js
@@ -61,10 +61,6 @@ uis.directive('uiSelectChoices',
 
         $select.dropdownPosition = attrs.position ? attrs.position.toLowerCase() : uiSelectConfig.dropdownPosition;        
 
-        scope.$on('$destroy', function() {
-          choices.remove();
-        });
-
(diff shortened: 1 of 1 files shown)


user378230:
@zollinger I'm going assume this is in reference to [#1365](https://github.com/angular-ui/ui-select/pull/1365#issuecomment-261025763)?

I need one of you to open an issue with a Repro just to confirm there is a problem and this does resolve it, before I go ahead and merge. (Up until no one else has reported this problem and there have been many versions since the changes - so I just want to be sure this is good to merge).

zollinger:
I added a second commit which prevents tElement from leaking to the link function scope.

user378230:
@zollinger I was thinking to ask you to do that. Thanks very much!

zollinger:
I'm not entirely sure what to test here. Adding and removing elements to the DOM is the job of AngularJS, so I we should do or test that here. Then the second change is about working with the linked element in link phase, which is obvious.

PR #1853, opened 2016-11-18
fix(uiSelectChoice): Isolate "link" function from "compile" scope

### LINKED ISSUE #1851: Ui-select-choices element removed permanently from template in some cases
### Bug description:

If ui-select with grouped options is nested within an ng-if that gets toggled a couple of times quickly without first activating the dropdown, the ui-select-choices-row element will be permanently removed from the template and thus, the options will be empty.

Simply removing the code that removes those elements from the template solves this immediate issue:
https://github.com/angular-ui/ui-select/pull/1850

This code is supposed to prevent a memory leak, however there was some refactoring done here since and I don't see why this should be a problem now. Maybe the author of  
https://github.com/angular-ui/ui-select/pull/1365 could create a plunker to reproduce the memory leak?

http://plnkr.co/edit/cpMpE6M9MU9FUfhCEoQN?p=preview

Angular: 1.5.8

UI-Select: 0.19.6

Bootstrap/Select2/Selectize CSS: n/a

raman-savitski:
In some cases when ui-select is wrapped in ng-if ".ui-select-choices-row" element
is being deleted forever.

Closes #1851 #1850


user378230:
I think this is actually resolved by #1850 so I going to close this one, as that PR is more concise. 

Thanks for your submission though.

PR #1858, opened 2016-11-29
Fix tagging without multiple.

kboga:
(Fix tagging without multiple, fixing a missing function & cleaning up changes to dist from the original pull request)


kboga:
[demo](http://plnkr.co/edit/8ZkVVOuzJqsID9ocNlNt?p=preview)

user378230:
Sorry I'm not willing to accept code duplication to solve this. The code needs to be factored out to a separate service/directive that can be applied to multiple or select.

Further more this PR contains **0** unit tests.

If you'd like to update this, I'd be happy to review again but in it's current state I cannot accept. 

Thank you.

PR #1866, opened 2016-12-08
Update common.css

c0dem0nkee:
allow-search=false does not work if theme is set to 'select2' on config.  Added these lines to fix this


aaronroberson:
@c0dem0nkee Thanks for the PR. Can you update your PR to conform to the contributing guide here: https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md

PR #1872, opened 2016-12-15
feat: Include padding when resizing multiple input

pete-otaqui:
Remove the hard-coded 10px when accounting for padding, and use the calculated padding size instead, in case the theme has been overridden.

Note that the test appends the generated uiSelectMultiple to the DOM (and later removes it) so that getComputedStyle functions normally.


pete-otaqui:
These four failing tests (which are not related to this PR) seem to happen on a fresh checkout from master now.  Not sure why, but I think it's probably a dependency-version thing.   I will be investigating further.

PR #1873, opened 2016-12-16
fix(uiSelectCtrl) calculateContainerWidth use width of ui-select-container

bmsdave:
( not from body, if I set attribute `append-to-body=true`)

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -522,7 +522,7 @@ uis.controller('uiSelectCtrl',
   ctrl.sizeSearchInput = function() {
 
     var input = ctrl.searchInput[0],
-        container = ctrl.searchInput.parent().parent()[0],
+        container = ctrl.$element[0],
         calculateContainerWidth = function() {
           // Return the container width only if the search input is visible
           return container.clientWidth * !!input.offsetParent;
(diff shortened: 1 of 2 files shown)


bmsdave:
These four failing tests (which are not related to this PR) seem to happen on a fresh checkout from master now. Not sure why, but I think it's probably a dependency-version thing. I will be investigating further.

pete-otaqui:
Related to #1872, suggest this PR should also account for actual padding of the field rather than a hard-coded 10px.

pete-otaqui:
Also the failing tests are logged as #1877

bmsdave:
@pete-otaqui, Hi.
I need some of the actions that have been adopted this PR?

bmsdave:
@Jefiozie, @pete-otaqui, hi. This PR is actualy. Pls, tell me what I should do to the  issue #1874 was solved?

Jefiozie:
I think you have added to much thing. Please clean up the select.js etc. Hope i can have a look later this week .

bmsdave:
@JeffGreat, done.

Jefiozie:
@bmsdave could you rebase this? Thanks

bmsdave:
@Jefiozie, done.

Jefiozie:
Could you also make a test around this? So that we now when behavior is changing? 

Thanks

bmsdave:
Jefiozie, 
OK. I'll do everything on the weekend! =)

Jefiozie:
Could you also have a look at: #1906 looks similar issue.

Thanks

easyest:
This line is still not changed:
var inputWidth = containerWidth - input.offsetLeft - 10;

At least in bootstrap there is no need to subtract 10 px because input width should be the same as button width.

bmsdave:
@Jefiozie, hi.
test and remove -10px is done.
plz, check it.

Jefiozie:
Hi could you rebase this? After this I will merge the PR. Thanks

bmsdave:
@Jefiozie, like that? i just resolve conflict.

PR #1879, opened 2016-12-19
fix(spec): Use exceptionHandler to test errors.

### LINKED ISSUE #1877: Failing tests with Angular 1.6
4 tests fail with Angular 1.6 (the latest version that is installed by default since it's release).

Specifically related to this breaking change between 1.5 and 1.6:

https://github.com/angular/angular.js/commit/e13eeabd7e34a78becec06cfbe72c23f2dcb85f9

Exceptions inside promises are now handled as rejections.

pete-otaqui:
Angular 1.6 alters the way uncaught promises are handled - errors cause promise rejection rather than outright exceptions to be thrown.  Update existing exception tests to use the exceptionHandler service instead of `.throws()`

Closes #1877

### DIFF
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -1,7 +1,7 @@
 'use strict';
 
 describe('ui-select tests', function() {
-  var scope, $rootScope, $compile, $timeout, $injector, $q,uisRepeatParser ;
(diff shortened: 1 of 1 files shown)


aaronroberson:
Thanks for the PR @pete-otaqui it will be in the next release!

PR #1897, opened 2017-01-27
Disabled animation

daywong1119:


Jefiozie:
I think something went wrong. 

On the commit you should not commit the dist page. As this should be done by the build server. Next I'm wondering why you made this PR and give some more information.

daywong1119:
Yea, this is not done right.

PR #1900, opened 2017-01-31
Allow manually send event for update element width (close #1666)

### LINKED ISSUE #1666: Placeholder text is clipped to 10px sometimes

indvd00m:
This is workaround to fix placeholder width=10px if ui-select was not visible when created. It is possible now fix this by sending special event: $broadcast('updateUISelectSize').

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -739,6 +739,8 @@ uis.controller('uiSelectCtrl',
     angular.element($window).off('resize', onResize);
   });
 
+  $scope.$on('updateUISelectSize', ctrl.sizeSearchInput);
+
   $scope.$watch('$select.activeIndex', function(activeIndex) {
     if (activeIndex)
       $element.find('input').attr(


indvd00m:
Pull request for #1666

Jefiozie:
Hi @indvd00m ,

I don't really agree on this approach. This is a very specific fix that will help but not resolve the problem.
Can you have a look at the real problem? If so i'm happy to review. But this PR i cannot approve.

PR #1906, opened 2017-02-08
Wrong input width calculated

easyest:
Search input width is currently calculated by subtracting 10 pixels from it's second parent width. But it should be the same, as the drop-down button width.


Jefiozie:
Thanks for the PR, can you check why the tests are failing. 
Second please add a test  to test the behavior for now and in the future.

Jefiozie:
HI @easyest , I have merged my PR could you rebase to the master?
Thanks.

easyest:
I would be glad to do that, but I do not use git at all. This PR was done by a simple edit in the web form :)

Jefiozie:
okay, i just saw that this is a duplicated issue. So closing it down as the other has a PR.

Issue #1916, opened 2017-02-17
Failing unit tests.


Jefiozie:
### Bug description:

It looks like there has been adjustments to the test what is resulting in a problem during PR's.

### Link to minimally-working plunker that reproduces the issue:
See last four PR's

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: 

UI-Select: 0.19.5/0.19.6

Bootstrap/Select2/Selectize CSS (if applicable):
All

Jefiozie:
PR is merged, if needed reopen it when problems is not solved.

PR #1917, opened 2017-02-17
fix(unit tests) Changes for angular 1.5+

Jefiozie:
For now i'm reverting these changes. As i believe this a problem (at this moment).
Needs some more investigation why this is happening and how to resolve it. 

#1916


Jefiozie:
Reverting: #1879

PR #1908, opened 2017-02-08
Change $select.clear() so it sets value to `null` instead of `undefined`

v-andrew:
Type: fix
Scope: uiSelectMatch

Fixes https://github.com/angular-ui/ui-select/issues/863

PR #1918, opened 2017-02-21
Change to set null instead of undefined when ctrl is clea

cloudrick:
[why] Null is more readable for empty value.
[how] * When ctrl is cleared, set null to its selection.
      * In ngModel.$parsers, add a handler to
        ) return null for null value.
	) return undefined for undefined value.
      * Also modify test case for checking clear value.


Jefiozie:
Closing  it down as it is a duplicate of #1908, please comment on that PR if you see problems with the proposed solution. 

Thanks.

PR #1924, opened 2017-02-23
Adding support for non existing values (allow free text) in single select

afro1114:
This uses most of the code already existing for tagging in multiple selection, however it uses a different attribute because there are parts of the code that are not shared and only apply to single selection.

Using the feature is done in the same way tagging does: 
allow-free="transformFunction"


Jefiozie:
HI @afro1114 ,

Thanks for the PR but for now we don't do any new features as there are many bugs.
Next to this I should refer you to the [CONTRIBUTING.MD](https://github.com/angular-ui/ui-select/blob/master/CONTRIBUTING.md) you have a lot of file changes as you didn't clean up everything.

Hope you understand the situation and maybe you can assist in other bugs.

Thanks.

afro1114:
Well there is actually a bug regarding this topic 1409

nickiowaspain:
So wait @Jefiozie and @afro1114. Has this been fixed. I am also running into this problem.

Jefiozie:
Looks like I made a mistake. I will reopen it and have a look at it. @nickiowaspain  It is not fixed for now.

afro1114:
Hi @Jefiozie in this case what is the tag _needs:work_ referring to?

Jefiozie:
Hi @afro1114 the cleaning up part as in my previous comment addressed [comment](https://github.com/angular-ui/ui-select/pull/1924#issuecomment-282074598)

MathRobin:
Hi,

What kind of works is needed to merge this request. I really need this feature currently and don't want to use a fork ? So if i can help but don't know where to start

brunosantosnasajon:
Any progress on this issue?

Issue #1278, opened 2015-10-27
Text overflow on ui-search control


smithk58:
Can be seen in the top ui control in your demo plunkr: http://plnkr.co/edit/a3KlK8dKH3wwiiksDSn2?p=preview

If you select too long of an element it overflows out of the control. It looks like it's happening because the .btn class is applying white-space:nowrap.

Here's a quick/dirty/temporary fix for someone if they need it:
.ui-select-match-text{
    display:inline-block;
    overflow:hidden;
    text-overflow: ellipsis;
    width:95%;
}

AlexEmashev:
Thank you, your fix helped me!

denis-bondarenko:
there is a way better solution
    .ui-select-toggle{display:flex;justify-content:space-between;}
    .ui-select-match-text{overflow-x:hidden; text-overflow:ellipsis;}
here is working plunker http://plnkr.co/edit/d4gCXgLHaPKmP89uiP6e?p=preview

jziggas:
Both of those examples show default demos.

smithk58:
If you're referring to my example and @denis-bondarenko's examples:
-Mine IS pointing at the default demo. I was only giving steps to reproduce the issue.
-Denis's example isn't the default demo. Apply the reproduction steps I listed to the one he linked to and you'll see it doesn't overflow.

Wenakari:
Thank you very much smithk58 ! It's work perfectly !

PR #1933, opened 2017-03-06
Using the `no-wrap` clas in the select in order to display multiline options

wuilliam321:
This allows the user to have options with multilines if the option names are too long.


Jefiozie:
Can you explain a bit more on this PR? And why is this only needed for the bootstrap theme?
Thanks.

wuilliam321:
The reason is: If you have to display a select tag in a short space 'col-md-2' for example, additionally with very long names, you will see that the option only uses one line, and sometimes we need to display the full option text, with no-wrap this Is possible. This is the main reason. Why bootstrap? We are using it.

Thanks for response

DarthVanger:
Related to issue #1278

SebSob:
@wuilliam321 Nice, this fix works and shows the label correctly when the dropdown is open, but once selected the label get overflowed:
image

Can you help?

DarthVanger:
@SebSob Did you restrict width of your `<ui-select>` ? What you're showing is not overflow, but the whole input gets wider.

See this plunker which has `300px width` for the `<ui-select>`, and the selected text gets cut.
https://plnkr.co/edit/RXNBGedUZktGShnHbFWG?p=preview
The only thing I've added is ellipsis and some space to make it look prettier.

SebSob:
@DarthVanger Thanks, I know it works if you restrict the width, but in my scenario i just have it inside a row/column. See this plunker (updated yours): https://plnkr.co/edit/g3DxMLR6WhcNt0KB6C5I?p=preview

DarthVanger:
@SebSob the reason is the `.input-group`, which has `display: table`.
If you remove that `.input-group` class, it works  
https://plnkr.co/edit/LD7n5ZjocZgh13vNSSES?p=preview

Don't know why input-groups work this way inside of columns, but looks like it's not compatible - they are just overflowing the column width.

Anyways it's not the problem of `ui-select`: you are just not restiricting the width, so it it has the same width as your `.input-group`.

Issue #1583, opened 2016-04-27
Calculates of dropdown position lead to scrollbar flickering


lexey111:
Hi, 
we're using custom scrollable container and ui-select inside. It looks like `calculateDropdownPos` leads to flickering the scrollbar if position set to 'auto' and long select is placed at the bottom of form.

I did some quick search and change function a bit: except just `dropdown[0].style.opacity = 0;` ... `dropdown[0].style.opacity = 1;` I added `dropdown[0].style.position = 'fixed';` ... `dropdown[0].style.position = 'absolute';` to move dropdown out of page flow temporary.

It solves the problem. Unfortunately I can't make PR due to some reasons, but I think this is not very big change.

user378230:
@lexey111 Can you use the issue template provided and provide plunkr and version numbers?

For reference: There was some attempt to fix something similar in #1557, #1123 & https://github.com/angular-ui/ui-select/pull/1557#issuecomment-210851842 maybe it would be useful for your PR.

Dexmaster:
http://plnkr.co/edit/FMUk3ZzrnDqpNJ3A9AUz here is a demo of a flicker.
The problem especially visible when we don't know direction +modal window or footer.

Dexmaster:
Fix for situation when item goes out of viewport appears flicker before changing dropdown positioning. Here is fixing PR: and plunk of a fix: http://plnkr.co/edit/QxycwbMMYUpNoTLiubEg

MatejQ:
We are seeing this same issue. Looking forward to a fix.

PR #1936, opened 2017-03-09
fix(uiSelect): add class to dropdown instead of simple opacity change…

Dexmaster:
…  to remove flicker

Previous behaviour and new one with punks are in issue #1583.

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -17,6 +17,15 @@
   top: 0px !important;
 }
 
+/* send offscreen without breaking sizes*/
+.ui-select-detached {
+    position: fixed !important;
+    opacity: 0 !important;
+    visibility: hidden;
+    top: 100% !important;
+    left: 100% !important;
+    z-index: -1 !important;
+}
 
 .ui-select-choices-row:hover {
   background-color: #f5f5f5;
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -373,7 +373,7 @@ uis.directive('uiSelect',
             }
 
             // Display the dropdown once it has been positioned.
-            dropdown[0].style.opacity = 1;
+            dropdown[0].classList.remove('ui-select-detached');
           });
         };
 
@@ -389,7 +389,7 @@ uis.directive('uiSelect',
 
            // Hide the dropdown so there is no flicker until $timeout is done executing.
            if ($select.search === '' && !opened) {
-              dropdown[0].style.opacity = 0;
+              dropdown[0].classList.add('ui-select-detached');
               opened = true;
            }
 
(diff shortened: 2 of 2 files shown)


src/common.css
```diff
@@ -17,6 +17,15 @@
...
 }
 
+/* send offscreen without breaking sizes*/
+.ui-select-detached {
+    position: fixed !important;
+    opacity: 0 !important;
+    visibility: hidden;
+    top: 100% !important;
+    left: 100% !important;
+    z-index: -1 !important;
```
Jefiozie: Can you explain why the important is needed?

src/uiSelectDirective.js
```diff
@@ -411,7 +411,7 @@ uis.directive('uiSelect',
             }
 
             // Reset the position of the dropdown.
-            dropdown[0].style.opacity = 0;
+            dropdown[0].classList.add('ui-select-detached');
```
Jefiozie: Is it possible that this class is already being set before it is in this function?

src/common.css
Dexmaster: Because you set top/left/position with propery of DOM element `dropdown[0].style.top` `dropdown[0].style.position`. z-index is additional and probably not needed took from other css class I used some time ago.

src/uiSelectDirective.js
Dexmaster: No it's new class, I haven't added it to templates so it's here first time and we see addition on two sides of `if ($select.open)` so it's either one or another.
 And even if class was added already it will not duplicase as `classList.add` adds only if not present.

Dexmaster:
What tests do you need specifically? It does not change interaction and just replaces opacity with class change?

Dexmaster:
I can do one test that it will fix by replication issue https://github.com/angular-ui/ui-select/issues/1583 itself but I believe to do it for such a small specific issue is a bit overkill.

Jefiozie:
I agree on the fact that it looks like a bit of overkill. But as this libary has many issue it is our best check for problems in the future. Could you make it and commit it?

Dexmaster:
To this same branch, no problem. I need to add comit with `test(uiSelect): PR #1936` or copy same messege but starting with test?

Jefiozie:
i think `test(uiSelect): PR #1936` will do. It can be in the same PR (just to be clear 😄 )

Jefiozie:
Any updates on this issue? 

Thanks

Dexmaster:
Oops, forgot completely, will take some time tomorrow evening to clean it up.

MatejQ:
We are seeing the same issue as in #1583 however when applying the changes in this PR, it did not help. Looking forward to a proper fix.

afedosenko:
Good solution, but doesn't seem to work when $select.$animate is null.

PR #1939, opened 2017-03-14
Add attr (tagging-equals) to set a property for equality in custom objects (tagging)

cgaggino:
When you supply a function for tagging, the new item is compared with the existing ones by angular.equals. This is good for primitives but when you are using custom objects they may not be equals due to one ore more properties.
The idea here is to add a new directive attribute (tagging-equals) to set a property by witch the custom object will be compared.


Jefiozie:
@cgaggino Thanks for submitting this PR. I think this could be something for the futher. For now the focus of this repo is to get less bugs and less new enhancements. 

Still i'm curious could you share a working plunker with some custom functions for this? 
Thanks.

cgaggino:
Hi @Jefiozie thanks for replying, I'll be adding the tests and the plunker as soon as possible

Kind regards!

PR #1955, opened 2017-04-04
Fixed issue with tagging and tagging-label='false'

giacomolicari:
Reference: https://github.com/angular-ui/ui-select/issues/1214#issuecomment-291153134


giacomolicari:
Test cases not updated

Jefiozie:
Hi, I think you have added to much as I see all files in the dist dir. Could you do a clean up so that we can have a look at the PR?

Thanks.

Jefiozie:
It looks like you still have some conflicts. Could you fix them before i'm reviewing them?

PR #1958, opened 2017-04-05
disable drop under CSP

graingert:


src/uiSelectSortDirective.js
```diff
@@ -31,6 +31,10 @@ uis.directive('uiSelectSort', ['$timeout', 'uiSelectConfig', 'uiSelectMinErr', f
         }
       });
 
+      if (element.data('disallowDrop')) {
```
graingert: rather than overriding the drop handlers in the html, simply don't set them.

Jefiozie: Is this needed? as I don't see anything in the code that is looking for `ondrop`?

graingert: That's because the browser does it

Jefiozie: you are right. then we are good to go here 👍

graingert:
@Jefiozie any idea when this might get released?

PR #1959, opened 2017-04-05
fix(uiSelectMatch): set model value to `null` when cleared

### LINKED ISSUE #863: Clearing the selection sets model value to 'undefined'
Clearing the selection in a non-multiselect which has the "allow-clear" attribute sets the model value to undefined, while the expected value would be null.

Being undefined causes Angular to ignore the change, so it won't be sent to the server.

jrbotros:
Closes #863

### DIFF
--- a/test/select.spec.js
+++ b/test/select.spec.js
@@ -15,6 +15,10 @@ describe('ui-select tests', function () {
     Escape: 27
   };
 
+  function isNil(value) {
+    return angular.isUndefined(value) || value === null;
+  }
+
   //create a directive that wraps ui-select
   angular.module('wrapperDirective', ['ui.select']);
   angular.module('wrapperDirective').directive('wrapperUiSelect', function () {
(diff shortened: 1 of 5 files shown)


test/select.spec.js
```diff
@@ -636,6 +639,31 @@ describe('ui-select tests', function () {
     expect(el.find('.select2-search-choice-close').length).toEqual(1);
   });
 
+  it('should clear selection (with object as source)', function() {
```
Jefiozie: Could you maybe make use of `createUISelect` and /or `createUIMultipleSelect` function in the tests?

jrbotros: The main `should clear selection` test still uses `createUISelect`, but I didn't see a way to test the "object as source" syntax without compiling the template manually (I used the `should bind model correctly (with object as source)` test as a template). Any ideas how to make this cleaner?

Jefiozie: ah missed that yeah hmm.. maybe we could make a new function `createUiSelectObjectSource` or something where the template is build and can be used in the future?

jrbotros: Let's get this landed and open another issue for doing some test cleanup? Happy to take a stab at it in a few days (probably over the weekend)!

Jefiozie: Maybe it is best I want to do some major refactoring to this repo, for example it needs to have seperate directives for tagging etc.

Issue #1966, opened 2017-04-12
Proposal chore changes


Jefiozie:
The last couple of months I have been working as a collaborator of this repo.
There are **MANY** issues that needs to be resolved and based on the amount of issues and 219 test cases it think something needs to be done.
Based on some previous issue (#1096, #784, #402, #765) I think there needs to be some chore changes.

Proposed changes:

1. Remove all built in tagging, tagginToken etc. and create separate directives for this
2. Have a look at the $watch property's and see if we can make this better. 
3. Have a look at the repeat parser there are some issues open for this where use cases don't work (maybe outdated code?)
4. Maybe migrate to ES6 => webpack?
5. Thinking about a release calendar or something where everybody can see when will be released and what?

After this I think we have some more readable and better maintainable code after this we could have a look at upgrading to Angular?

Please fill free to comment and make proposals on how this can be done best or what I missed.

Issue #1444, opened 2016-02-17
Angular 1.5 - Uncaught Error: [ui.select:transcluded] Expected 1 .ui-select-match but got '0'.


mikemclin:
Code broken when I upgraded from Angular 1.4 to 1.5  Now I get the following error:

`Uncaught Error: [ui.select:transcluded] Expected 1 .ui-select-match but got '0'.`.

Looks like I'm not the only one running into the issue (look at the last comment) - https://github.com/angular-ui/ui-select/issues/71.

mikemclin:
Actually https://github.com/angular-ui/ui-select/commit/b1c4fdc5919dd0f5c66926498c28114a48334af9 fixes this issue.

Waiting on a release.

rodmax:
+1 to quickly release this :) because it is blocker to bump angular 1.5.x

dbsoftltd:
+1 for a quick release :) Testing the changes in local environment but would prefer the official build.

misaizdaleka:
+1

Wereq:
+1

Mo7medFouad:
+1

gino8080:
+1

Azsael:
+1

armanforghani:
+1

Narretz:
@aaronroberson wdyt about a quick patch release that includes the 1.5. fix?

aaronroberson:
#1430 has been included in the v0.14.3 release.

Sixthdim:
Hmm, I'm still getting the `Expected 1 .ui-select-match but got '0'.` error in 0.14.3 and Angular 1.5.

aaronroberson:
@Sixthdim Can you provide a stack trace, any additional details or plunkr example?

Sixthdim:
@aaronroberson  Sure, here is the stack. I'll see if I can get a fiddle up today if this doesn't help.

Error: [ui.select:transcluded] Expected 1 .ui-select-match but got '0'.
    at http://localhost:8080/js/select.js:125:12
    at http://localhost:8080/js/select.js:1038:19
    at publicLinkFn (http://localhost:8080/js/angular.js:8105:29)
    at http://localhost:8080/js/angular.js:8447:29
    at boundTranscludeFn (http://localhost:8080/js/angular.js:8244:16)
    at controllersBoundTransclude (http://localhost:8080/js/angular.js:9020:20)
    at http://localhost:8080/js/select.js:1026:9
[...]

Issue #1096, opened 2015-07-21
Code Audit


ProLoser:
Hey guys, this repo pops up in discussion, and I realize with 400+ issues you are barely able to keep up with the onslaught of issues. I did some poking around and found the distributed un-minified file size is [74kb](https://github.com/angular-ui/ui-select/blob/master/dist/select.js). That's [nearly half of ui-bootstrap's filesize](https://github.com/angular-ui/bootstrap-bower/blob/master/ui-bootstrap-tpls.js) if you throw in templates.

I poked around the src files and I have to say I'm completely lost most of the time.

I think (if it's at all possible) it's time to streamline this project and get rid of and merge as many bells and whistles as possible. I realize this is one of the most complex widgets that tend to pop up (aside from a wysiwyg editor or a grid) but if ever there was an opportunity for simplification, this project could merit it.

Not that I have been keeping up-to-date as to the latest trends of the project, but here are some examples:

1. Use flexbox instead of JS sizing the input box.
2. Choice locking? Disabled choices? Choice-related garbage? Nah, lets git rid of all that from the core
3. Implement lifecycle callback hooks: `beforeSelect`, `afterSelect`, `beforeRemove`,`afterRemove`. If we use these, it's ridiculously easy for developers to implement <kbd>2</kbd> however the hell they want.
4. Remove `repeat` attribute (+ the parser)? This may be more difficult, but what if we stop making `<select-choices>` represent 1 row and instead put the onus on the developer to handle repeating? What if you had to specify the `<ul>` yourself too? No more passing of classes or HTML, no more positioning or styling garbage. No more grouping or sorting or filtering garbage. The user entirely decides what he wants to show on the list, and what he wants to perform when the user clicks on a row. We simply relocate this logic to the templates.
5. Sortable should be distributed in a separate file. It should not come with the core. The whole point of repacking sub-features into sub-directives is so that it wouldn't bloat the core, it would reduce filesize, and it would teach / encourage other developers to create their own plugins instead of adding to the 400+ bugs on this repo.
6. Grouping should be a separate file / plugin. If we can't implement it as a plugin, our API needs to be improved.
7. What is this `refresh` stuff? Can't we remove it?
8. Get rid of `angular.closests()`. It's not even used in the repo, just in the tests!
9. Get rid of `querySelectorAll()` polyfill. [It's in all the browsers that count](http://caniuse.com/#search=querySelectorAll) and I don't even think the version of angular we're using supports prior browsers.
[...]

PR #1973, opened 2017-04-14
refactor(build): replace Gulp build with Webpack

jrbotros:
Address part of #1096 and #1966.

This is pretty much all working, but I'm getting a couple frustrating errors in the tests. It looks like there's no `.ui-select-match` rendered by the time this [`transcludeFn`](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectDirective.js#L217) runs for the `selectize` theme (getting the same error as in #1444).

It looks like in `selectize/match-multiple.tpl.html` the `ui-select-match` class only exists on `ng-repeat` elements, so maybe those elements haven't been loaded yet? I modified the directives to use the template HTML directly, which I'm assuming subtly changed the timing of template loading and broke the sort-of-weird transclusion logic. Any help/ideas here would be much appreciated!

Also, what's the easiest way to actually see the code in use? I didn't see any examples that read in from the dev code directly, but I could be just missing something obvious! The other tests pass, but want to make sure everything looks shipshape.

@ProLoser @Jefiozie I know you both had interest in getting things cleaned up, so let me know what you think. If this looks good, I'd love to do a broader ES6 rewrite and standardize code style.


ProLoser:
Can you fix yarn and travis so that we can see the errors you're getting?

jrbotros:
My bad, bumped the node by a minor version which looks like it fixed things.

ProLoser:
Sorry I'm too far removed from angular and this code to be of any help any longer. I was looking through it but I can't recall at all what's going on.

jrbotros:
Gotcha, do you know anyone who would be the go-to person? I'd be down to rewrite some of the transclusion code but I'd love to get this build stuff working without changing any underlying logic.

@Jefiozie any ideas on your end?

ProLoser:
You can try brute force debugging. Try changing the logic back to using templateUrl (temporarily) and see if that fixes it. Try to eliminate other factors by hard coding and reverting things  in small pieces until you can identify the problem area. Then once that problem area behind working, try rolling back your modifications to the other code and see if it remains working until you eventually are left with the only part of code that breaks the build temporarily hardcoded hopefully making it easier to fix.

Jefiozie:
Hi, 

First off all thanks for all the effort you already put in this PR.

As I described in #1966 I feel the need to go this way but at the moment I think we first focus us on getting the code cleaned.  One of the main reasons there are this many bugs is that everything it put in the one or two files that should have been put in separate directives.  

One of the things we will see during the refactoring is that we will have code that is being used ( one of the points you pointed out).

Do you agree on this? Or  would you first convert to es6 before cleaning it up?

ProLoser:
Based on my own experience and Ember's findings: small transitions are best, don't focus on waiting for major factors if you can. Also, don't focus on backwards compatibility as it can be very limiting. You can emulate angular by using frequent major releases, deprecating and then removing old stuff, and just trying to iterate quickly.

jrbotros:
Converting to ES6 should be a pretty painless process, I'm mainly just planning on making sure the repo conforms to a standard style ([airbnb](https://github.com/airbnb/javascript) is my preference, but let me know if you feel differently) without doing any logic refactors. Once webpack is set up, we can do this in as small chunks as we want, probably something like one file at a time. And then I think it will be much easier to see how all the parts fit together. Also, we'll have access to all the luxuries of ES6 (the module system in particular), which I think will be crucial in a broader cleanup. @Jefiozie does that sound reasonable?

I'll try to fix the transclusion bug in the next few days—will keep you updated!

jrbotros:
I fixed the transclude bug by adding the `ui-select-match-wrapper` class to all the match templates. Not very elegant, but a good hack to get this in, and later we can explore [transclusion slots](https://blog.thoughtram.io/angular/2015/11/16/multiple-transclusion-and-named-slots.html) in Angular 1.5 or a [third-party solution](https://github.com/zachsnow/ng-multi-transclude) if we don't want to upgrade. I think this is ready and the demos in my local docs are working for me, but would love a second look!

@Jefiozie did you have thoughts on the refactor strategy I proposed? I'd be happy to finish file-wise ES6ifying by the end of the weekend!

Jefiozie:
Hi @jrbotros , I will try and have a look today at your PR and make a comment on the way forward. 
Sorry for the delay busy times 😄

Jefiozie:
So I had a look at your repo and I think we need to merge it (in the end). 
Can we also merge the test suite to webpack?  (as in can we run the tests before building the bundle?)

Probably have some questions later but for now I think we need to get this finished and merged.

jrbotros:
Awesome, sounds like a plan!

We can't run the tests before building the bundle since the code needs to be transpiled from ES6 first. Right now karma is set up to use webpack, so `karma start` will internally watch and bundle before running the tests and `gulp test` will work the same as before. (Karma will also transpile the tests, so we can use ES6 there, too.)

jrbotros:
What specifically do you want cleaned up? Since we're now using webpack it's a good idea to include all the sourcemaps, and I've included the minified build as well for users who want that optimization.

jrbotros:
Actually forgot, I made sure to build exactly the same files in `dist` that are currently there!

jrbotros:
@Jefiozie Any update here?

PR #1977, opened 2017-04-18
fix(uiSelectController): invoke onRemove callback method for single s…

### LINKED ISSUE #1225: `on-remove` doesn't work with single select and `allow-clear`.
It appears that currently the `onRemove` callback is only executed when a multiselect option is removed. It makes sense that it should also be called when the user clears a single select when `allow-clear="true"`.

kuuurt13:
…elect on clear selection

Closes #1225


test/select.spec.js
```diff
@@ -1425,6 +1425,35 @@ describe('ui-select tests', function () {
...
 
+  it('should invoke remove callback on remove for single select with allowClear enabled', function () {
+    scope.selection.selected = scope.people[5].name;
+
+    scope.onRemoveFn = function ($item, $model) {
+      scope.$item = $item;
+      scope.$model = $model;
+    };
+
+    var el = compileTemplate(
```
kuuurt13: Sure, no problem. Updated test to use `createUiSelect` instead.

src/uiSelectController.js
```diff
@@ -454,9 +454,16 @@ uis.controller('uiSelectCtrl',
   };
 
   ctrl.clear = function($event) {
+    var locals = {};
+    locals[ctrl.parserResult.itemName] = ctrl.selected;
+
     ctrl.select(null);
     $event.stopPropagation();
     $timeout(function() {
+      ctrl.onRemoveCallback($scope, {
```
Jefiozie: Could this function been refactored to be on the `uiSingleSelect` directive?

Jefiozie: I mean then the `clear` function. As the `uiSelectMultiple` has his own `remove` function. I think it would be good to keep  this kind of functionality on the correct place. What do you think?

kuuurt13: @Jefiozie I see what you are saying. I might be overlooking something here but it seems to me that `uiSelectMultiple` has the `removeChoice` method on its controller and then the `uiSelectMatchDirective` view uses the `uiSelectMultiple` controller to call it like so `ng-click="$selectMultiple.removeChoice($index)"`. 

Currently `uiSingleSelect` doesn't have a controller so I would have to add the method directly to the scope of `uiSingleSelect` in the link function. Then I would call it from the `uiSelectMatchDirective` single view like so `ng-click="remove($event)"`. I just haven't seen too many methods bound directly to the scope in the library so I wasn't sure if this was correct. What do you think?

cime:
Are we waiting an anniversary for this to be solved? It's really annoying there is no way to know when clear is called...

PR #1982, opened 2017-04-23
feat(Header & Footer): Add header & footer to the dropdown list - iteration 2.

felixmosh:
This feature allows to add header & footer to the dropdowns for each of the default templates.

image

select2
selectize

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -13,10 +13,9 @@
   overflow: hidden !important;
   position: absolute !important;
   outline: 0 !important;
-  left: 0px !important;
-  top: 0px !important;
+  left: 0 !important;
+  top: 0 !important;
 }
-
 
 .ui-select-choices-row:hover {
   background-color: #f5f5f5;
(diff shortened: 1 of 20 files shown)


felixmosh:
Any updates?

okolomiets:
The feature seems to become very expected one. Could anyone to review and merge it within next release? Thanks in advance.

Jefiozie:
It is on the list. Only problem is I don't have a lot of time at the moment.
When I find the time I will do the review and then it will be there in next release.

src/common.css
```diff
@@ -26,68 +25,84 @@
...
+  border-radius: 4px; /* FIXME hardcoded value :-/ */
+  border-bottom-left-radius: 0;
+  border-bottom-right-radius: 0;
 
-    border-top-width: 1px;  /* FIXME hardcoded value :-/ */
-    border-top-style: solid;
+  border-top-width: 1px; /* FIXME hardcoded value :-/ */
+  border-top-style: solid;
 
-    box-shadow: 0 -4px 8px rgba(0, 0, 0, 0.25);
```
Jefiozie: Can you change the spacing on this file? Looks like a lot of changes have been made but it are just spaces. 
thanks.

src/uiSelectFooterDirective.js
```diff
@@ -0,0 +1,15 @@
...
+  return {
+    templateUrl: function (tElement) {
+      // Needed so the uiSelect can detect the transcluded content
+      tElement.addClass('ui-select-footer');
+
+      // Gets theme attribute from parent (ui-select)
+      var theme = tElement.parent().attr('theme') || uiSelectConfig.theme;
+      return theme + '/footer.tpl.html';
+    },
+    restrict: 'EA',
```
Jefiozie: Should it match also the `attribute` tag? Or only the `element`?

src/uiSelectHeaderDirective.js
```diff
@@ -0,0 +1,15 @@
...
+  return {
+    templateUrl: function (tElement) {
+      // Needed so the uiSelect can detect the transcluded content
+      tElement.addClass('ui-select-header');
+
+      // Gets theme attribute from parent (ui-select)
+      var theme = tElement.parent().attr('theme') || uiSelectConfig.theme;
+      return theme + '/header.tpl.html';
+    },
+    restrict: 'EA',
```
Jefiozie: Should it match also the `attribute` tag? Or only the `element`?

test/select.spec.js
```diff
@@ -3378,4 +3378,52 @@ describe('ui-select tests', function () {
       expect(el.scope().$select.activeIndex).toBe(2);
     });
   });
+
+  describe('Header & Footer', function () {
+
+    var el;
+
+    function setupSelectComponent(theme, isMulti) {
```
Jefiozie: We have 2 functions for building up the template `createUitSelect` and `createUiSelectMultiple`.  I believe this functions can be changed to support theses tests.

test/select.spec.js
```diff
@@ -3378,4 +3378,52 @@ describe('ui-select tests', function () {
...
+          <ui-select-footer>{{' + selectedStr + '.name}}</ui-select-footer> \
+        </ui-select>'
+      );
+    }
+
+    ['selectize', 'bootstrap', 'select2'].forEach(function (theme) {
+      describe(theme + ' theme', function () {
+        it('should show the header', function () {
+          setupSelectComponent(theme);
+          expect($(el).find('.ui-select-header').text().trim()).toBe(scope.selection.selected.name);
```
Jefiozie: Why are your trimming the text? Is there any reason for it?

felixmosh: Cause it is a text in html element, and it can contain spaces...

src/uiSelectFooterDirective.js
felixmosh: It can be both

src/uiSelectHeaderDirective.js
felixmosh: It can be both

src/common.css
felixmosh: The spacing at the PR is the correct one, pay attention, there is inconstant spacing at the original file... i`ve just formatted it to be aligned.

test/select.spec.js
felixmosh: Yeah, but it doesn't have the `header` & `footer` tags...

trosenfeld:
Just a comment:
As I was looking to use this feature I had made a local copy with these modifications.
When using the header and/or footer feature the options area in the dropdown no longer scrolls.
This is fairly annoying as the whole list of options is displayed which "elongates" the site when opening the dropdown. (We have lists of sometimes over 200 items)
You might also want to look into this when/if you get the time before releasing.

For now I'm just using the "master" ui-select and adding "disabled" items to the select items list using the "ui-disable-choice" attribute on "ui-select-choices" to prevent selection.

felixmosh:
@trosenfeld, did u copied the styles as well?
U try to clone the PR, and run it locally.
I will be happy to look into your copy, to figure out what is wrong with it.

felixmosh:
@Jefiozie any updates?

GeneGenie:
So. when it will be available?

felixmosh:
It depends on the author, he from some reason not approves it :/

shyamal890:
@Jefiozie May be we should update this: https://github.com/angular-ui/ui-select/wiki/ui-select-header-and-footer

bpaugam:
When will this be released ?

besafx:
Can You Give Us A Code That We Can Add Into Select.JS to Enable This Feather

pavel-blagodov:
Does anybody work on this repo? No commits since Mar 5

PR #1993, opened 2017-05-11
Feature/ada update

zorigitano:


Jefiozie:
Hi, Could you clean the dist file and maybe explain why you changed the package.json file?

PR #1994, opened 2017-05-17
BugFix: removing the label when new tag is clicked on

shmuel-web:
### TL;DR
When creating a new tag and clicking to add it to multi tag select, the tag remains named "Tag(new)" with the label even after it's created.
How to reproduce:
go to this [plunker](http://plnkr.co/edit/?p=preview)
Create a new tag by typing something then you will see the label added in the select options if you select it via keyboard enter the label vanishes, & the tag is added without the label 
but ...
if you select via mouse click the label stays...

### description
Desired outcome: When you type in "Tag" the selection screen will show "Tag(new)". When you select "Tag(new)", the actual tag will be called "Tag".
Actual outcome: The created tag is still called "Tag(new)".
### images
the bug!
image
the desired outcome!
image
the label added to the select option.
image

thanks!

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -386,8 +386,10 @@ uis.controller('uiSelectCtrl',
       if (!item || !_isItemDisabled(item)) {
         // if click is made on existing item, prevent from tagging, ctrl.search does not matter
         ctrl.clickTriggeredSelect = false;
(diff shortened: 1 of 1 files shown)


shmuel-web:
thanks!

Jefiozie:
Hi , thanks for submitting this PR. Could you clear the dist folder? Then I can have a good view on your changes. Thanks.

PR #1997, opened 2017-05-25
Ui select header footer

kevinwc:


Jefiozie:
Closing this one down as i think this is a duplicate of #1982

PR #2003, opened 2017-06-02
fix: close drop-down regardless of `stopPropagation()` calls

ReicoC:
### Bug description
As of now, a click event-handler on the `$document` is responsible for closing an open drop-down.
The click-event stops bubbling when `stopPropagation()` is called anywhere on the DOM before the `$document`. This breaks the closing of the drop-down. 

### Proposed solution 
Registering the event-listener in the capture phase instead of the bubble phase fixes this bug.

### Plunkers
Failing use-case: http://plnkr.co/edit/MWOhIKdUNpoQK5OWTC5G
Fixed use-case: http://plnkr.co/edit/jC2R9QJGUrbzljr0YR1C

### Additional information
I tested the examples and they're still working with my proposed solution. Tests are also succeeding, including the one i added. 

Angular version used downstream: v1.5.11
angular-ui version used downstream: 0.19.4

### DIFF
--- a/src/uiSelectDirective.js
+++ b/src/uiSelectDirective.js
@@ -1,6 +1,6 @@
 uis.directive('uiSelect',
-  ['$document', 'uiSelectConfig', 'uiSelectMinErr', 'uisOffset', '$compile', '$parse', '$timeout',
-  function($document, uiSelectConfig, uiSelectMinErr, uisOffset, $compile, $parse, $timeout) {
+  ['$document', 'uiSelectConfig', 'uiSelectMinErr', 'uisOffset', '$compile', '$parse', '$timeout', '$window',
+  function($document, uiSelectConfig, uiSelectMinErr, uisOffset, $compile, $parse, $timeout, $window) {
 
   return {
     restrict: 'EA',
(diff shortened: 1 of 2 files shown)


src/uiSelectDirective.js
```diff
@@ -206,11 +206,15 @@ uis.directive('uiSelect',
...
         }
 
-        // See Click everywhere but here event http://stackoverflow.com/questions/12931369
-        $document.on('click', onDocumentClick);
+        // See Click everywhere but here. Similar approach to http://stackoverflow.com/questions/12931369
+        // but using the capture phase instead of bubble phase of the event propagation.
+        //
+        // Using the capture phase avoids problems that araise when event.stopPropatagion()
+        // is called before the event reaches the `document`.
+        $window.document.addEventListener('click', onDocumentClick, true);
```
Jefiozie: Why should we not use `$document.addEventListener` instead of `$window.document`?

ReicoC: In [uiSelectchoicesDirective.js#L51](https://github.com/nzzdev/ui-select/blob/6f19d5fb738c75c960a6cfc4eda0bfa6a0769350/src/uiSelectChoicesDirective.js#L51) ```$window.document``` is used, so i wanted to be consistent.

Also it is not possible to add a native ```addEventListener``` to the angular $document, which is a jQuery [```window.document```](https://docs.angularjs.org/api/ng/service/$document) wrapper. 
Attaching native eventListener to jQuery-objects is not possible, which is confirmed [for example here](https://stackoverflow.com/questions/21680825/event-capturing-event-bubbling-and-jquery-on).

jQuery's aim is browser compability and capture phase is not supported by older IE's. Because Angular only [supports IE9+](https://angular.io/docs/ts/latest/guide/browser-support.html) this should not be an issue.

ReicoC:
@Jefiozie i responded to your question.

ReicoC:
Thanks for your quick review and approval !

PR #2015, opened 2017-06-17
fix($selectMultiple): allow user to override `getPlaceholder()` Behavior

### LINKED ISSUE #1796: Keep Placeholder activated in multipleSelect component
Hello!
I stumbled upon a problem today that is caused by the following lines:
`uiSelectMultipleDirective.js`

```
ctrl.getPlaceholder = function(){
        //Refactor single?
        if($select.selected && $select.selected.length) return;
        return $select.placeholder;
      };
```

I would like to have my placeholder activated even after I selected the first option in a multiple select component. Is there a plan to refactor this? Maybe to use a property like: `keepPlaceholder`

Thanks!

DmitryGonchar:
User can provide their custom `getPlaceholder()` function. E.g. to keep placeholder activated in multipleSelect component, you can use this one:
```
angular.module('myApp').directive('myAppUiSelectPlaceholderAlwaysVisible', function() {
  return {
    require: 'uiSelect',
    link: function($scope, $element, attrs, $select) {
      $select.getPlaceholder = function () {
        return $select.placeholder;
      }
    }
  }
});
```
And then add this directive to the `<ui-select>` html element

closes #1796

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -66,7 +66,6 @@ uis.controller('uiSelectCtrl',
   };
 
   ctrl.getPlaceholder = function(){
-    //Refactor single?
     if(ctrl.selected && ctrl.selected.length) return;
     return ctrl.placeholder;
   };
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -65,6 +65,12 @@ uis.controller('uiSelectCtrl',
     return isNil(ctrl.selected) || ctrl.selected === '' || (ctrl.multiple && ctrl.selected.length === 0);
   };
 
+  ctrl.getPlaceholder = function(){
+    //Refactor single?
+    if(ctrl.selected && ctrl.selected.length) return;
+    return ctrl.placeholder;
+  };
+
   function _findIndex(collection, predicate, thisArg){
     if (collection.findIndex){
       return collection.findIndex(predicate, thisArg);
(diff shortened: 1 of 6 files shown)


DmitryGonchar:
NOTE: user cannot override `$selectMultiple.getPlaceholder()` function, bc `<ui-select-multiple>` is appended inside the component. That's why I had to move it to the $select

src/uiSelectController.js
```diff
@@ -65,6 +65,12 @@ uis.controller('uiSelectCtrl',
     return isNil(ctrl.selected) || ctrl.selected === '' || (ctrl.multiple && ctrl.selected.length === 0);
   };
 
+  ctrl.getPlaceholder = function(){
+    //Refactor single?
```
Jefiozie: Can you remove this comment?

test/select.spec.js
```diff
@@ -15,6 +15,14 @@ describe('ui-select tests', function () {
     Escape: 27
   };
 
+  var multipleTagsTmpl =
```
Jefiozie: Can't you use `createUiSelectMultiple`

DmitryGonchar:
@Jefiozie done

Btw, have you considered splitting tests into multiple files? Better readability, mainainability, it will be clear which tests relate to what and which helper functions available. Also will be easier to run only tests you are working on during dev phase.

DmitryGonchar:
@Jefiozie How about merging this one?

Issue #2010, opened 2017-06-09
Wrong inputWidth, input too long


NaturalL:
### Bug description: Wrong inputWidth, the inputWidth shoud  minus the  container border-width or padding with different theme.

    -          var inputWidth = containerWidth - input.offsetLeft - 10;
    +          var inputWidth = containerWidth - input.offsetLeft;

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/OAemHKmhjJmB0iF2lMX0

UI-Select: Version: 0.19.8

Jefiozie:
I think you are refering to this [line 839](https://github.com/angular-ui/ui-select/blob/master/dist/select.js#L839) but i don't see the - 10. 

Please share some more info. Thanks.

NaturalL:
@Jefiozie This commit 7ad4ef13273bf0900de0735fc5d6e04003c0bfc3 remove "- 10", but I think "- 10" is necessary.

aurelienlt:
Already known issue: https://github.com/angular-ui/ui-select/issues/1980
But strangely it has never been fixed.

PR #2016, opened 2017-06-17
fix(uiSelectCtrl): fix input width calculations

### LINKED ISSUE #1980: Input box for ui-select multiple ALWAYS forcing text entry at new line
### Bug description:
**Input box for ui-select multiple forcing text entry at new line, even if values do not take up entire first line**

There is already closed issue for this #1556 undesired behavior.. but for older version. Since I got no help there, opening new issue.

[plunker ](http://plnkr.co/edit/97IqGQc1u2245IPZpt3P)
All selects forces new line when there is at least 1 item selected.. this was not happening before..

Angular: 1.6.4
UI-Select: ^0.19.8
Bootstrap/Select2/Selectize CSS (if applicable): ^3.3.7

DmitryGonchar:
input is not too big or too small - it takes the whole remaining part of the current row.
If smaller than the min input width - goes to the next row

closes #1980, possibly #2010

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -528,14 +528,14 @@ uis.controller('uiSelectCtrl',
         container = ctrl.$element[0],
         calculateContainerWidth = function() {
           // Return the container width only if the search input is visible
-          return container.clientWidth * !!input.offsetParent;
+          return $(container).width() * !!input.offsetParent; // width must be WITHOUT padding - to calculate space for text correctly
         },
         updateIfVisible = function(containerWidth) {
           if (containerWidth === 0) {
             return false;
           }
           var inputWidth = containerWidth - input.offsetLeft;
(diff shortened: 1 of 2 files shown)


DmitryGonchar:
@Jefiozie 
It works locally when I run `gulp test` (and it opens Chrome), but does not work in Travis )in Firefox)
Looks like selected item width is different in Firefox somehow.

Do you know how to fix this test?

Jefiozie:
I think it has something to do with the Firefox version (`31`) but I don't know this for sure.  Basically your test is saying it should be 792 px ( it think) and in the Firefox 31 browser this isn't the case. Do you see local problems when running it on `IE` or `Edge`?

DmitryGonchar:
@Jefiozie Yes, works in both.

Btw, added a new commit for missing karma plugins to run in these browsers

PR #2017, opened 2017-06-18
feat(event): add event to select and remove calls

randdusing:
Add event to select and remove calls. This allows the developer to how the the item was selected or removed.


Jefiozie:
To clarify the use case you need. You want public events to be fired after a `select` or `remove` call has been done?

randdusing:
I just need select, but I thought it would be nice to pass the same event to the remove callback in case someone else needed it.

Issue #1218, opened 2015-09-30
CSP errors caused by inline script


rchdly:
I'm seeing CSP errors due to `.ui-select-row-choices-inner` elements having an `href="javascript:void(0);"` - inline scripts are not allowed by CSP.

I suggest using `href="#"` here and adding `$event.preventDefault()` to the `select` function.

pra85:
This issue has been addressed by the pull request #1182

wesleycho:
Closing as already fixed.

graingert:
@wesleycho should be re-opened because there's yet more inline JS. #1958

PR #2023, opened 2017-06-27
Remove inline javascript causing CSP errors

juliushaertl:
This PR removes inline javascript call in the select2 template that causes CSP error messages when 'unsafe-inline' is not allowed via CSP headers.

Related to https://github.com/angular-ui/ui-select/issues/1218


juliushaertl:
Closing since no activity here, feel free to reopen and merge if it is of interest.

barbossa:
I confirm the issue with href="javascript:;". Can we merge this PR to master?

Issue #1151, opened 2015-08-19
Tagging tokens and different languages


rusmaxim:
In English comma - 188, in Russian - 191. And when i set tagging-tokens to comma it not work in Russian.

Can we change tagging tokens for not special keys from codes to letters?

Sorry for bad english.

donghwicha:
sad but nobody cares for this. anyone?

Jefiozie:
Dont know if this is still a issue for you but I just read the documentation and the version there is right now you can set your own character for this function. Hope this will help on your problem.


> Specify keyboard keys that will trigger creation of a new tag. Multiple keys can be separated by a pipe |, SPACE is declaration for literal spacebar. (Defaults to ENTER and ,)
[Documentation](https://github.com/angular-ui/ui-select/wiki/ui-select)

rusmaxim:
It still same. If I set comma, tagging work on english button comma, and if I switch to Russian I can't press button with letter.

Jefiozie:
Looking at the code there is a mapping between the keys. So comma is mapped to key code 188.
A thing you could do is add your custom stuff to the list (maybe not the best way but the quickest).

But maybe we could have a look on how to fix it. Could you test something for me?
I have made a plunker where the e.code is outputting your input in the console. Trying to find out if there is a general mapping between multiple layouts on keyboards.

[Link to plunker](https://plnkr.co/edit/hG6NgjfzVAKOAobX4KMc?p=preview)

Thanks.

rusmaxim:
@Jefiozie For example MacBook Pro Keyboard
English:
```
Code: Comma key: ,
Code: ShiftRight key: Shift
Code: Digit6 key: ^
```

Russian
```
Code: Comma key: б
Code: ShiftRight key: Shift
Code: Digit6 key: ,
```

kln3wrld:
I seem to have a similar problem. I'm typing on a Belgian dot (azerty) keyboard (not to confuse with the french azerty).

My semicolon gets detected as a comma in your plunker:

    Code: Comma key: ;

And my comma gets detected as an m key:

    Code: KeyM key: ,

However, only the semicolon is giving me problems, that one won't trigger the tagging no matter what I try.

PR #2041, opened 2017-08-15
Fix issue #1151

nornik:
Hi!

Please review my solution of https://github.com/angular-ui/ui-select/issues/1151.

I think i do something wrong with versions, please correct me and if possible implement corrections.

Thank you.


Jefiozie:
Yeah there are some things that should not be there. 
Could you clean up the dist files, package.json, bower.json

Thanks.

PR #2045, opened 2017-08-22
fix(uiSelectCtrl): properly calculate container width

### LINKED ISSUE #1980: Input box for ui-select multiple ALWAYS forcing text entry at new line

zollinger:
- Account for paddings on input container in case box-sizing: border-box
is set.
- Added some tests for this case

The other open PR  (
https://github.com/angular-ui/ui-select/pull/2016) regarding this issue does not work without jQuery.

Closes #1980.


joebandenburg:
I find that the bug persists even without padding because the `ui-select-match` element has a non-integer width (e.g. `138.34`) and there for search box is 1 pixel too large.

zollinger:
I'm running this fix on a production app without problems. Could you maybe create a jsfiddle showing the issue?

joebandenburg:
After some more digging, I've found that the cause is if there is no left or right padding on the element. http://plnkr.co/edit/EKXqPepAaswBpOAr9W34?p=preview

Issue #2050, opened 2017-08-31
Bug: _ensureHighlightVisible throws [ui.select:choices] Expected multiple .ui-select-choices-row but got '0'.


zollinger:
### Bug description:

As _ensureHighlighVisible() is being called in a setTimeout() might be triggered in situations where the ui-select is actually closed and thus, the options aren't rendered. This clashes with the assertion of "choices.length > 1" made in _ensureHighlightVisible. It happens frequently on slower machines when double clicking the input field.

This is the same bug that was observed in #1355 (which was not a duplicate of #224, as the error thrown is quite different).

### Link to minimally-working plunker that reproduces the issue:

To reproduce: 

1. Open plunkr below
2. Enable Chrome performance throttling and set it to 10 - 20x slower
3. Double click the ui-select element

http://plnkr.co/edit/ShmVlq3Xk7QVHurt6BLU?p=preview

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: latest

UI-Select: latest

Bootstrap/Select2/Selectize CSS (if applicable): Bootstrap

PR #2051, opened 2017-08-31
fix(uiSelectCtrl): ensureHighlightVisible triggered with closed dropdown

zollinger:
This fix ensures that the dropdown actually is open before calling
_ensureHighlightVisible.

I didn't manage do reproduce the issue meaningfully in a test in the time I have available, but I think the change is quite obvious.

Here's a plunkr with the fix:
http://plnkr.co/edit/iV43PiaAzk0paO0ElZzT?p=preview

Fixes: #2050


egarkavy:
Why was this commit reverted?

PR #2057, opened 2017-09-13
feat(uiSelectMultiple): click anywhere to search when using bootstrap theme

andreiho:


Jefiozie:
Can you collabrator a bit more on what this feature should do? The use case etc?

Thanks.

andreiho:
@Jefiozie This PR fixes https://github.com/angular-ui/ui-select/issues/1980. All the information you need is in the issue. Simply put, this gets rid of the over-complicated calculations used to set the width of the search input and provides a basic solution to allow clicking anywhere in the box to focus on the search input.

parker789:
@Jefiozie can we get this PR merged, or another css solution using flexbox? I'm having this same issue, and would like to not have to run a forked version of this in production

Mangesh-P:
@Jefiozie can we get this PR merged

PR #2063, opened 2017-10-13
fix highlight when using animate

gboros:
when animate is used, the dropdown is not scrolled to the selected item - this commit calls "_ensureHighlightVisible" in this case, to fix it


Jefiozie:
HI @gboros can you show a working an non working example for me? 
So I can see what the original problem is ,thanks

gboros:
hi
here is a non-working example : https://plnkr.co/edit/TWDf59jpdY7UuFf3BtTI?p=preview The dropdown should scroll to 'eleven'. If you change to use the 'select-fixed.js', that contains the PR, it will scroll

PR #2069, opened 2017-10-24
fix(uiSelectHeaderGroupSelectableDirective): do not issue error if feature is disabled

Klinton90:
My company uses wrapper around ui-select, so we always need `uiSelectHeaderGroupSelectableDirective` directive to exist on template even when ui-select is used for `single`/`no-groups` mode. In current implementation, if directive is disabled and other conditions do not pass, `console.error()` will be executed. That fix prevents error message to be shown for disabled directive.


DarkIsDude:
Hi @Klinton90 , look good to me 👍 .

PR #2074, opened 2017-11-03
Bootstrap4 template choices.tpl.html

huggler:
Change ng-show="$select.open && $select.items.length > 0"
for this ng-class="{show : $select.open && $select.items.length > 0}"


huizhoujava:
Can someone help review and merge the pull request?

By the way, I tried the built select.js and select.css from https://github.com/bmontgomery/ui-select/tree/master/dist with Bootstrap 4.1.1 and AngularJS 1.7.2, but the select widget is not working.

If I add attribute theme="bootstrap4" on ui-select element, it is rendered like this
ui-select-bootstrap4

If I add attribute theme="bootstrap" on ui-select element, it is rendered like this
ui-select-bootstrap4-bootstrap

For both theme, when I click the input box , the choices are not displayed.

If I change to use Bootstrap 3.3.7 with attribute theme="bootstrap" on ui-select element, the select widget is working as expected.

Here is the code I use for testing [ui-select-test.zip](https://github.com/angular-ui/ui-select/files/2156367/ui-select-test.zip)

PR #2080, opened 2017-11-16
 Stop Safari from triggering autocorrect

jpike88:
This makes behaviour consistent


src/bootstrap/select.tpl.html
```diff
@@ -1,7 +1,12 @@
 <div class="ui-select-container ui-select-bootstrap dropdown" ng-class="{open: $select.open}">
   <div class="ui-select-match"></div>
   <span ng-show="$select.open && $select.refreshing  && $select.spinnerEnabled" class="ui-select-refreshing {{$select.spinnerClass}}"></span>
-  <input type="search" autocomplete="off" tabindex="-1"
+  <input type="search" 
+         autocomplete="off"
+         autocorrect="off"
+         autocapitalize="off"
```
Jefiozie: Is the autocapitalize not deprecated on some browsers?

PR #2091, opened 2017-12-22
Fix border radius when using with input-group

### LINKED ISSUE #2014: Bootstrap border-radius broken when using input-group
On the [demo](https://angular-ui.github.io/ui-select/demo-bootstrap.html) and when using in local, when an ui-select is wrapped inside an `input-group`, the border radius behavior is lost on the `form-control`

northteam:
Fix #2014 .

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -195,6 +195,16 @@ body > .select2-container.open {
   position: static;
 }
 
+.input-group .ui-select-bootstrap:first-child .form-control {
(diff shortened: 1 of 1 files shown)


northteam:
Plnkr showing the bug: http://plnkr.co/edit/GW08U5fapGjRd4n24D4D?p=preview
Plnkr with the fix: http://plnkr.co/edit/F2nVeCutBo5Fze9dl9Nv?p=preview

Issue #2054, opened 2017-09-11
Navigate on items of the select with keyboards directional keys affect the cursor of the "input match"


lucasscarduelli:
### Bug description: 
When I navigate on items of the select with keyboards directional keys the "input match" cursor is affected. When I press key down, the cursor goes to the end of the text, when I press key up, the cursor goes to the begin. 

This behavior is prejudicial to the user because it is has to pay attention of the "input match" when write a new caracteres to filter.

### Link to minimally-working plunker that reproduces the issue: 
http://angular-ui.github.io/ui-select/demo-basic.html

### Version of Angular, UI-Select, and Bootstrap/Select2/Selectize CSS

Angular: v1.4.8

UI-Select: v0.19.7

Jefiozie:
Can you explain a bit more, I think i'm misunderstanding you on your problem.

lucasscarduelli:
Were you able to reproduce the behavior in the example?

Example: http://angular-ui.github.io/ui-select/demo-basic.html

lucasscarduelli:
The problem is behavior. It damages the usability of the component to the user.

Jefiozie:
No sorry, maybe can you make a video of gif off the issue?

chimmi:
I think I`m having the same issue. You can go to https://angular-ui.github.io/ui-select/demo-basic.html, input "ad" without quotes in Select2 or Selectize input, Now when you have some results and some text in search field you can press UP and DOWN keys to navigate results, but this also moves caret in the search field to start/end respectively.

PR #2093, opened 2017-12-26
Update uiSelectController.js

chimmi:
Possible fix to #2054

### DIFF
--- a/src/uiSelectController.js
+++ b/src/uiSelectController.js
@@ -617,7 +617,7 @@ uis.controller('uiSelectCtrl',
 
     var key = e.which;
 
-    if (~[KEY.ENTER,KEY.ESC].indexOf(key)){
+    if (~[KEY.ENTER,KEY.ESC].indexOf(key) || KEY.isVerticalMovement(key)){
       e.preventDefault();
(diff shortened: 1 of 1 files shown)


Jefiozie:
Can you build a test around it?

chimmi:
Do you mean a test that proves that this fixes the issue?

Jefiozie:
yes if you want 😄

chimmi:
@Jefiozie Test should be as simple as this:
```
describe('Caret position', function () {
    it('should not change caret position in search field when selecting option with keyboard', function () {
      var el = createUiSelect(), searchField = $(el).find('.ui-select-search');
      openDropdown(el);
      setSearchText(el, 'ad');
      triggerKeydown(searchField, Key.Down);
      expect(searchField[0].selectionStart).toBe(2);
      searchField[0].setSelectionRange(0, 0);
      triggerKeydown(searchField, Key.Up);
      expect(searchField[0].selectionStart).toBe(0);
    });
  });
```
unfortunately `triggerKeydown(searchField, Key.Up);` does not seem to trigger carret move, so the test passes even without a fix.

PR #2099, opened 2018-01-11
Fix: Search box not shown to search when data is an empty array with select2 theme

### LINKED ISSUE #1805: Search box not shown to search when data is an empty array
I have a ui-select that uses refresh to fetch data from server whenever a user inserts a char in input box (search box). 
Since UI-Select v0.19.3 the input box is no longer available when data is an empty array, so the user cannot search to fetch data.
Needless to say that I only want to present data when the user search for it.

Angular version: 1.5.5
UI-Select: 0.19.3

Sparksx:
Fix #1805 

Juste fork, fix into `src/uiSelectDirective.js` and build


Jefiozie:
Please clean up the dist

Sparksx:
dist cleaned

Sparksx:
After more investigation, lot of problems come from empty array set and $animate event.
Without $animate, empty array works perfectly

We need to investigate more about that

Jefiozie:
@Sparksx do you  want to investigate a bit more on how to resolve the animation problem? Next , thanks for the PR could you explain if this problem is only present in the select2 theme?

PR #2101, opened 2018-01-18
Update uiSelectController.js

wowDaiver:


Jefiozie:
Could you describe what bug/feature this fix is for?

wowDaiver:
For example https://plnkr.co/edit/2ajfL8?p=preview
Moving from ui-select to ui-select without closing the list results in an incorrect focus.
This fix allows you to close the list when you click on TAB, but do **not go to the next item**.
So do not rush to merge

PR #2108, opened 2018-02-13
chore(tests): use headless browser for tests

neolitec:
Hi,
I've made a very small update on the test to use a headless browser instead of a real one that pop in my face every time I run tests.
I would like to share it, so... voilà.

:)


neolitec:
Tests don't pass because Node is too old for Puppeteer...
Is it necessary to use a non headless browser for tests? Couldn't we also use Puppeteer on Travis?

PR #2125, opened 2018-04-19
Input box for ui-select multiple ALWAYS forcing text entry at new line #1980

sunilkandakatla:
Fix(#1980) Input box for ui-select multiple ALWAYS forcing text entry at new line #1980


biggieman:
I think it is better:
```
var inputWidth = containerWidth - input.offsetLeft;
if (multiple) {
	inputWidth -= 10;
}
```

PR #2129, opened 2018-04-23
Fix issue #2110

rogercoder:
Simple fix to satisfy WCAG compliance checkers. The drop-down caret is now marked aria-hidden. 

As long as the input-id attribute on the ui-select tag is used along with a matching <label for="..." then I would think screen readers should be able to make sense of it.


rogercoder:
For issue 2110: https://github.com/angular-ui/ui-select/issues/2110

PR #2150, opened 2018-07-19
Udpate karma version in package.json

hongyanh:
When I was trying to use `gulp` to build the project, there is a dependency issue cased by outdated karma package:
```
Store.prototype.__proto__ = EventEmitter.prototype;
                                         ^
TypeError: Cannot read property 'prototype' of undefined
```
After updated karma version from 0.12.16 to 2.0.4 in package.json file, the error is gone when I build the project.


andrewkittredge:
Updating to karma 3.0.0 also works.

Issue #2025, opened 2017-06-27
CSP inline style error when using ui-select


corbinolds:
We are seeing two inline styles being applied when using the ui select directive, specifically with the ui-select-match directive. We are using the ng-csp directive at the top level and that is stopping inline styles. This error is happening on line 2888 of angular.js in version 1.5.8 (See tmp.innerHTML = wrap[1] + html.replace(XHTML_TAG_REGEXP, "<$1></$2>") + wrap[2];) We are using version 0.19.5 of ui select.

````
function jqLiteBuildFragment(html, context) {
  var tmp, tag, wrap,
      fragment = context.createDocumentFragment(),
      nodes = [], i;

  if (jqLiteIsTextNode(html)) {
    // Convert non-html into a text node
    nodes.push(context.createTextNode(html));
  } else {
    // Convert html into DOM nodes
    tmp = fragment.appendChild(context.createElement("div"));
    tag = (TAG_NAME_REGEXP.exec(html) || ["", ""])[1].toLowerCase();
    wrap = wrapMap[tag] || wrapMap._default;
    tmp.innerHTML = wrap[1] + html.replace(XHTML_TAG_REGEXP, "<$1></$2>") + wrap[2];

    // Descend through wrappers to the right content
    i = wrap[0];
    while (i--) {
      tmp = tmp.lastChild;
    }

    nodes = concat(nodes, tmp.childNodes);

    tmp = fragment.firstChild;
    tmp.textContent = "";
  }
````

Here is a console stack trace:

> Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-4IfJmohiqxpxzt6KnJiLmxBD72c3jkRoQ+8K5HT5K8o='), or a nonce ('nonce-...') is required to enable inline execution.
> 
> jqLiteBuildFragment @ lib.built.js:2889
> jqLiteParseHTML @ lib.built.js:2921
> JQLite @ lib.built.js:2966
> JQLite @ lib.built.js:2962
> removeComments @ lib.built.js:10195
> (anonymous) @ lib.built.js:9609
> processQueue @ lib.built.js:16384
> (anonymous) @ lib.built.js:16400
> $eval @ lib.built.js:17683
> $digest @ lib.built.js:17496
> $apply @ lib.built.js:17791
> done @ lib.built.js:11832
> completeRequest @ lib.built.js:12034
> requestLoaded @ lib.built.js:11967

[...]

PR #2181, opened 2019-08-18
 fix(CSP): avoid inline execution of styles in match and select-multi…

evilaliv3:
This pull request fixes issue #2025 using the proper ng-style directive in place of style attribute.


evilaliv3:
@Jefiozie: Being this a fix for the CSP it does probably fix an issue for many users.

When you find the time would you please review it and eventually issue a new package?

thank you!

